PR #41, opened 2016-05-11
Fixed null pointer crash

sgehrman:
Had a problem with the order Start was getting called.  I was getting a crash. Let me know if you want to reproduce the problem or have questions.


sgehrman:
still crashing, need to fix it more... will close for now.

PR #42, opened 2016-05-11
Fixed crash

sgehrman:
I'm not using your CameraRig, so I got some crashes. Here's a fix when using the SteamVR camera rig instead.  Any reason why you require your own CameraRig?  Let me know.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_InteractGrab.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_InteractGrab.cs
@@ -58,7 +58,7 @@ public class SteamVR_InteractGrab : MonoBehaviour
     }
 
     // this was being called in Start, but that would crash if SimplePointers Start had not been called yet.  Just get when needed
-    private void GetAttachPoint()
+    private Rigidbody GetAttachPoint()
     {
         //If no attach point has been specified then just use the tip of the controller
         if (controllerAttachPoint == null)
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_InteractGrab.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_InteractGrab.cs
@@ -23,6 +23,8 @@ public class SteamVR_InteractGrab : MonoBehaviour
     public event ObjectInteractEventHandler ControllerGrabInteractableObject;
     public event ObjectInteractEventHandler ControllerUngrabInteractableObject;
 
+    public float throwMultiplier = 1.0f;  // set this higher to give the throw some juice
+
     FixedJoint controllerAttachJoint;
     GameObject grabbedObject = null;
 
@@ -140,12 +142,12 @@ public class SteamVR_InteractGrab : MonoBehaviour
         var device = SteamVR_Controller.Input((int)controllerIndex);
         if (origin != null)
         {
-            rb.velocity = origin.TransformVector(device.velocity);
+            rb.velocity = origin.TransformVector(device.velocity) * throwMultiplier;
             rb.angularVelocity = origin.TransformVector(device.angularVelocity);
         }
         else
         {
-            rb.velocity = device.velocity;
+            rb.velocity = device.velocity * throwMultiplier;
             rb.angularVelocity = device.angularVelocity;
         }
         rb.maxAngularVelocity = rb.angularVelocity.magnitude;
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_InteractGrab.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_InteractGrab.cs
@@ -57,7 +57,8 @@ public class SteamVR_InteractGrab : MonoBehaviour
         controllerActions = GetComponent<SteamVR_ControllerActions>();
     }
 
-    private void Start()
+    // this was being called in Start, but that would crash if SimplePointers Start had not been called yet.  Just get when needed
+    private void GetAttachPoint()
     {
         //If no attach point has been specified then just use the tip of the controller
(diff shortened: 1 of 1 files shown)


thestonefox:
Hey,

Thanks for taking time to raise a pull request, nice to see someone else taking an active interest!

The reason for the custom CameraRig prefab is the one included with the SteamVR Unity Plugin is inconsistent. If you look at their default CameraRig, it has no children in the `Controller -> Model` (e.g. tip, base, buttons, etc). Now if you open the example scene TestThrow in the SteamVR Unity Plugin, you'll see the CameraRig does have the children in the `Controller -> Model`

These children are required to create the attach point in their own example, yet on their default prefab they don't exist. So even using their own default CameraRig prefab with their own scripts would fail.

I think this is something wrong with their plugin (There seems to be many issues with the quality of Valve's plugin for some reason).

The CameraRig prefab I have included is the same one lifted directly from their TestThrow example as that seems to be the correct one and they just forgot to update the one in their own prefabs directory.

I would love to understand why there is this inconsistency in their plugin, knowing that would certainly lead to a proper fix.

Also, I'll put up a contributing guide, to ensure that commit messages adhere to a standard. If people are going to contribute to this then I'm going to set up a gitcop account to check the integrity of the commit messages.

I'm a bit of a stickler for good, rigid commit messages that follow a structure. Git commit history forms a great record of changes, and the format I use (adopted from Google's official format) allows for generation of changelogs.

This is a huge response! Holy wall of text! But, do you think your code is only required with the use of the SteamVR Plugin CameraRig prefab? If so, then for now I'd say until it's understood what's inconsistent about it, it's best to just ue the prefab included in this toolkit.

Aaaaand, finally :)

The throw multiplier is something I've tried, but having high numbers doesn't seem to create the desired effect, anything over about 1.5f seems to go crazy. I'm not sure how your tests came out. I've also tried adding a force behind the object to increase velocity and that's even worse!

I think it would be a good idea to move that throw multiplier into it's own pull request (as it's independent of the other changes) and then it can be tested and if a solution works it can be merged.

Thanks for supporting this and helping out, it's much appreciated ❤️ 💚 💙 💛 💜

sgehrman:
I noticed that the children get added to the model when things startup.  I
didn't see where that was happening, but it's working for me just fine as
far as I can tell.  I can teleport around and the laser and ball thing show
up just fine.  I did have to add the rigidbody to the tip, but otherwise it
seems OK.  I can also grab and throw stuff.  I haven't noticed any further
problems with using the one from SteamVR.

the throwmultiplier thing works fine.  I can set it to 3-5 (haven't tried
others) and it seems to work as expected.

I'm going full speed ahead on Vive dev and will try to submit push
requests.  I'm very careful with my edits, so don't worry, I would just
save time and merge everything just to get some momentum.

-steve

thestonefox:
Yeah I've seen that they get added at runtime, but I can't see how that happens, nor do I understand why Valve have a prefab where they are not added at runtime for their TestThrow example. I'm reluctant to use their prefab until these things are clear as to why it's that way. Also, adding the tip rigidbody in the code doesn't seem as how Valve intended it because they expect the attach point to be any of the controller elements (as long as they have a rigidbody). So specifying the tip rigidbody in code means only that element can be used as the attach point. Personally, I think Valve have left it in a half finished state, because many of the override models don't work (for different controller types)

I'll raise an issue on their repo to see if I can get an answer from them, but for now I think it's best holding off on that commit.

As for the throw multiplier, do you want to create a new pull request with that just on it's own (following the new contribution guidelines I've just put up at: https://github.com/thestonefox/SteamVR_Unity_Toolkit/blob/master/CONTRIBUTING.md

(effectively, ensure the commit message following the correct format)

Then I'll pull it down, do some checks and if it's all ok I'll merge it in.

sgehrman:
I'm not so good with Github pull requests and since I'm always swamped, I
don't know if I'll have time to do things properly.

Let me see how things go, it's hard enough as it is.  Do you know how to
make this project somekind of submodule?  I make changes to this code in my
copy of it in my app, but then I have to hand copy over the changes to my
fork of your project and then push it and make the pull request.  Is there
an optimal way of setting it up so that changes in your code get pushed to
the right place rather than in my app?

-steve

thestonefox:
Not to worry, I can implement your throw multiplier and test it, then push it if it's working.

You can use submodules in git (add the repo to your existing repo using `git submodule add`) then you should be able to maintain that module within your parent git repo.

I think it may also be pushed to your main app too but if your changes get merged into the main project then you'll have no conflicts. The issue may arise if you make custom changes to the toolkit repo and they don't get merged, then newer versions could cause conflicts in your repo.

I really should look at getting this toolkit on as a Unity Plugin, but as you say, time is always against us! :(

vad710:
Hello Stone,

I ran into this same issue and ended up making changes similar to the original PR - the slight difference is that the changes I made would work on either Valve's prefab or the prefab that is included in this repo.

I understand your reluctance of working with the prefab - My guess is that the GameObjects are generated dynamically during runtime because SteamVR is based on OpenVR which is supposed to work with any VR hardware, not just the Vive.

Anyway, I was just wondering if you would be interested if I made a pull request where the code would work with both Valve's Prefab and the Prefab that is included in the project.

Thanks for making this available! I am looking forward to contributing to this project.

thestonefox:
@vad710 sure I'm happy to see the pull request, I'd love to be able to get rid of the custom prefab, but I need to understand why Valve provided 2 prefabs in the first place.

SteamVR and OpenVR seem to be a bit of a black box mystery in some parts, it's easy to assume because Valve are behind it and it's documented then it's correct, but I've seen far too many examples of where they've been sloppy, it's a real concern.

thestonefox:
Outdated PR, would be an uphill battle to fix the merge conflicts. If you think anything in here is still relevant please raise a new PR (based on the feedback above)

PR #91, opened 2016-05-24
Added Archery Demo

Blueteak:
Added:
- Parenting pickup type
- ForceTouch(object) method
- Hide controller for parented objects
- Bow/Arrow models (Bow animated)
- Bow/Arrow firing scripts
- Archery Demo scene


Assets/SteamVR_Unity_Toolkit/Scripts/Archery/Quiver.cs
```diff
@@ -0,0 +1,48 @@
...
+    public GameObject VRHead;
+
+    float internalCD = 0.15f;
+    float cd;
+
+    void Start()
+    {
+        Bow = GetComponent<BowAim>();
+        if (VRHead == null)
+            VRHead = FindObjectOfType<SteamVR_GameView>().gameObject;
```
thestonefox: Finding the headset using SteamVR_GameView doesn't work in Unity 5.4 because it's been deprecated.

Touchpad_Walking.cs has the same issue so this is how I resolved it:

https://github.com/thestonefox/SteamVR_Unity_Toolkit/blob/master/Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_TouchpadWalking.cs#L37-L41

Same should work here.

Assets/SteamVR_Unity_Toolkit/Scripts/Archery/ArrowNotch.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    public GameObject arrow;
+    SteamVR_InteractableObject obj;
+
+    void Start()
+    {
+        obj = GetComponent<SteamVR_InteractableObject>();
+    }
+
+	void OnTriggerEnter(Collider col)
```
thestonefox: crazy spacing here :)

Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_ControllerEvents.cs
```diff
@@ -28,13 +28,15 @@ public enum ButtonAlias
     public int axisFidelity = 1;
 
     public bool triggerPressed = false;
+    public bool triggerDown = false; //Seems to work more reliably than triggerPressed? Added 5/24/2016 - Blueteak
```
thestonefox: This is a bit of an anomaly, The way triggerDown is being set is not really that much different to how triggerPressed is being set.

the `OnTriggerClicked` method is called by passing the results of the `SetButtonEvent` method and that method sets the boolean.

There "should" be no difference between the two variables.

Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_InteractGrab.cs
```diff
@@ -116,9 +122,24 @@ private void SnapObjectToGrabToController(GameObject obj)
...
+            ParentObject(obj);
+        //End modification
+    }
 
-        CreateJoint(obj);
+    //Added 5/24/16 - Blueteak
+    Transform lastParent;
+    private void ParentObject(GameObject obj)
+    {
+        //lastParent = obj.transform.parent;
```
thestonefox: If the commented out lines are not required, best to just delete them

Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_InteractGrab.cs
```diff
@@ -116,9 +122,24 @@ private void SnapObjectToGrabToController(GameObject obj)
...
+        obj.transform.SetParent(transform);
+        obj.GetComponent<Rigidbody>().isKinematic = true;
+        //Physics.IgnoreCollision(obj.GetComponent<Collider>(), GetComponent<Collider>());
```
thestonefox: Same here, if it's not needed, just delete the comment :)

thestonefox:
One last thing to be picky about :) is the commit message.

Would you be able to change the commit message to follow the structure of the other commit messages (gitcop has probably already picked up the message format).

it would be something like

`feat(Interaction): add bow and arrow mechanic and new grab type`

thestonefox:
Oh also, one more thing. in the example scene, you can probably delete the [SteamVR] and [Status] objects from the hierarchy as they get generated at run time anyway if they don't exist.

thestonefox:
and another "one more thing"! :)

I've noticed in the example scene, when you put the bow back down (i.e. drop it) the controller that was responsible for grabbing the arrows doesn't reappear

xEvoGx:
Very cool! I have a couple things as well:

- First, the arrow doesn't arc. No matter how hard or light, at any given angle, the arrow remained flat

- The bow feels small due to how short the pull is. Never mind going off reality for a moment, games like **The Lab**, **Vanishing Realms** and **Spell Fighter** have very nice feeling bows with long pulls, so the right controller is by/behind the ear.

Anyway love it!!

jakk2809:
Very nice! Thank you for working on this. I've been trying to get my own bow to work, but I'm taking a different approach. Instead of just pointing the arrow in the direction of the bow I'm making the arrow a child of the bow when you nock it. I tried using the LookDirection, but sometimes it looks like the arrow moves a little while on the arrow rest.

thestonefox:
Hold fire on doing any of the above changes. I've done some modifications that may cause some git conflicts.

Blueteak:
NP, out of town for rest of week, let me know when!

thestonefox:
@Blueteak don't worry about it, I've been porting your code into the latest changes, save you some time :)

Thanks for your hard work so far!

thestonefox:
FYI, I've started a new branch with the rework in.

Bits are the same, bits have changed. I've separated certain things out into different commits.

Fixed a couple of the issues above, also made it so the arrow attaches to the side of the bow you touch it on (rather than always snapping depending on which hand you picked it up with).

I've added the arrow trajectory too so it arcs when you fire it.

I've also rebuilt a new bow model as a demonstration (and to see how it worked) my bow is out of unity primitives. I'm still going to include your bow and arrow, to show how it works with whatever model.

I've not finished with it yet, but just in case anyone is following this PR, then this is the new branch that will eventually be merged:

https://github.com/thestonefox/SteamVR_Unity_Toolkit/tree/feat/archery-mechanic

thestonefox:
Closing this as the work has been done in another branch and merged.

thedonquixotic:
Hey, what happened to the archery demo? I can't find it in the latest master repo, has it been taken out?

thestonefox:
It's still here, it's in the example scene ChildOfController

thedonquixotic:
Oh cool, thanks!

PR #98, opened 2016-05-26
Small safety check added

Frostlock:
Nothing major but I lost some time looking for this when I tried to make the head collision example work in my own project. :-)

On a side note, thanks for all the great examples! This toolkit is a great initiative :)

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_HeadsetCollisionFade.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_HeadsetCollisionFade.cs
@@ -30,6 +30,13 @@ public class SteamVR_HeadsetCollisionFade : MonoBehaviour {
     }
 
     protected void Start () {
+        //Make sure that there is a SteamVR_Fade script available.
+        SteamVR_Fade fadeScript = gameObject.GetComponentInChildren<SteamVR_Fade>();
+        if (fadeScript == null)
+        {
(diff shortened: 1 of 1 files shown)


Assets/SteamVR_Unity_Toolkit/Scripts/SteamVR_HeadsetCollisionFade.cs
```diff
@@ -30,6 +30,13 @@ protected HeadsetCollisionEventArgs SetHeadsetCollisionEvent(Collider collider,
     }
 
     protected void Start () {
+        //Make sure that there is a SteamVR_Fade script available.
+        SteamVR_Fade fadeScript = gameObject.GetComponentInChildren<SteamVR_Fade>();
+        if (fadeScript == null)
```
thestonefox: May be worth just doing the if inline like so:
`if (gameObject.GetComponentInChildren<SteamVR_Fade>() == null)`

Rather than assigned to a variable then never use it anywhere other than the if statement.

Frostlock:
Sure, happy to adapt :)

thestonefox:
You can just force push your changes to an existing pull request, no need to close it to make amends :)

Also, when you raise the next pull request, ensure your commit message meets the contribution guidelines otherwise gitcop will complain (it didnt before because the service was down, but its back up now).

so this would be something like:

`fix(Headset): show warning if fade script is not present`

Frostlock:
I'm learning all kinds of fun stuff :)

thestonefox:
the joys of git! it's such a minefield :/

if you wanna learn more git, then try this git based game called githug

https://github.com/Gazler/githug

PR #107, opened 2016-05-27
feat(CameraRig): Basic Player Gravity

Frostlock:
New example scene and a basic implementation of player gravity.

I'm not entirely sure about the meta files in this commit. I assume Unity updates these as required?


thestonefox:
I had the same idea as what you're proposing.

The TouchpadWalking script already does exactly what you're suggesting here (i.e. adds the rigidbody)

I'm going to separate the rigidbody code out (so you can use it independendly of the touchpad walking).

The reason i'm going to do this is because of the climbing with remotes suggestion will require this rigidbody with gravity.

I'd close this pull request and keep your eyes peeled for an update in the future :)

Frostlock:
Ok, I thought you would like to keep it as a minimal example on how to implement gravity.
But up to you, I'll store it in case you change your mind :)

PR #115, opened 2016-05-30
fix(Controller): correct haptic duration argument

thesbros:
The first argument for device.TriggerHapticPulse is not strength, but
the duration of the pulse in microseconds.
This has been changed so ControllerActions.TriggerHapticPulse takes one
argument, the duration of the pulse in *milliseconds*.
There is not actually a haptic strength value in the OpenVR API. [See here.][1]

[1]:
https://github.com/ValveSoftware/openvr/wiki/IVRSystem::TriggerHapticPulse

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_ControllerActions.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_ControllerActions.cs
@@ -3,13 +3,11 @@ using System.Collections;
 
 public class VRTK_ControllerActions : MonoBehaviour {
     private bool controllerVisible = true;
-    private ushort hapticPulseStrength;
-    private int hapticPulseCountdown;
+    private float hapticPulseCountdown;
 
     private uint controllerIndex;
     private SteamVR_TrackedObject trackedController;
     private SteamVR_Controller.Device device;
-    private ushort maxHapticVibration = 3999;
 
     public bool IsControllerVisible()
     {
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: e7a4d0e084db9d1db72eb71f82b85523b6c6a367

---

thestonefox:
Have you tried the device.TriggerHapticPulse?

If you don't include it in an Update method, then it only happens for a split second at the intensity of the parameter you provide.

(well at least it did in SteamVR 1.0.8)

The purpose of my wrapper was to allow to provide a duration and an intensity. If device.TriggerHapticPulse works that way now then that's fine.

thesbros:
I was just trying to be informational. :confused:

thesbros:
The reason it happens for a split second is because the function takes microseconds, so it would barely be noticeable. It is still called in the Update method, but there isn't actually any way to increase the intensity of the haptics, which is why I removed the strength value.

thestonefox:
You can increase the intensity by the number you put into device.HapticFeedback.

Try using my method and put in a low number for strength and then a high number (under 4000). Leave the time the same.

You'll get a different intensity of vibration.

As I say, unless they've actually changed this that's how it was working (physically actually working, not what their documentation says)

thesbros:
I guess you are correct. In my testing I didn't notice any difference when I changed the strength value, but apparently the haptic motor on one of my controllers is janky. The other controller I can notice the difference on.

thestonefox:
Crazy how the OpenVR docs say one thing and it does something else.

I pulled my hair out reading those docs :/

thesbros:
:confused: I created an issue on the OpenVR repository. ValveSoftware/openvr#129

thestonefox:
I considered being really active on their repos when I first discovered the amount of bugs in their stuff...then I quickly saw they hardly accept pull requests (or fix things!)

That's why I started this repo, just to get around that :)

PR #131, opened 2016-06-05
feat(Interaction): improve grabbable detection

10FingerArmy:
This is a first independent part of a bigger change to support rigidbody based buttons. So far when enabling the rigidbody feature of the InteractGrab script it was checking for interactable objects. This is too broad. It should only check for grabbable objects. Also the rigidbody on the controller should freeze all constraints to eliminate strange side effects due to a spinning collider.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
@@ -124,6 +124,7 @@ public class VRTK_InteractGrab : MonoBehaviour
         rb.useGravity = false;
         rb.mass = 100f;
         rb.collisionDetectionMode = CollisionDetectionMode.Continuous;
+        rb.constraints = RigidbodyConstraints.FreezeAll;
         ToggleRigidBody(false);
     }
 


GitCop:

 * Commit: bb44c8eb3247ffce8dfd3e4ff2c7ee7a2fc1a79c

---

thestonefox:
Hi, thanks for taking time to make a PR.

The issue is though you now have grab related code inside the touch code.

The InteractTouch should not know about the concept of grabbing.

The IsValidGrab check is checking to see if the object is interactable because without that first check there is never a touchable object and you cannot grab an object that cannot be touched.

Perhaps the wording of IsValidGrab is too ambiguous, but it basically means Is this item something I can grab because it's touchable.

Also, freezing rigidbody rotations/positions causes havoc with the physics engine if you're holding objects using fixed joints, I wouldn't recommend this practice or you get some crazy physics where items fly off to never be seen again!

10FingerArmy:
Misunderstanding with the freezing I think. You can easily see the wrong behaviour when selecting the controller in the hierarchy. Then the sphere collider will be shown. Now gently touch another rigidbody so that it moves a bit. All of a sudden the sphere collider on the controller starts to rotate. Now you get all kinds of weird effects. Must not happen.

thestonefox:
@10FingerArmy Ah you're talking about the invisible rigidbody that gets created to mimic having the controllers push items around, yeah?

If you freeze it's position and rotation does it still track the controller correctly? If so then I don't see an issue with freezing that because it's just invisible anyway.

10FingerArmy:
I reduced the change to the freeze fix only for now until I have some more time to incorporate your other hints

thestonefox:
Cool, I'll give it the once over in the morning and if everything is good I'll merge it 👍

thestonefox:
I redid the rigidbody code and included this change in the rework (this wont merge now, but it's also redundant because of the other commit).

Commit for reference: 2da56bd91c0b54a4122169867988128816c758f3

PR #135, opened 2016-06-06
feat(Interaction):support for Steam Rig Prefab

vad710:
This allows the code to work with both the original Steam prefab and the prefab included with the code.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
@@ -92,7 +92,12 @@ public class VRTK_InteractGrab : MonoBehaviour
         //If no attach point has been specified then just use the tip of the controller
         if (controllerAttachPoint == null)
         {
-            controllerAttachPoint = transform.GetChild(0).Find("tip").GetChild(0).GetComponent<Rigidbody>();
+            var attachObject = transform.Find("Model/tip/attach");
+
+            if (attachObject != null)
+            {
+                controllerAttachPoint = attachObject.GetComponent<Rigidbody>();
+            }
         }
 
         if (GetComponent<VRTK_ControllerEvents>() == null)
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 6ce4413d6e5f6b147de2946cdfa75765cb8e3c6f

---

thestonefox:
The one reason I can think for why Valve have the other CameraRig prefab is because if you look at their test throw example scene, you can specify a rigidbody attachpoint within the editor.

These rigidbodies exist within the editor in the test throw example scene, meaning they're able to be dragged and dropped to set the attach point.

If you're using the standard CameraRig (not the one from the test throw scene). Then they have no attachpoints you can specify.

Is this correct? Or is the standard CameraRig the one that is broken?

Why oh why do Valve have 2 camera rigs?! :(

10FingerArmy:
How I understand it one can freely exchange controller models at any point in time. Having this in mind it is actually mandatory to prepare for that and dynamically find the attach points instead of using them from the initial setup which is only valid for one controller. Or are the attachments the same for every controller?

thestonefox:
I'm not sure how their controller models work. They seem to only support their official vive models which seem to be streamed in via the library. They are supposed to have examples that shows how other models are supported but they don't seem to work.

I'd love to understand this to improve the current situation, because it sucks not being able to use the default the camera rig.

vad710:
They published a plugin for the Hydra and I believe it has the Hydra models - but I do not have a hydra to test. I saw on reddit the other day that there is a leap motion plugin, but not sure how they deal with the model. I have been pushing the folks at Ximmerse to produce a plugin - if they do, this is something I could test as I have their hardware.

thestonefox:
@vad710 that would be good if you could test on another hardware model.

Really want to get to the bottom of this CameraRig issue. Valve are tight lipped on it :(

vad710:
Did GitCop reject the commit because I missed the space between the ":" and the message?

thestonefox:
@vad710 yeah I think so

thestonefox:
This is superseded by PR https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/233

PR #151, opened 2016-06-07
UI Button Control

10FingerArmy:
First UI Control up for review


thestonefox:
I'm in two minds about how to approach this stuff.

Up until now, examples of things working (e.g. guns, swords, bows/arrows) have all be example scripts as they don't provide core functionality that would be generic across most games.

The thought process is, show people how to achieve something then they can copy/paste the code if they want or just use it as a guideline. Then they don't need to include the Examples folder in their project as it probably contains code they don't care about.

I wonder if a button (and other examples of interactions) would/should be examples rather than considered part of core functionality.

Originally the archery mechanics were considered core functionality, but it was separated out into the examples section as not everyone would want it.

Maybe worth getting the thoughts and opinions of some others to see which was is best to progress?

thestonefox:
Also...You have a merge commit in your PR. You don't want that, you just want your change commit :)

Let me know if you need any git assistance to sort it

10FingerArmy:
Fixed the merge commit issue and also changed the pushback logic slightly. Now it is possible to throw objects at the button and it will react properly.

thestonefox:
Cool :)

How do you envisage the Pushed event working?

What would happen if you had a number of buttons (they all throw the pushed event). How would you go about listening to the pushed event of specific buttons to know when they have been pushed?

10FingerArmy:
It would indeed be great to better understand each of our visions for the VRTK. To start of, let me quickly outline mine. I want to create multiple games in VR and I want to do that with minimal redundant efforts. 

That means I need to have the head tracking and controllers readily available. That is done by steam. Now I need to be able to react to the controllers. This is where the scripts here come into play. They provide a nice eventing solution. 

The next step is to interact with the world. Currently in VRTK three aspects of interactions are picked out: touching, grabbing and throwing (I don't understand using so far so I am ignoring it for now). That seems a bit random but is a nice foundation. 

Now, once you allow touching and grabbing you nearly immediately get into manipulating not only free floating objects but also those somehow connected to the environment, somehow attached and not throwable, e.g. doors, buttons, shelfs, drawers etc. In my opinion, if something is called VRTK it should definitely not stop here since this will be required so many times. 

And just like with Unity 2D UI controls there will be standards of interaction and not everybody will do their own because frankly, getting one single UI control right can be quite a pain. And I see us in that role. We are laying the foundation here to something that I will in the long-term future expect natively from Unity: VR UI. If we can share that load of fine-tuning then this will be tremendous value to everybody. 

That does not mean we should implement each and every control but the basic building blocks which can then be assembled to bigger ones and extended.

thestonefox:
@10FingerArmy I'm inclined to agree with you on having default controls part of the main repo. They are different enough from examples like swords, bows, guns, etc.

They're more a collection of widgets that people would use more often than not.

Let's see if anyone else chirps in, but if not then I'm cool with this approach.

Also, the Use interaction is when you want to trigger an action on an item. So if you look at the gun example, when you are touching the gun, the grab state picks the gun up, and the Use state causes the gun to do a Fire operation, which instantiates a bullet and propels it from the gun.

Other good examples of Use, are when you can toggle the WhirlyGig cubes on or off (even with using the pointers) so you could have like a Charging Station that you need to keep using to increase your health (or something like that).

10FingerArmy:
You had a good point there with the push event. Quite embarrassing. Fixed it and enhanced the example.

thestonefox:
I'm trying to check out the example scene but I'm having a couple of issues (one is really weird!)

Firstly, the buttons don't seem to have any materials on them:

image

Secondly, the strangest thing. When I run, the Game tab is just grey and when I look in the HMD the world is upside down. Can't see why it would be doing this, are you using the SteamVR CameraRig or the one from the toolkit?

10FingerArmy:
I reassigned the material and they should be red now. I seems I had another identically named material in an outside folder which was very unobvious :-/ 

Regarding the upside-down. While that might be a fancy effect, it is not intended indeed. I don't have it at least. I also did not change the rig. I basically saved another example scene (005 I think) and only added the buttons. Do you have any image effects going on? Some are broken and might cause this. I think also anti aliasing in quality settings did that in the past.

thestonefox:
I deleted the camera rig in the scene and re-added it back from the SteamVRUnityToolkit prefabs folder and it worked fine. I'll pull down your latest version and see if I have any issues with that.

thestonefox:
Ok So the same issue on the new build, But I've noticed that on the Camera (head) object there is no SteamVR_GameView script attached. Are you using Unity 5.4 beta?

I think Unity 5.4 beta doesn't use the GameView as it's been deprecated, but 5.3 still needs it.

Also, on Camera (eye) there is no SteamVR_CameraFlip script (which is what is causing the issue).

I wonder if Unity 5.4 is removing these scripts permanently? This is assuming you're using 5.4?

10FingerArmy:
Yes I am indeed on 5.4b20. How should we proceed? Could you amend the PR with the fixed scene?

thestonefox:
Hmm, can you try adding the SteamVR_CameraFlip script back onto the Camera (eye) object, then run it and see if that script is being removed from the object?

I don't think I can change your PR because it's on your fork.

Also, few points:

Firstly, works really well, :+1: 

Secondly, the right slider button can be slid off the table (does it have the option to provide limits of how far it can go?)

Thirdly, the yellow sphere may be better being swapped out for something that spawns a new object that perhaps floats up (like in the NewtonVR scene) because that yellow sphere gets big quickly and you can't see the buttons anymore.

Finally, so with the events, the way you hook up the button to another object (for it to activate it) is you add a script to the button that has a reference to the object you want to interact with? So in effect, you could have a button, then another script called "ActivateDoor" which has a door object (e.g the one from the UsingADoor scene) and the script calls it's StartUsing method, which means the button would open and close the door? That's my understanding of how you're expecting it to work, correct me if I'm wrong, but if I'm right seems like a good idea.

thestonefox:
I've broken your PR with the latest commit (hopefully it won't be too much of an issue to resolve the conflicts)

10FingerArmy:
Don't worry, I'll manage :-) That will be next week though as I am going onto a short vacation.

thestonefox:
Ok cool :) I think it's only the Readme that is causing the conflict so no big deal.

10FingerArmy:
I rebased now onto the latest changes and added the namespace as well as a better button push action. To the open questions:

* Limiting the button push area I would like to do in a follow up and stick to MVP here, I already have quite a lot of cool ideas how to simplify and continue
* Camera Flip: that is indeed removed automatically in 5.4 with the message below. I see two options now: I revert my project to 5.3, I just tried that actually and I get a weird error in SteamVR_ExternalCamera that LightProbeUsage does not exist which is correct since it is 5.4 stuff. Does anybody else get this? On the other hand this is rather a pain and 5.4 is in the final betas I feel and Vive support is built in so I'd much rather continue there. Alternative: we take this PR in and someone does a small "5.3 correction" commit readding the missing flip script.

SteamVR_GameView is deprecated in Unity 5.4 - REMOVING
UnityEngine.Debug:Log(Object)
SteamVR_GameView:Awake() (at Assets/SteamVR/Scripts/SteamVR_GameView.cs:14)
UnityEditor.DockArea:OnGUI()

thestonefox:
I'm happy to have this as the MVP and update limits in the future.

The camera flip one is a bit annoying. Maybe we'll go with your option of merging and fixing. Best to wait for my vive to be back in working order so I can test things. Hopefully I won't be without it for long!

10FingerArmy:
I will try to setup a different workflow today, having 5.3 to develop VRTK and 5.4 to use it in my project. That should make me compatible again with the community. New patch will come soonish.

thestonefox:
You can have 5.3 and 5.4 installed by at the same time. They don't clash with each other.

10FingerArmy:
Ok, now the PR should finally work nicely with 5.3 and the latest updates inside the framework. Would be good to get the basic structure in rather soon as I plan to build multiple components now based on this one.

10FingerArmy:
Just updated the PR with the newest and much evolved state. Now contains 4 controls and editor scene view integration.

thestonefox:
Just tried out the latest work, looking really good!!

Few points:

 * When spawning the items in, I'd add a "Destroy" for about 5 seconds maybe? perhaps if they're never interacted with? otherwise the scene can get crazy cluttered and drop fps

 * the buttons could do with some surrounding casing so you can tell which way they can be pushed. I didn't immediately know which way some were supposed to go to start with.

 * The Knob that works on the Y direction felt odd, I'm guessing it works on rotating the controller, whereas my initial thought was i could grab it and like spin it with the controller, but it didn't do what I expected.

 * The levers work well, but i noticed like a sliding motion when you let go of them, like they're lerping into position maybe? i was expecting it to stay put when I let go.

10FingerArmy:
I fixed 1,2 and 4 (since it's physics based and does not yet support snapping it is natural to have a small slide to it but snapping will eliminate that in the future). 3 I suggest to keep for now.

thestonefox:
In regards to the button surrounds. I had something more in mind like the below image. Where each button is encased to show it's limits. (Also it looks as if the right slider button was changed into a normal press button?)

image


Another thing to consider is setting the ButtonTable gameObject scale to 1,1,1 as having a non-uniform scale means all of the child object scales (and rotations) get skewed. I would say have the scales as 1,1,1 all the way down till you're sizing the button (then change the button size and position to remove skewing issues).

I would also suggest changing the spawn position of the balls, I got into a situation where they would just continuously spawn because i pressed the button, then they would spawn a new one that fell on the button, rinse and repeat! :)

10FingerArmy:
I added a button frame to the buttons on the table. The ones on the wall I propose to leave as they are as otherwise it becomes harder to see the auto-detection in action. 

I also added a nicer spawn area so it feels a bit like a slot machine now.

The scale is something that cannot easily be changed. All example scenes suffer from that since we are using Unity cubes. The only trick that works is to unparent them and put them all in the main hierarchy. I didn't want to do that to have a clean hierarchy. People can still unparent them however they like. To overcome that we would have to either model the tables in an external program or use ProBuilder but both I think are not worth it. In the end these are "only" samples.

thestonefox:
This is pretty much ready now! Really loving the work.

The one last thing to do is to update the Readme with your new Example scene in the Examples Section between these lines:

```
024_CameraRig_ExcludeTeleportLocation: A scene that shows how to exclude certain objects from being teleportable by either applying a named Tag to the object or by applying a Script of a certain name. In the scene, the yellow objects are excluded from teleport locations by having an ExcludeTeleport tag set on them and the black objects are excluded by having a script called ExcludeTeleport attached to them. The ExcludeTeleport script has no methods and is just used as a placeholder.

<<Your example description here>>

026_Controller_ForceHoldObject: A scene that shows how to grab an object on game start and prevent the user from dropping that object. The scene auto grabs two swords to each of the controllers and it's not possible to drop either of the swords.
```

After this then it's ready to merge if you're happy with it!

10FingerArmy:
Thanks! I added the examples entry, was not even aware of the section. Thanks for pointing that out (and keeping the 25 reserved ;-) )

thestonefox:
Cool, this is ready for merge then!

As a side note, did you ever see this repo?
https://github.com/JScott/ViveGrip (https://www.assetstore.unity3d.com/en/#!/content/62324)

May be some good inspiration in there?

PR #154, opened 2016-06-08
feat(Controller): disable walking per controller

drifio:
Adds a property that allows you to toggle whether or not the left or
right controller will allow walking.


thestonefox:
You gots some crazy commits going on there. Need to clean out the two commits before your disable walking per controller and force push that bad boy

drifio:
Yeah I just noticed that just now, I think I fucked something up D:

How do I go about getting rid of those extra commits?

thestonefox:
urm... not sure how they got in there. I think you may need to make sure your local branch is up to date (git fetch your upstream then rebase into master) then make sure your commit is above everything else. Those other commits are quite old and will have been superseded by the master branch

PR #155, opened 2016-06-08
feat(Controller): disable walking per controller

drifio:
Adds a property that allows you to toggle whether or not the left or
right controller will allow walking.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_TouchpadWalking.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_TouchpadWalking.cs
@@ -1,13 +1,45 @@
 ﻿using UnityEngine;
 using System.Collections;
 
-public class VRTK_TouchpadWalking : MonoBehaviour {
+public class VRTK_TouchpadWalking : MonoBehaviour
+{
+    [SerializeField]
+    private bool leftController = true;
+    public bool LeftController
+    {
+        get { return leftController; }
+        set
+        {
+            leftController = value;
+            SetControllerListeners(controllerManager.left);
+        }
+    }
+
+    [SerializeField]
+    private bool rightController = true;
+    public bool RightController
+    {
+        get { return rightController; }
+        set
+        {
+            rightController = value;
+            SetControllerListeners(controllerManager.right);
+        }
+    }
+
     public float maxWalkSpeed = 3f;
     public float deceleration = 0.1f;
 
+    private SteamVR_ControllerManager controllerManager;
     private Vector2 touchAxis;
     private float movementSpeed = 0f;
     private float strafeSpeed = 0f;
+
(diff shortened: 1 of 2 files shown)


Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_TouchpadWalking.cs
```diff
@@ -21,14 +50,17 @@ private void Awake()
         {
             Debug.LogError("The VRTK_TouchpadWalking script requires the VRTK_PlayerPresence script to be attached to the [CameraRig]");
         }
+
+        controllerManager = GameObject.FindObjectOfType<SteamVR_ControllerManager>();
```
thestonefox: The touchpad walking script is on the CameraRig so you don't need to do this.

`this.GetComponent<SteamVR_ControllerManager>()` would work here because they're on the same game object.

Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_TouchpadWalking.cs
```diff
@@ -91,16 +123,33 @@ private void FixedUpdate()
...
+        }
+        else if (controllerEvent && !toggle && subscribed)
         {
-            var controllerEvent = controller.GetComponent<VRTK_ControllerEvents>();
-            if (controllerEvent)
-            {
-                controllerEvent.TouchpadAxisChanged += new ControllerClickedEventHandler(DoTouchpadAxisChanged);
-                controllerEvent.TouchpadUntouched += new ControllerClickedEventHandler(DoTouchpadUntouched);
-            }
+            controllerEvent.TouchpadAxisChanged -= new ControllerClickedEventHandler(DoTouchpadAxisChanged);
```
thestonefox: I have a feeling this can unregister the last listener on that event (even if it's not the one you're expecting).

I (think) the solution is to store your listener you're adding to a variable then subtract that stored listener to prevent this, I should probably read documentation more :p

thestonefox:
Tested. all works fine.

One last thing to do is to update the README.md with the new parameters. something like:

```

The following script parameters are available:

  * **Left Controller:** If this is checked then the left controller
  touchpad will be enabled to move the play area. It can also be
  toggled at runtime.
  * **Right Controller:** If this is checked then the right controller
  touchpad will be enabled to move the play area. It can also be
  toggled at runtime.
  * **Max Walk Speed:** The maximum speed the play area will be moved
  when the touchpad is being touched at the extremes of the axis. If
  a lower part of the touchpad axis is touched (nearer the centre)
  then the walk speed is slower.
  * **Deceleration:** The speed in which the play area slows down to
  a complete stop when the user is no longer touching the touchpad.
  This deceleration effect can ease any motion sickness that may be
  suffered.
```

PR #163, opened 2016-06-12
fix(InteractGrab): add support for kinematic objects

bjennings76:
I noticed objects that were set to kinematic were not properly supported. I added a 'wasKinematic' variable that is set when an object is grabbed and then used to restore the state when the object is released.

Also, both types of grabs (SetControllerAsParent and CreateJoint) now explicitly set the kinematic state so they work properly when the object was originally set to kinematic.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
@@ -163,6 +163,9 @@ namespace VRTK
 
         private void CreateJoint(GameObject obj)
         {
+            Rigidbody rb = obj.GetComponent<Rigidbody>();
+            if (rb) { rb.isKinematic = false; }
+
             if (obj.GetComponent<VRTK_InteractableObject>().grabAttachMechanic == VRTK_InteractableObject.GrabAttachType.Fixed_Joint)
             {
                 controllerAttachJoint = obj.AddComponent<FixedJoint>();
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
@@ -210,8 +210,6 @@ namespace VRTK
         private Rigidbody ReleaseParentedObjectFromController()
         {
             var rigidbody = grabbedObject.GetComponent<Rigidbody>();
-            grabbedObject.transform.parent = null;
-            rigidbody.isKinematic = false;
             return rigidbody;
         }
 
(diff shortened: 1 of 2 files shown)


Assets/SteamVR_Unity_Toolkit/Scripts/Helper/DeviceFinder.cs
```diff
@@ -35,9 +35,10 @@ public static SteamVR_TrackedObject TrackedObjectByIndex(uint controllerIndex)
         public static Transform HeadsetTransform()
         {
 #if (UNITY_5_4_OR_NEWER)
-        return GameObject.FindObjectOfType<SteamVR_Camera>().GetComponent<Transform>();
-#endif
+            return GameObject.FindObjectOfType<SteamVR_Camera>().GetComponent<Transform>();
+#else
```
thestonefox: this isn't required.

If the first if is successful then the return happens. If it's not successful then the default return happens.

This doesn't fix anything, just does the same thing in a slightly different way.

thestonefox:
thanks for the PR.

It needs tidying up a bit: 

  * commit 25d91e3 is not required as it doesn't do what it's suggested, there is no unreachable code
  * commits 87f8b33 and 	627da9c should really be part of the same commit as they are fixing the same issue
  * commit e7f015e should not be there.

Also, I wonder whether the kinematic state of an object should be stored on the object itself and not the object doing the grabbing. Potentially the same for the parent.

Should it be the responsibility of the interactable object to remember it's own state? If the controller is doing it then this leads to potential issues down the line where maybe a controller can hold 2 objects but only ever know the state of one.

thestonefox:
I've raised a pull request to implement the same outcome but using the interactable object to hold the state: See here:

https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/165

Check it out and see if it works for your needs? Feels like a more appropriate way of solving the problem.

Assets/SteamVR_Unity_Toolkit/Scripts/Helper/DeviceFinder.cs
bjennings76: It does do it in a slightly different way, but it stops the 'unreachable code' warning that comes up each time the code compiles while using Unity 5.4+. (Notice that without the #else, the second line will still be in the compiled code, but be unreachable.

I can keep the change locally. I just thought anyone else using the beta might benefit from it as well.

thestonefox: No, you're correct, if the warning is throwing in Unity 5.4 then your fix is the right way to go.

However, I'd push just this fix on it's own in it's own Pull Request. As the other changes in this PR may well be superseded by the other pull request I've referenced.

bjennings76: Will do!

bjennings76:
Your solution works perfectly for kinematic objects set to use 'Child_Of_Controller', but when I try to use it with 'Fixed_Joint', I'm not able to move it. I think the 'Fixed_Joint' path needs to turn isKinematic off somewhere along the line.

(See VRTK_InteractGrab.cs:178 of this PR)

thestonefox:
@bjennings76 ok i'll have a look and see what happens :)

thestonefox:
@bjennings76 I've updated it now to affect joint grabs and tracked object grabs.

Let me know if it's ok, then I'll merge it

bjennings76:
Closing this since all kinematic bugs are fixed in #165.

PR #165, opened 2016-06-12
fix(Interaction): save previous object state on grab

thestonefox:
The Interactable Object's previous parent and kinematic state is now
saved when the object is grabbed and reset back to it's original state
when it is no longed grabbed. Supports swapping hands without needing
to drop the item first.


bjennings76:
Works great in all modes! Thank you!

And now that I know how to submit little fixes and tweaks via Github, I have separated out a number of adjustments for ya. I'll submit them as a bunch of little pull requests, but feel free to pass if they aren't appropriate.

PR #166, opened 2016-06-12
fix(DeviceFinder): fix for compile waring in Unity 5.4+

bjennings76:
Here's that commit for the compile warning separated out as requested.


thestonefox:
just fix the commit message spelling `waring` should be `warning` then it's good to go :)

bjennings76:
Ha! Will do.

PR #167, opened 2016-06-13
fix(ControllerEvents): fix order of controller events

bjennings76:
Solves an issue where `TouchpadAxisChanged` occurred before `TouchpadTouchStart` and `TriggerAxisChanged` occurred before `TriggerPressed`. Now, these can only occur in the following frame if the value continues to change, so now you can be assured that a 'Start' event will occur before a 'Changed' event for initialization purposes. (I needed this to get the first change delta and start tracking changes at the right moment.)

This also keeps the 'Changed' events from happening on the 'End' event frame, which would often cause a value jump. (At least they would with my tests on `TouchpadAxisChanged`.)

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_ControllerEvents.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_ControllerEvents.cs
@@ -304,28 +304,6 @@
             Vector2 currentTriggerAxis = device.GetAxis(Valve.VR.EVRButtonId.k_EButton_SteamVR_Trigger);
             Vector2 currentTouchpadAxis = device.GetAxis();
 
-            if (Vector2ShallowEquals(triggerAxis, currentTriggerAxis))
-            {
-                triggerAxisChanged = false;
-            }
-            else
-            {
-                OnTriggerAxisChanged(SetButtonEvent(ref triggerAxisChanged, true, currentTriggerAxis.x));
-            }
-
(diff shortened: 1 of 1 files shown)


Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_ControllerEvents.cs
```diff
@@ -337,6 +315,17 @@ void Update()
                 OnTriggerReleased(SetButtonEvent(ref triggerPressed, false, 0f));
                 EmitAlias(ButtonAlias.Trigger, false, 0f, ref triggerPressed);
             }
+            else
```
thestonefox: do these need to be in the else? would this mean they don't trigger on the initial press or on the final release?

bjennings76:
Avoiding trigger on the initial press and final release was intentional.

**First Press**: The value hasn't technically changed yet. It's just being set initially.

**Final Release**: Originally, I included the change event and release event on the same frame, but I always got back weird values on that final release frame. To get my features to work correctly, skipping the final frame worked best.

To see the problem, try out this example. Grab an object, then slide your thumb up/down along the TouchPad to change the object's scale. Notice how it jumps around both at the beginning (because the initial scale hasn't been set) and then again when you release (because of the sudden jump in the touchpad axis). With my fix, it runs smoothly. If you remove the second else that keeps the change from running on the last frame, you will see the jump at the end of the scale swipe come back again.

[touchpad scale example.zip](https://github.com/thestonefox/SteamVR_Unity_Toolkit/files/311841/touchpad.scale.example.zip)

PR #169, opened 2016-06-13
feat(Interaction): add RequireComponents to grab and touch

bjennings76:
Since `InteractGrab` and `InteractTouch` don't work unless certain
components are a part of the controller, the `RequireComponent`
attributes will make sure those components are there or automatically
add them. They also keep required components from being removed in
the Unity editor.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
@@ -16,6 +16,7 @@ namespace VRTK
     using UnityEngine;
     using System.Collections;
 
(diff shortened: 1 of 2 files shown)


thestonefox:
That's neat, didn't know about that.

So just to confirm, it's better to do this (and apply a bit of runtime magic) than to throw an error and get a user to go and fix the problem themselves?

bjennings76:
[`RequireComponent`](https://docs.unity3d.com/ScriptReference/RequireComponent.html) is a pretty standard Unity technique for making sure your game object has the pieces you require. The magic all happens at editor time, though, when the user adds the component.

The Unity Editor will automatically add on required components for you, but it is still possible to have a required component get deleted programmatically or have the component added to an object programmatically without it's required components in tow, so checking for their existence and throwing an error is definitely good housekeeping. :)

PR #170, opened 2016-06-13
feat(InteractableObject): add a rigid body if missing

bjennings76:
If no rigid body is found when `InteractableObject` starts up, it will
add one, but make sure the rigid body is set to 'kinematic' so it
doesn't start unexpectedly rolling around, popping up, or falling.


thestonefox:
I guess this makes sense too, I don't think an error is thrown if you don't have one does it?

It just doesn't work? Which is wrong, so this is valid

bjennings76:
Correct, there's no error currently.

I was originally considering the `RequireComponent` attribute but thought this would be better since `RequireComponent` doesn't support a way to control the default Rigidbody settings. This way, we can set `isKinematic = true` to avoid those unexpected side effects mentioned above.

PR #172, opened 2016-06-14
chore(Configs): add editor configs for VS and ReSharper

bjennings76:
These two files help me keep my code formatting straight between different projects and their varying style requirements. I thought perhaps they might help other collaborators keep to this projects specific styles too.

1. **`.editorconfig`** - Allows Visual Studio to change its tabs, spaces, and indent settings from project to project when used with the [EditorConfig plugin](https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328). See [EditorConfig.org](http://editorconfig.org/) for more info.
2. **`SteamVR_Unity_Toolkit.sln.DotSettings`** - Saves [per-project settings](https://www.jetbrains.com/help/resharper/2016.1/Sharing_Configuration_Options.html) for [ReSharper](https://www.jetbrains.com/resharper/) so they can be shared with all collaborators.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+; Used by the 'EditorConfig' VS plugin
+; See http://editorconfig.org
+
+; Use tabs preference set to 2 spaces per tab.
+[*.cs]
+indent_style = space
+indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,3 +5,4 @@
 [*.cs]
 indent_style = space
 indent_size = 4
+charset = utf-8-bom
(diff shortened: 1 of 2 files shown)


rob100:
Adding a ".editorconfig" would be great. But I would suggest to set this two parameters too:
`end_of_line = lf`
`charset = utf-8-bom`
I had several times the problem that visual studio messed up with the line endings until i created a editorconfig with this settings.

bjennings76:
Hmm... I tried out `end_of_line = lf`, but after setting that, I start getting warnings of inconsistent line endings in Unity after editing files in VS. I think the problem is that my git repo is set up to handle line endings automatically, so on a Windows machine, I already have all my files set up using windows-friendly `crlf` locally. Changes going forward are then inconsistent with the rest of the file.

Since git is handling all of that for me, I think I'll keep it out.

I'll definitely add in `charset = utf-8-bom`, though!

PR #177, opened 2016-06-16
fix(Sword): update to #ifs to support Unity 5.1

bjennings76:
Unity 5.1 doesn't have support for 'collision.impulse', so we use
'collision.relativeVelocity' instead.


thestonefox:
I'm planning on re-doing how this sword works. The force magnitude thing is really flaky. I've done something else where it uses the velocity of the controller swing to determine how hard it's hitting and that seems to be a better feel, also you don't need to worry about rigidbody collisions with it either.

I need to update the InteractGrab script to report the controller velocities. I'm going to do it when I get my headset back up and working.

paulc755:
Ooh, sounds like that could be useful for a golf club or whatnot. Lets hope HTC hurries up.

thestonefox:
Should be resolved with PR https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/193

PR #187, opened 2016-06-19
chore(format): switch to text mode serialization

10FingerArmy:
In order to get better merge behavior with GIT it is recommendable
to use the "Force Text" mode in the Unity serialization settings.


thestonefox:
What is the benefit of using Force Text over using the .meta files that are generated with each asset?

10FingerArmy:
This is independent of meta files. Text serialization is actually something helping with/enabling merges. Git is based on text files, not binary files. Saving in text mode gives it and us a chance to merge instead of having to decide for one version.

thestonefox:
From reading this:

http://kihira.uk/unity-and-git-getting-them-to-play-nicely/

It seems that just having asset files as "Force Text" may not be enough to make things better. I'm guessing as they're binary at the moment that git just doesn't try and merge them and expects you to choose the version you're interested in.

But from reading that article, unless you're using the Unity merge tool then leaving it up to git to resolve the merge may actually make the scene worse.

I'm new to shared unity repos so I'm not sure of the best way to manage the Unity assets within git. But I'm concerned that if people think merging is working but it breaks the scene then that's more of a headache than having to choose the asset to go with.

Any input/insight on this would be useful!

10FingerArmy:
I used the yaml mergetool several times already and while it can be a bit clumsy to setup (but still under 5 minutes) it at least gives you a chance to save work. With the current setup this is not possible. I added the gitattributes file from the link you posted as this is a really nice addition to ensure we only get better and not worse!

thestonefox:
But if you don't use the yaml mergetool can doing a normal git merge with text asset files screw up the asset?

That's the thing I'm concerned about is someone doing an auto merge and it breaking something.

I'm happier for someone to break something because they chose the wrong asset (it's their fault then :p) rather than someone having something broke due to git auto merge.

PR #193, opened 2016-06-22
feat(Interaction): add haptic feedback coroutine

drifio:
Changes the TriggerHapticPulse method to call a coroutine rather than
performing the haptic feedback event in the update method. This allows
feedback events specified in time intervals, rather than number of
frames. There are now two TriggerHapticPulse methods.
TriggerHapticPulse(ushort strength) does not call a coroutine, and
simply triggers a single pulse of specified strength.
TriggerHapticPulse(ushort strength, float duration, float pulseInterval)
allows specification of strength, total pulse duration, and the interval
that the controller should pulse at for that duration.


thestonefox:
Linking to the issue related:
https://github.com/thestonefox/SteamVR_Unity_Toolkit/issues/159

Issue #189, opened 2016-06-20
Enhancement(RoomExtender #158): Requests for additions to the RoomExtender Scripts


appdevtk:
There are some requests for additions to the RoomExtender #158 scripts and I'll try to add them.

- [x] One thing that I think would really improve things would be perhaps if you press a button on a controller it prevents the movement tracking, so when you walk to the edge of your physical play area (and cannot extend anymore_ then pressing a button stops moving the play area when you walk, then you can walk back to the center of your room and release the button to walk some more.

- [x] along with what Stone suggests (completely agree with this btw), if possible it would be nice to show the extended playspace along with the Chaperone. This way we of course see our play area, and then exactly just how far it's extended, so you can build content to this new volume.
I wonder if it's as simple a n*Chaperone Size? (where n is the extend multiplier)

I'll take a look.

xEvoGx:
Just a further suggestion regarding using a button...

...but perhaps you'd press the button to _engage_ the room extender, as if you were pressing the *Sprint* button in a given game. Walk normally, press/hold button to 'run'.

At the very least make it a toggle: **Press to Extend/No Press to Extend**

appdevtk:
@xEvoGx Do you think this would be enough to show the extended area? 

![roomextender](https://cloud.githubusercontent.com/assets/7008526/16275194/4ec0f63a-38a9-11e6-9b87-cb12317bb998.png)

xEvoGx:
Oh that's perfect! Heck you probably could even get away with a red "Extended Chaperone" that mirrors the play-space chaperone. 

But yea that's awesome! :smile:

appdevtk:
I tried do create the chaperone, but keeping the normal chaperone was not easy. This renders a gizmo.

xEvoGx:
Yea no that's great...really just being able to build to any visual of the extended space is awesome.

Great work man!

xEvoGx:
Not sure if it's worth reopening in case it's something I'm doing/not doing, but simply loading the scene **028 ...RoomExtender**, I do not see the Room Extension Gizmo, **SelectedOnly** on nor off. 

Again nothing special, just loading the scene, so one would expect to see it no?

Thanks!

-Steven

PR #197, opened 2016-06-24
feat(RoomExtender): adds toggle, gizmo and controller example #189

appdevtk:
I have added the ability to toggle the script, control it
with the controllers and extended the example scene
to showcase the new usage.
See VRTK_RoomExtender_ControllerExample.cs

I also added a Gizmo to show the extended playArea,
see example (028)

See #189 for more information.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Examples/028_CameraRig_RoomExtender.unity
+++ b/Assets/SteamVR_Unity_Toolkit/Examples/028_CameraRig_RoomExtender.unity
@@ -192,35 +192,6 @@ MeshFilter:
   m_PrefabInternal: {fileID: 0}
   m_GameObject: {fileID: 11091544}
   m_Mesh: {fileID: 10202, guid: 0000000000000000e000000000000000, type: 0}
---- !u!1 &25986066
-GameObject:
-  m_ObjectHideFlags: 0
-  m_PrefabParentObject: {fileID: 161786, guid: 1bb80779de0d85b4bba906974dbaa477, type: 2}
-  m_PrefabInternal: {fileID: 806555681}
-  serializedVersion: 4
-  m_Component:
-  - 4: {fileID: 25986067}
-  m_Layer: 0
-  m_Name: button
-  m_TagString: Untagged
-  m_Icon: {fileID: 0}
-  m_NavMeshLayer: 0
-  m_StaticEditorFlags: 0
-  m_IsActive: 0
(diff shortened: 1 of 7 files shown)


thestonefox:
I think i know the issue now with start up. If you have the "Press to enable" flag on the controller example then the play area only moves when you press the button... But when the scene first loads, even if you're not pressing the enable button it still adjusts your position by default, which causes your play area to wizz out to the extents of the extended boundaries.

I would expect it to leave the play area in the original starting position unless i'm pressing the button to move around.

Would you expect this to be solved by the user just unchecking the "additional movement enabled" checkbox on the RoomExtender script? (As this does solve the problem).

Also, I'd use TouchpadPress instead of TouchpadTouch. just brushing my finger on the touchpad obviously triggers it, which can move my play area when I wasn't intending to do it.


Finally, I also don't seem to be able to move to the extents of the extended boundary (as visualised with the red gizmo border in the editor). It seems to show my limits as being probably twice as large as I can actually reach: settings are:

image


And the gizmo shows:

image

I can probably walk from the hinged box (and weird character joint thing) to about the far edge of my blue play area boundary, but I cannot reach the drawer that is on the other extended boundary.

thestonefox:
Just tested it again, all seems to be working as I expected now.

Anything else you want to do on it before merge?

appdevtk:
I've updated the Readme and removed the linearTransform movement. I think it's ready to go.

couba:
Great work with RoomExtender. This is probably easy but I can't figure it out. In my program I move my CameraRig to fixed locations. Now I'd like to use your script, but everytime I teleport the position of the extended room space is in different position depending on where I actually stand in the real room. What I need is that the extended room is centered at my destination. I think if I had a vector between center of the CameraRig and the extended room, it would do the job, but I don't know how to get it. Any  tips? Thanks

appdevtk:
There is a vector that keeps track of the CameraRig movement:
[HideInInspector]
	public Vector3 relativeMovementOfCameraRig = new Vector3();

This is used to render the gizmo in the correct place. You could look at the VRTK_RoomExtender_PlayAreaGizmo script.

Lets say you teleport to point X. I think you can teleport to 
X - vrtk_RoomExtender.relativeMovementOfCameraRig to get the correct position.

I hope this helps.

couba:
Thats exactly what I needed. But there should be a + in the equation. Thanks a lot

Issue #173, opened 2016-06-14
Custom Edge of Chaperone Events


xEvoGx:
This might be up there with _already available_ or _easy to implement yourself_, but then again maybe not?

I was looking for an easy way to detect when either the HMD or a controller is near the boundaries of the Chaperone, and up on that call an event (outside of the Chaperone grid simply appearing). 

I was going through the **Steam VR** code but couldn't find anything that pertains to this, so now I'm wondering if I'm blind or maybe the Chaperone stuff is black-boxed? (I don't see why that would be the case, so I'm going with blind if nothing else) lol.

Cheers

-Steven

sdouglas:
Hi Steven - I'd be interested to know what application you have in mind because I was working on something similar this morning.

One simple thing you can do is use the locations of the camera and playarea to get a rough idea of where the player is standing. For example, I used this information to modify `VRTK_WorldPointer.SetPlayAreaCursorTransform` to center the pointer on where the player is standing and offset the room bounds so you have a better idea where they will end up.

```
playArea = GameObject.FindObjectOfType<SteamVR_PlayArea>();
hmdCamera = GameObject.FindObjectOfType<SteamVR_Camera>();
Vector3 roomPos = new Vector3(playArea.transform.position.x, 0, playArea.transform.position.z);
Vector3 hmdPos = new Vector3(hmdCamera.transform.position.x, 0, cam.transform.position.z);
Vector3 offset = roomPos - hmdPos;

playAreaCursor.transform.position = destination + offset;
```

I think this is similar to the hinting provided in The Lab's teleport function, but they also include a circle to make it even more obvious where the player is standing:

![teleport](https://cloud.githubusercontent.com/assets/45583/16059229/f8c389f0-3236-11e6-9d79-ebe33876ca8d.jpg)

xEvoGx:
Ah Shawn I was thinking the exact same thing, though with Teleport, as Stone added that bit to restrict a teleport by play area collision! 

Your solution is elegant and exactly what I was looking for...

...as for _use_, I'm not entirely certain yet. It dawned on me that I may want to know when the HMD or controllers were near the Chaperone edge, and do something with that information. Something perhaps as subtle as vibrating the controller so having a blue boundary floating around you isn't necessary (I still think it is). Obviously we're speaking about the edges of safe play, so there may not be a ton that can be done, unless you have ideas?

I really appreciate the response Shawn thank you!

-Steven

[...]

PR #200, opened 2016-06-25
feat(Pointer): offset playAreaCursor to hint at relative player position

sdouglas:
Add offset to playAreaCursor relative to hmd position, per discussion in issue #173

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
@@ -97,7 +97,10 @@ namespace VRTK
 
         protected virtual void SetPlayAreaCursorTransform(Vector3 destination)
         {
-            playAreaCursor.transform.position = destination;
+            Vector3 playAreaPos = new Vector3(playArea.transform.position.x, 0, playArea.transform.position.z);
+            Vector3 headsetPos = new Vector3(headset.position.x, 0, headset.position.z);
+            Vector3 offset = playAreaPos - headsetPos;
+            playAreaCursor.transform.position = destination + offset;
         }
 
         protected virtual void EnablePointerBeam(object sender, ControllerInteractionEventArgs e)


Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
```diff
@@ -97,7 +98,12 @@ protected virtual void InitPointer()
 
         protected virtual void SetPlayAreaCursorTransform(Vector3 destination)
         {
-            playAreaCursor.transform.position = destination;
+            playArea = GameObject.FindObjectOfType<SteamVR_PlayArea>();
```
thestonefox: I would recommend doing this assignment in the Start() or Awake() method to reduce the need for it to be called every time the cursor is drawn (reduces overhead on a costly Find call).

If it doesn't work in Start() or Awake() then maybe do a check to see if it's already been set:

```
if (playArea == null) {
  playArea = GameObject.FindObjectOfType<SteamVR_PlayArea>();
}
```

Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
```diff
@@ -97,7 +98,12 @@ protected virtual void InitPointer()
...
+            hmdCamera = GameObject.FindObjectOfType<SteamVR_Camera>();
```
thestonefox: Use the VRTK.DeviceFinder.HeadsetTransform() method for getting the hmd as it handles the differences between unity 5.3 and 5.4

the method is: https://github.com/thestonefox/SteamVR_Unity_Toolkit/blob/master/Assets/SteamVR_Unity_Toolkit/Scripts/Helper/DeviceFinder.cs#L35

so this line would be (with a null check as well for performance):

```
if (hmdCamera == null) {
  hmdCamera = DeviceFinder.HeadsetTransform();
}
```

thestonefox:
Thanks for the PR, looks good! I've just left a couple of points to look into :)

sdouglas:
Great suggestions, thanks! I'll revise.

GitCop:

 * Commit: 0681644d8748d4858604860fd54524a4152cf618

---

thestonefox:
Close to being mergeable now :)

Just need to fix the commit message so it adheres to the gitcop rules.

Your original commit message would be fine, would be good to also include a longer description too.

sdouglas:
Ok, I revised my second commit message that violated your gitcop rule and used `git push --force` to overwrite it. But I realize now that it might have been cleaner to repeatedly revise the original commit, since you don't necessarily want to merge a bunch of incremental commits from my branch as I figure out what to do. Or do you mind having those commits as long as they conform to the formatting rules?

thestonefox:
I'd squash your changes so there is only 1 commit.

May be easier to just re-implement your changes in a new commit if you're not used to squashing git commits (it's only 3 lines really).

There's no need to have the history of the commit as the latest commit is pretty much the work you want to commit.

Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
```diff
@@ -28,6 +28,7 @@ public abstract class VRTK_WorldPointer : VRTK_DestinationMarker
         protected bool playAreaCursorCollided = false;
 
         private SteamVR_PlayArea playArea;
+        private SteamVR_Camera hmdCamera;
```
thestonefox: This is not used, best to delete it :)

sdouglas: Yikes, of course.

PR #231, opened 2016-07-03
feat(Interaction): added chest UI control

10FingerArmy:
This change will add the chest as a new 3D control. The control will
add physics and events to a set of three game objects making up a chest.
Extensive auto-detection, editor visualization and default
events are supported.

Error gizmos are more useful. All controls were also moved into the
3D folder.


thestonefox:
Nice addition, much prefer it to the standard hinge, you don't get weird physics artifacts when the hinges play up.

Couple of overall points (i'll add individual things as code comments).

1) commit message should be in present tense, so `add` instead of `added`

2) would it be possible to place the boxes maybe on a shelf around the existing blue walls (maybe at knee height?) in my tiny 1.6mx2m place space I can't reach the boxes at the back of the black table.

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Chest.cs
```diff
@@ -0,0 +1,227 @@
...
+
+    public class VRTK_Chest : VRTK_Control
+    {
+        public enum Direction
+        {
+            autodetect, x, z // no support for y at this point in time
+        }
+
+        public Direction direction = Direction.autodetect;
+        private float min = 0f;
```
thestonefox: one of your pet hates ;) , but could the private params be grouped at the bottom and the public params at the top as per the contrib guide? :)

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Chest.cs
```diff
@@ -0,0 +1,227 @@
...
+        public float max = 160f;
```
thestonefox: I wonder if this should be something more descriptive than max. Perhaps `maxRotation` ?

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Chest.cs
```diff
@@ -0,0 +1,227 @@
...
+        protected override void InitRequiredComponents()
+        {
+            InitBody();
+            InitLid();
+            InitHandle();
+        }
+
+        protected override bool DetectSetup()
+        {
+            if (lid == null || body == null || handle == null) return false;
```
thestonefox: requires braces here as per contrib guidelines.

keeps things consistent if all code blocks are in braces even if it's a one liner

so

```
{
  return false;
}
```

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Chest.cs
```diff
@@ -0,0 +1,227 @@
...
+                    subDirection = (handleBounds.center.x > lidBounds.center.x) ? -1 : 1;
+                    break;
+                case Direction.z:
+                    subDirection = (handleBounds.center.z > lidBounds.center.z) ? -1 : 1;
+                    break;
+            }
+
+            if (lid & handle)
+            {
+                if (handle.transform.IsChildOf(lid.transform)) return false; // handle should be outside lid hierarchy, otherwise anchor-by-bounds calculation is off
```
thestonefox: braces required

10FingerArmy:
All should been corrected now. New formatting feels like second skin already ;-)

thestonefox:
One thing I've noticed is when pulling on the handle, the joints can go crazy, like you can disjoint the handle and the lid hinge:

image


I think it may be ok to merge this now as a first pass but then look at seeing if there's a way to fix that joint behaviour

PR #237, opened 2016-07-05
feat(UI): add ability to interact with unity ui

thestonefox:
Unity provides a collection of UI elements that are traditionally for
use with a mouse or a touch screen. It is now possible to use these
elements within VR utilising the new `VRTK_UIPointer` script that can
be attached to any object such as the controllers or headset and it
casts a ray out to the UI elements for interaction.

The `Pointer` alias button on the associated controller determines if
the ray is being cast or not. The `Use` alias button on the
associated controller determines if the UI element has been clicked.

It works well in conjucntion with the Simple Pointer to provide visual
feedback to where the ray is hitting.

The UIPointer works by replacing the `Standalone Input Module` on the
`EventSystem` with a custom Input Module. It also finds all canvases
that are set to `World Space` and replaces the default
`Graphic Raycaster` with a custom one that can identify the rays
coming from the in game world objects.

The UIPointer script contains the following parameters:

  * `Controller` - The controller which will operate the pointer and
  use functions for the UI Pointer.

An example scene has been added to demonstrate the feature.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Examples/034_Controls_InteractingWithUnityUI.unity
+++ b/Assets/SteamVR_Unity_Toolkit/Examples/034_Controls_InteractingWithUnityUI.unity
@@ -1,0 +1,3502 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!29 &1
+SceneSettings:
+  m_ObjectHideFlags: 0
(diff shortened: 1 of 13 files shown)


GitCop:

 * Commit: 1f9b366b5e9c9ccb424ec2a8a252b628bc73a35b

---

PR #245, opened 2016-07-08
feat(Headset): add headset controller glance events - resolves #241

### LINKED ISSUE #241: Emit an event when a controller is looked at
It would be really useful to know when the controllers were being looked at by the headset so other actions could be triggered (like showing tooltips only on glance)

thestonefox:
The Headset Controller Glance script allows for an event to be emitted
when the headset is looking at either of the controllers.


thestonefox:
resolves #241

thestonefox:
Superseded with https://github.com/thestonefox/VRTK/pull/657

PR #249, opened 2016-07-08
Feat/controller apperance

thestonefox:


baemoth:
Tested, works flawless. 5.3 and latest 5.4

PR #259, opened 2016-07-11
fix(Locomotion): add null check for Play Area

rocky1138:
Game objects which are set to not destroy on level load and that contain a
VRTK_Bezier Pointer script will throw a null exception for the Play Area on the
next level load.

This commit fixes that by adding a null reference check to the Play Area.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
@@ -106,7 +106,7 @@ namespace VRTK
                 activateDelayTimer -= Time.deltaTime;
             }
 
-            if (playAreaCursor.activeSelf)
+            if (playAreaCursor && playAreaCursor.activeSelf)
             {
                 UpdateCollider();
             }


GitCop:

 * Commit: c22abe89cc81a80c526b26171638be162a60625e

---

thestonefox:
I'd change the scope from `Locomotion` to `Pointer` as the fix is actually addressing an issue with the World Pointer and could be completely separate from Locomotion issues.

Also i'd make Play Area lowercase.

So the commit head message should read:

`fix(Pointer): add null check for play area`

then it looks good to merge :+1:

rocky1138:
It turns out this may be a bit more work than a simple one-line fix.

I get more null objects after this one.

I'm going to close out this PR and see about making sure it's water tight.
There may be a better workflow than I am using or I may need to add more
null checks before I'm confident it's 100%.

On Mon, Jul 11, 2016 at 5:52 AM, StoneFox <notifications@github.com> wrote:

> I'd change the scope from Locomotion to Pointer as the fix is actually
> addressing an issue with the World Pointer and could be completely separate
> from Locomotion issues.
>
> Also i'd make Play Area lowercase.
>
> So the commit head message should read:
>
> fix(Pointer): add null check for play area
>
> then it looks good to merge 👍
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/259#issuecomment-231690475>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAi_g--MOniJsCVTIjQvfML2IVOFNT6cks5qUhJKgaJpZM4JI9kv>
> .
>



-- 
http://www.johnrockefeller.net

PR #274, opened 2016-07-16
feat(Interaction): add rotator track grab attach mechanic

thestonefox:
The `Rotator_Track` grab attach mechanic allows the Interactable Object
to rotate around a joint (usually hinge) to simulate grabbing and
spinning an item (such as a wheel on an axel).

This is heavily influenced by how NewtonVR handle rotation items.


thestonefox:
You need docs!

PR #275, opened 2016-07-17
feat(Teleport): Allow multiple tags and component types to be specified as invalid teleport locations

omgwtfgames:
This patch allows multiple tags and component types to be specified as invalid teleport locations. It also prevents a potential corner case bug that could have occurred when a tag had the same name as component class. 

Reduces some code duplication by putting some of the common code into static methods on VRTK_WorldPointer (I'm not sure this is the ideal place, it feels like there should be a dedicated 'teleportation target validation' component to do the handle the common business logic for both the pointer/marker renderers and any teleportation movers like VRTK_BasicTeleport).

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_DestinationMarker.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_DestinationMarker.cs
@@ -22,6 +22,7 @@ namespace VRTK
 {
     using UnityEngine;
     using System.Collections;
+    using System.Collections.Generic;
 
     public struct DestinationMarkerEventArgs
     {
@@ -42,7 +43,8 @@ namespace VRTK
         public event DestinationMarkerEventHandler DestinationMarkerExit;
         public event DestinationMarkerEventHandler DestinationMarkerSet;
 
-        protected string invalidTargetWithTagOrClass;
+        protected List<string> ignoreTargetWithTags;
+        protected List<string> ignoreTargetWithComponents;
         protected bool checkNavMesh;
         protected bool headsetPositionCompensation;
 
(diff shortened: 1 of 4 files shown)


GitCop:

 * Commit: af27d5c34dfd70b9a4f2eaff59d19764b36b2509

---

thestonefox:
What is the benefit to allowing for multiple tags? If you're using tags for something else, why not just apply a custom class to all the items you want excluding?

You can already have multiple classes on objects so wouldn't this achieve the same thing?

GitCop:

 * Commit: af27d5c34dfd70b9a4f2eaff59d19764b36b2509

---

omgwtfgames:
@thestonefox The benefit of allowing multiple tags and components is that it makes the teleportation location filtering more flexible when integrating into an established project. As you note, you may already be using the tags for something else and might not be able to easily change them all to a common 'NoTeleport' tag. Adding something like a custom 'InvalidTeleportTarget' component would do the trick, however I found myself in the situation with a bunch of existing prefabs that I didn't want to update with an additional component if I could avoid it.

With regard to potential bugs due to a collision between a tag name and a component class name - it's more defensive coding than anything. I'm not so much worried about the case where you remember that you have both a tag and a class with the same name and actively deal with it, but more the case where you aren't necessarily aware of a component in a third party asset that shares a name with your tag. It's unlikely to happen but in the case that it does it could become a tricky bug to chase down.

thacio:
The last comment was 19 days ago -_-
Anyway, I wanna comment on this as I just went throught similar situation.

 I think in order for the toolkit to be useful, it needs to be simple to use and flexible enough to be customizable when developing a game. With a single teleport tag (exclusive) or class name, you can't do much without digging through the toolkit code or knowing how it works to do some workaround.
So in my opinion the changes in this PR needs to be done so more people could use it out-of-the-box.

I've added a solution myself (haven't PR as I need to update from the main branch) as I still couldn't use it to my needs. It uses events so outside classes can implement their own validation logic, so more complex valdiation systems can be used out of the box. (this would be to give much more flexibility, doesn't exclude the tags/class name feature)

thestonefox:
This has been resolved with a new feature called PolicyLists

PR #283, opened 2016-07-19
fix(Pointer): activate delay timer for teleport

gpvigano:
Prevent multiple undesired teleport.

Even if the teleport activateDelayTimer is set, when users double-click
the teleport button, the teleport is activated twice.

Check the timer in VRTK_WorldPointer.SetPointerDestination() to prevent
this.

When the button is released let the beam be disabled immediately,
otherwise the beam cannot be activated again because of the timer and a
user feels like the button is not responding.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
@@ -162,9 +162,8 @@ namespace VRTK
 
         protected virtual void DisablePointerBeam(object sender, ControllerInteractionEventArgs e)
         {
-            if (isActive && activateDelayTimer <= 0)
+            if (isActive)
             {
-                activateDelayTimer = activateDelay;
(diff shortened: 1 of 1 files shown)


thestonefox:
This fix isn't working at all well for me.

The pointer is off for a lot longer than the time I enter in Activate Delay. Also the pointer will delay being activated again even if i dont teleport.

thestonefox:
I know what the problem is.

You can't set the activeDelayTimer in the `SetPointerDestination` because that is always called even if you have no target.

The solution is to do it in the actual `PointerSet` method.

I'll do a PR to show it and if it works for you then we can merge the other PR in and close this one.

thestonefox:
This is the other PR:

https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/287

gpvigano:
Yes, it works perfectly, please please merge your modifications, thank 
you very much.

StoneFox wrote:

thestonefox:
Superseded by https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/287

PR #250, opened 2016-07-08
feat (Interaction): make 3D controls expose UnityEvents

10FingerArmy:
Controls now expose UnityEvents which are neatly exposing their own
inspector UI so that it becomes very easy to wire controls together
through configuration instead of using code. Programmatic access is
still possible.

The events are wrapped in event container classed to fold them per
default in the inspector.

PR #286, opened 2016-07-20
feat(RadialMenu): add OnHoverEnter / OnHoverExit events to RadialMenu buttons

omgwtfgames:
This allows events to be triggered when the button enters/exits the hovered state.
An example use case is showing/hiding of an attached laser pointer component depending on which button is being hovered over.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/Controls/2D/RadialMenu/RadialMenu.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/Controls/2D/RadialMenu/RadialMenu.cs
@@ -48,7 +48,7 @@ public class RadialMenu : MonoBehaviour
         //Keep track of pressed button and constantly invoke Hold event
         if (currentPress != -1)
         {
-            buttons[currentPress].Press();
+            buttons[currentPress].OnHold.Invoke();
         }
     }
 
@@ -71,12 +71,13 @@ public class RadialMenu : MonoBehaviour
         {
             ExecuteEvents.Execute(menuButtons[currentHover], pointer, ExecuteEvents.pointerUpHandler);
             ExecuteEvents.Execute(menuButtons[currentHover], pointer, ExecuteEvents.pointerExitHandler);
+            buttons[currentHover].OnHoverExit.Invoke();
         }
         if (evt == ButtonEvent.click) //Click button if click, and keep track of current press
         {
             ExecuteEvents.Execute(menuButtons[buttonID], pointer, ExecuteEvents.pointerDownHandler);
             currentPress = buttonID;
-            buttons[buttonID].Click();
+            buttons[buttonID].OnClick.Invoke();
         }
         else if (evt == ButtonEvent.unclick) //Clear press id to stop invoking OnHold method
         {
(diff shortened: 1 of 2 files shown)


10FingerArmy:
I think it's good the radial menu is already using UnityEvents. I hope other parts of the framework will follow soon. I want to point out two things only:
* Why do you use wrapper methods instead of calling the .invoke() directly? That seems rather unnecessary and only creating boilerplate code
* While converting the 3D controls to UnityEvents in the still to-be-merged https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/250 I suggested to make all events foldable in the inspector. I think that improves the easy of use quite a lot since the UI for events is really huge and with more events the inspector becomes really long. What do you think of that? I think it would be nice to have a common way here.

omgwtfgames:
Thanks @10FingerArmy - useful observations. 

On point (1): I don't see any compelling reason not to just call the `Invoke()` methods directly, I just followed the existing pattern being used without thinking to much about it. The only argument I can think of is that the current wrapper methods would allow some other internal logic to execute pre/post the user subscribed UnityEvents. But that option currently isn't being used, and it wouldn't break anything in a future refactor if it was added back when required, so I'd vote for removing the wrappers. Maybe @thestonefox has a particular reason ?

On point (2): Making the events section foldable (via putting them all in a `Serializable` container class) is a good idea - the UI for events does take up considerable space in the Inspector, especially once you have several. If there are no objections I'll update the PR to do this.

omgwtfgames:
I've removed the wrapper methods as per point (2).

After testing the making the UnityEvent for each button foldable as per point (1), I decided it was better in this case to leave them as-is since the RadialMenuButtons themselves are already foldable since as list elements in the Inspector. I think in this case having two levels of nesting will just get annoying when assigning events to each button. In other parts of the code where a set of UnityEvents on a component would appear unfolded in the top level of the Inspector, I agree, put them in a container class and make them foldable.

I see this as done. Mergeable @thestonefox ?

10FingerArmy:
Good point about the already existing folding. +1 from my side which should also make the path clear for #250

thestonefox:
@omgwtfgames you've now got some merge conflicts you'll need to resolve (they shouldn't be too much of a big deal to fix, I'm guessing its from the refactor PR)

thestonefox:
I think your merge has screwed up as you've pushed a bunch of other commits. Did you rebase the changes in?

omgwtfgames:
Yes.

thestonefox:
Looks like it's a git pull with a merge rather than a git fetch with a rebase. Either way the pr is borked :/

You may find it easier to just make a new branch from master and bring your changes over into a new pr.

omgwtfgames:
Should be fixed now - SourceTree was preventing me from force pushing but I'd forgotten about the config option to allow it. Just delete any local working copy of the branch and pull down a new copy.

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/2D/RadialMenu/RadialMenu.cs
```diff
@@ -48,7 +48,7 @@ private void Update()
         //Keep track of pressed button and constantly invoke Hold event
         if (currentPress != -1)
         {
-            buttons[currentPress].Press();
+            Buttons[currentPress].OnHold.Invoke();
```
thestonefox: Looks like you're merge is overriding an old change.

`Buttons` was renamed to `buttons` 

Just change all instances of the variable locally in your commit and it and it should be fine and not show as a change.

thestonefox:
I've broken your PR again :( this time by taking out the documentation from `README.md` docs should now go into `DOCUMENTATION.md`

omgwtfgames:
No probs - this *should* be right now :stuck_out_tongue_winking_eye:

PR #288, opened 2016-07-21
fix(Pointer): switch off usage on pointer off

gpvigano:
If pointerActivatesUseAction is true in VRTK_InteractableObject, after
pointing at the object with the pointer, the object continues to be used
also when the pointer is switched off.

Add a check in VRTK_WordPointer.TogglePointer() to switch off the
object usage if the pointer is switched off (add also a reference to the
object as private class member to keep track of the pointed object).

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
@@ -46,6 +46,8 @@ namespace VRTK
 
         private float activateDelayTimer = 0f;
 
+        private VRTK_InteractableObject interactableObject = null;
+
         public virtual void setPlayAreaCursorCollision(bool state)
         {
             if (handlePlayAreaCursorCollisions)
@@ -179,7 +181,7 @@ namespace VRTK
 
             OnDestinationMarkerEnter(SetDestinationMarkerEvent(pointerContactDistance, pointerContactTarget, destinationPosition, controllerIndex));
 
-            var interactableObject = pointerContactTarget.GetComponent<VRTK_InteractableObject>();
+            interactableObject = pointerContactTarget.GetComponent<VRTK_InteractableObject>();
             if (interactableObject && interactableObject.pointerActivatesUseAction && interactableObject.holdButtonToUse)
             {
                 interactableObject.StartUsing(gameObject);
@@ -195,7 +197,6 @@ namespace VRTK
 
             OnDestinationMarkerExit(SetDestinationMarkerEvent(pointerContactDistance, pointerContactTarget, destinationPosition, controllerIndex));
 
-            var interactableObject = pointerContactTarget.GetComponent<VRTK_InteractableObject>();
             if (interactableObject && interactableObject.pointerActivatesUseAction && interactableObject.holdButtonToUse)
             {
                 interactableObject.StopUsing(gameObject);
@@ -239,6 +240,10 @@ namespace VRTK
         {
             var playAreaState = (showPlayAreaCursor ? state : false);
             playAreaCursor.gameObject.SetActive(playAreaState);
+            if (!state && interactableObject && interactableObject.pointerActivatesUseAction && interactableObject.holdButtonToUse && interactableObject.IsUsing())
+            {
+                interactableObject.StopUsing(this.gameObject);
+            }
         }
 
         protected virtual void SetPointerMaterial()


thestonefox:
I think the issue you're describing is actually a feature.

If the `Hold Button To Use` checkbox is ticked then the pointer needs to be on the object (and active) for the pointer to work.

If `Hold Button To Use` is unchecked then the pointer will turn it on with the first destination set, and turn it off with the second destination set.

Look at the `019_Controller_InteractingWithPointer` example to see this working.

gpvigano:
This fix is intended just when the Hold Button To Use is checked (notice 
the "&& interactableObject.holdButtonToUse" condition in my change to 
the TogglePointer() method), if you switch off the pointer (you stop 
holding button) the object continues to be "used". This happens if you 
set e.g. both "Pointer Toggle Button" and "Use Toggle Button" to 
"Trigger" (sorry for having omitted this detail before).

StoneFox wrote:

thestonefox:
Can you provide a video of the issue you believe you're seeing?

I've looked over the example scene and it seems to be working as I would expect.

If I hover my pointer over the cube that has `Hold button To Use` checked, then the spinner is only active when the pointer is on and touching the cube.

The other cube is correctly having it's using state toggled.

gpvigano:
You can test it loading the example 019_Controller_InteractingWithPointer and setting Trigger both as Pointer Toggle and Use Toggle Button:
vrtk_controller_settings
then point to the active object on the right "Whirlygig (2)", press the controller trigger:
vrtk_trigger_activate

then release it:
vrtk_trigger_release

The capsule will continue spinning even if the pointer was disabled.

thestonefox:
Cool, i'll try that out in a bit.

thestonefox:
I wouldn't say this is a bug, I'd say it's a misconfiguration.

Because of the way the code works, if you set the Pointer Toggle button and Use button to the same button (e.g. Trigger). but leave the Pointer Set button as something different then it's going to cause all sorts of problems because it's a bad configuration.

If in your example you change the "Pointer Set" button to "Trigger" as well then it works as expected.

Whilst your PR does seemingly fix the initial issue, it doesn't fix other issues that are caused by this bad configuration. So I'd say rather than sticking plaster one issue, it's better just to not configure the system to work in this way,

gpvigano:
This configuration was intended to activate objects with a kind of "laser shot".
Maybe this should be a different type of pointer (or a parameter could be added to enable this feature), but for now, if this is not allowed, probably it is better to point out in the documentation that we cannot set the same button to two different functions.

gpvigano:
I changed the button settings, but the problem seems to be the same: if the interactive object has "Pointer Activates Use Action" and "Hold Button To Use" enabled and you just disable the pointer while the object is activated "in use" it won't stop to be "in use".

thestonefox:
I'll do a video of my solution in a moment to make sure it's what you're expecting (and to show off the settings)

Maybe worth you doing a video too to show it not working?

gpvigano:
These are the changed settings:

vrtk_controller_settings_2

The result is the same, I have a small video in this ZIP (made using current version on master):
[VRTK 2016-08-02 10-24-08-54.zip](https://github.com/thestonefox/SteamVR_Unity_Toolkit/files/396152/VRTK.2016-08-02.10-24-08-54.zip)

thestonefox:
Shouldn't your settings be:

Pointer Toggle Button: Trigger
Pointer Set Button: Trigger
Use Toggle Button: Trigger

Or are you saying with the default settings it doesn't work?

thestonefox:
Yeah i agree your new settings shouldn't keep the item going.

Confirmed bug :)

Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
```diff
@@ -244,6 +245,10 @@ protected virtual void TogglePointer(bool state)
         {
             var playAreaState = (showPlayAreaCursor ? state : false);
             playAreaCursor.gameObject.SetActive(playAreaState);
+            if (!state && interactableObject && interactableObject.pointerActivatesUseAction && interactableObject.holdButtonToUse)
```
thestonefox: should you also check here if the interactable object `IsUsing()` before calling stop using?

If someone has some code in their StopUsing() call that relies on something that was in the StartUsing() and for some reason this object is no longer not using (i dunno how! but people are creative :p)

Then they may not want the stop using to be called (unless it's in the process of being used)

gpvigano:
Sorry, I need your help.
After rebasing to master I tried to  push and I got this error:

error: failed to push some refs to 'https://github.com/gpvigano/SteamVR_Unity_Toolkit'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

I Fetched from your repository, what should I Pull?

GitCop:

 * Commit: 7d6e769b977735b5d878c13dd089af68321965ce

---

gpvigano:
Sorry, finally I pushed using the "Force update" option and probably I broke something, could you check again this PR please?
Thanks

thestonefox:
Still looks ok to me, let me pull it down and try it locally

No idea what gitcop is moaning about :/

PR #303, opened 2016-07-26
fix(Interaction): react to enablement state of interactable_object

10FingerArmy:
So far disabling an interactable_object at runtime did not have any
effect as it could still be touched and grabed. This patch checks the
enablement state explicitly.

A fix in the sample scene was needed to remove a boxcollider which
made it nearly impossible to showcase the feature that disables the
drawer contents.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractTouch.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractTouch.cs
@@ -76,7 +76,23 @@ namespace VRTK
 
         public bool IsObjectInteractable(GameObject obj)
         {
-            return (obj && (obj.GetComponent<VRTK_InteractableObject>() || obj.GetComponentInParent<VRTK_InteractableObject>()));
+            if (obj)
+            {
+                VRTK_InteractableObject io = obj.GetComponent<VRTK_InteractableObject>();
+                if (io)
+                {
+                    return io.enabled;
+                }
+                else
+                {
+                    io = obj.GetComponentInParent<VRTK_InteractableObject>();
+                    if (io)
+                    {
+                        return io.enabled;
+                    }
+                }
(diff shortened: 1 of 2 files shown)


Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractTouch.cs
```diff
@@ -109,8 +124,8 @@ private void Start()
                 return;
             }
 
-            Utilities.SetPlayerObject(this.gameObject, VRTK_PlayerObject.ObjectTypes.Controller);
-            CreateTouchCollider(this.gameObject);
+            Utilities.SetPlayerObject(gameObject, VRTK_PlayerObject.ObjectTypes.Controller);
```
thestonefox: I've noticed pretty much all of the changes after this point are to do with removing `this` prefix (or other similar prefixes)

Is there a reason for this?

Personally, I like the `this` prefix for readability as it can tell this isn't a local variable but an inherited monobehaviour.

But, if there is a reason for it, then I think perhaps it should be a separate PR that does it across the board to ensure all classes are consistent.

thestonefox:
What were the changes to the example scene that were commited for this fix?

10FingerArmy:
I pulled the two changes apart.

thestonefox:
what's the purpose of the change to the example scene for this PR?

10FingerArmy:
I had described it in the commit message but as it seems the message is not updated automatically here. Weird. I just did that.

thestonefox:
Ah, so the drawer contents disables the interactable object? Therefore this change is demonstratable in that scene?

I can't test it obv, but I'll take your word for it and merge it if that's the case :)

10FingerArmy:
exactly, and there was a collider on the drawer that messed up the grabbing so it was really hard to try it out.

PR #304, opened 2016-07-27
chore(Code): remove unneeded scope references

10FingerArmy:
While looking through the code of touch and interactable object I
found many references to 'this' which is not a common practice (they are
simply provided by the extended MonoBehaviour), clutters the code and
causes Visual Studio to highlight these as unneeded making the code
very colorful.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractTouch.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractTouch.cs
@@ -8,7 +8,6 @@
 namespace VRTK
 {
     using UnityEngine;
-    using System.Collections;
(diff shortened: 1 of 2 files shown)


thestonefox:
I think if this is something we're going to do, it should be done everywhere and the contributing guidelines should be updated to ensure people stay consistent.

You can take a stab at doing it everywhere if you want, or I can do it when I'm back.

I think it's mostly going to be find and replace

10FingerArmy:
I think you can still use this change as it already touches two files, the work was done, and then do the remaining ones yourself as needed.

thestonefox:
This is superseded by https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/321

PR #317, opened 2016-07-29
feat(Interaction): add door 3D control

10FingerArmy:
This change adds the door control. A door consists of a door, handles
and an optional frame. Auto-detection works very reliably. Advanced
gizmos are included. Multiple options can be set already. More to come
in the near future.

While testing the door this use-case became apparent so 3D controls now
also support enablement and disablement at start and at run-time.

Utilities contain a much better bounds calculation now and a new
convenience method for collider creation.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Door.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Door.cs
@@ -1,0 +1,365 @@
+﻿namespace VRTK
+{
+    using UnityEngine;
+
+    public class VRTK_Door : VRTK_Control
+    {
+        public enum Direction
+        {
+            autodetect, x, y, z
+        }
+
+        public Direction direction = Direction.autodetect;
+        public float maxAngle = 120f;
+        public bool openInward = false;
+        public bool openOutward = true;
+        public bool snapping = true;
+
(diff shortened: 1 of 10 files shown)


Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Door.cs
```diff
@@ -0,0 +1,365 @@
...
+            }
+
+            handleIo = handles.GetComponent<VRTK_InteractableObject>();
+            if (handleIo == null)
+            {
+                handleIo = handles.AddComponent<VRTK_InteractableObject>();
+            }
+            handleIo.isGrabbable = true;
+            handleIo.precisionSnap = true;
+            handleIo.grabAttachMechanic = VRTK_InteractableObject.GrabAttachType.Spring_Joint;
```
thestonefox: try this instead:
`handleIo.grabAttachMechanic = VRTK_InteractableObject.GrabAttachType.Track_Object;`

I feel the `Track_Object` mechanic gives a better grab feel on the door. The rotator track may work even better, but I couldn't get it to work on the tiny door (under the chest)

10FingerArmy:
I tried it and it works indeed better! Good suggestion. New patch online.

thestonefox:
The real MVP - :heart: :green_heart: :yellow_heart: :blue_heart:

PR #324, opened 2016-07-31
feat(Testing): add movement simulator resolves #262

### LINKED ISSUE #262: How can I test this without a headset ?
Can there be a mock headset or something so that when I am other places or do not want to stand up and use the headset I can just use the keyboard to move the headset/controllers around ?

10FingerArmy:
For many trivial tests one still has to put on the headset in order to
move to the right spot. This increases turn-around times. This script
can be put onto the camera rig and will then allow movement through
the scene using keys. Once at the desired location the controllers can
be used to interact while looking at the monitor.

Supported so far are move, strafe, rotate and up/down. Being an early
version this is only tested in one context so far and especially after
rotating the move directions are inverted currently. These are all
things to look at in future versions. Help always welcome.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_Simulator.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_Simulator.cs
@@ -1,0 +1,109 @@
+﻿namespace VRTK
+{
+    using UnityEngine;
+
+    public class VRTK_Simulator : MonoBehaviour
+    {
+        [System.Serializable]
+        public class Keys
+        {
+            public KeyCode forward = KeyCode.W;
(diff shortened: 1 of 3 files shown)


thestonefox:
Works as expected.

I'd suggest though the script lives in the main scripts folder and not the Helper folder. I'm trying to make it so the Helper folder is just classes that store common functions or hidden away stuff that developers dont need to worry about.

Also worth adding some docs on this to the new `DOCUMENTATION.md` file.

thestonefox:
resolves #324

10FingerArmy:
New patch pushed

thestonefox:
are you happy for this to be merged now? Looks like a decent enough start on the feature.

10FingerArmy:
yep

PR #337, opened 2016-08-01
Chore/upgrade to 5.4 resolves #336

### LINKED ISSUE #336: Upgrade VRTK to be Unity 5.4 compatible as standard
As Unity 5.4 is now released, the toolkit should support this version as standard and have backwards support for 5.3.

The examples that use the Headset Camera Fade will need updating as they don't work correctly between versions. It may be better to have a script in the scene that adds the camera fade script appropriately to make it version agnostic.

thestonefox:


thestonefox:
resolves #336

10FingerArmy:
Cool. In case you need changes to scene 25 please shortly sync with me since I have open changes there that would probably not be mergeable.

thestonefox:
@10FingerArmy there are no scene changes to 025 so you shouldn't be affected :)

10FingerArmy:
From a risk management perspective I'd probably not include this change shortly before the 2.2 release but right afterwards so that we gather some feedback from the community first.

PR #338, opened 2016-08-01
feat(RotationalTeleport): teleport with player reorientation

thacio:
Include teleport system which allows view rotation. The pointer which rotates must be a VRTK_SimplePointer. The teleport controller can be any type inherited of VRTK_WorldPointer.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
@@ -8,6 +8,8 @@ namespace VRTK
 {
     using UnityEngine;
 
+    public delegate void PointerToogleEventHandler(object sender);
+
     public abstract class VRTK_WorldPointer : VRTK_DestinationMarker
     {
         public enum pointerVisibilityStates
@@ -16,6 +18,9 @@ namespace VRTK
             Always_On,
             Always_Off
         }
+
+        public event PointerToogleEventHandler OnPointerOn;
+        public event PointerToogleEventHandler OnPointerOff;
 
         public VRTK_ControllerEvents controller = null;
         public Material pointerMaterial;
@@ -121,6 +126,7 @@ namespace VRTK
 
             if (playAreaCursor != null)
             {
+                ToggleBeam(false);
                 Destroy(playAreaCursor);
             }
         }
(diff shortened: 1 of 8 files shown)


thestonefox:
I'll check this out in the morning :)

Is it ok to close the other PR now? https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/312

thacio:
Yup! It's updated with the latest commit.

Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
```diff
@@ -283,6 +289,9 @@ private void TurnOnBeam(uint index)
                 TogglePointer(true);
                 isActive = true;
                 destinationSetActive = true;
+
+                if (OnPointerOn != null)
```
thestonefox: curly braces around the conditional block

Assets/SteamVR_Unity_Toolkit/Scripts/Abstractions/VRTK_WorldPointer.cs
```diff
@@ -293,8 +302,12 @@ private void TurnOffBeam(uint index)
                 controllerIndex = index;
                 TogglePointer(false);
                 isActive = false;
+
+                if (OnPointerOff != null)
```
thestonefox: curly whirly braces again :)

Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_SimplePointer.cs
```diff
@@ -22,6 +22,8 @@ public class VRTK_SimplePointer : VRTK_WorldPointer
         public float pointerLength = 100f;
         public bool showPointerTip = true;
         public GameObject customPointerCursor;
+        [Tooltip("Check this if you would like to keep the custom pointer cursor material.")]
+        public bool keepCustomPointerMaterial = false;
```
thestonefox: is it worth moving this into WorldPointer as it seems both SimplePointer and BezierPointer have it (and they both extend world pointer anyway)

thacio: customPointerCursor is implemented by them, instead of WorldPointer, to they probably should stick together.
I think it would be better to leave there for now and move customPointerCursor and keepCustomPointerMaterial to WorldPointer in another commit.

thacio:
Done. Updated the example scene with instructions how to use the feature and changed the direction pointer to something that kind of maybe from far away looks like an arrow.

PR #341, opened 2016-08-02
feat(Interaction): service pack for 3D controls

10FingerArmy:
This patch is the result of adapting the first set of controls to a
real game scene.

So far only the drawer allowed to set an optional content object. This
was now refactored into the control and now also the door and chest can
profit from the same logic.

In addition objects that were taken out of a drawer would disappear
again if the drawer was closed even if the objects were outside. This
needed to be handled with custom logic. Now there is a helper script
that will manage reparenting automatically.

In addition some code cleanup was done and where already applicable the
direction was extracted with the ultimate goal to have only one
direction enum.

Finally the button contains a small bug fix to define the force in a
much more reliable way.


thestonefox:
I'm happy to merge this if you're happy

10FingerArmy:
Sure go ahead. I actually only put up things i tested 😊

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/Utilities/VRTK_ContentHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// This approach is much more stable and allows for many more scenarios.
+    /// Not supported yet: detect if an object is taken from one content manager and dropped into another one.
+    /// </summary>
+    public class VRTK_ContentHandler : MonoBehaviour
+    {
+        public VRTK_Control control;
+        public Transform inside;
+        public Transform outside;
+
+        void Start()
```
thestonefox: missing private keyword

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/Utilities/VRTK_ContentHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+                        VRTK_ContentHandler ch = childIo.gameObject.AddComponent<VRTK_ContentHandler>();
+                        ch.control = control;
+                        ch.inside = inside;
+                        ch.outside = outside;
+                    }
+                }
+            }
+        }
+
+        public void OnCollisionEnter(Collision collision)
```
thestonefox: should this be public? Is anything utilising this method?

If it should be public then the method should be documented

thestonefox:
Just a couple of things I've noticed on one final read through, no biggies though :)

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/Utilities/VRTK_ContentHandler.cs
10FingerArmy: this is the Unity default which I would not change

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/Utilities/VRTK_ContentHandler.cs
10FingerArmy: This is the definition of the MonoBehaviour and should not be changed, it's the official signature

thestonefox: Their example has it as a private method :/

https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnCollisionEnter.html

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/Utilities/VRTK_ContentHandler.cs
thestonefox: all methods/variables should specify public/protected/private as per the contrib guidelines.

Otherwise there is a mix and match of when it's included and when it's not.

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/Utilities/VRTK_ContentHandler.cs
10FingerArmy: ah, no it's not private, it's internal visibility which means public inside the project. I tried to find out the guideline from Unity but was not successful. Definitely all MonoBehaviour methods are public (you can simply press F12 on MonoBehaviour), the Messages (Start, Awake) are not covered there though, they are magic. For sure they need to be more than private to work but what the exact visibility needs to be is a mystery. Unity creates Start/Awake with internal visibility. VS creates the remaining Messages (OnCollision...) with public visibility when selecting "Implement MonoBevahviours". I don't know enough scenarios to judge if it is safe to reduce the visibility of these from public to internal. Will that break scenarios where one could have reused a C# component from another project somehow? It's quite theoretical but still important to think about. What do you think? For now we should probably be fine by making them identical to Start/Awake and remove the visibility declaration.

10FingerArmy:
I just looked through the code, e.g. headsetcollisionfade, interactableobject... they actually do it wrong and elevate the visibility from private to protected. Why is that?

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/Utilities/VRTK_ContentHandler.cs
thestonefox: According to the C# docs on msdn https://msdn.microsoft.com/en-us/library/ba0a1yw2.aspx (and this video from Unity https://youtu.be/XKGT_4mR1C4?t=163)

Any method that omits the access modifier is private by default. But they even (in that video) suggest to specify private/public if you have a mixture of access modifiers in the class.

That would lead me to believe the reason they're created without an access modifier by default (e.g. you get a `void Start()` and a void `Update()`) is because they are the only methods in that class (and therefore they seem to think omitting the access modifier is ok because it's assumed it's all private).

From this I would assume that the contrib guidelines are adhering to what they're suggesting. I wonder if anyone else has any ideas/understandings about this?

thestonefox:
it's fine to have the access to the method different from private (like in the headset collision fade) You may require access to that method further down the inheritance tree.

It's also fine having the method public if it needs to be publicly accessible.

If that OnCollisionEnter was required to be accessed from outside of the class then it's cool to have it public. But as Unity say in their videoz if you don't need access then it should be private.

But if you do find instances in the code where it doesn't make sense, like in headset collision fade, should it actually be protected? does anything inherit off it? And I just looked and it doesn't look like it does, then you're spot on this is wong! And I will do a PR to fix it, cos it's against what I've just said :)

I've no idea why i set them to protected :/ i was probably randomly hitting keys :D

Issue #335, opened 2016-08-01
Using items doesn't work as expected when the object is swapped between hands.


thestonefox:
If you pick up an object in one hand and start using it, then without releasing the use button on that controller, grab the item with the other controller

It will continue to be used until the button on the original controller is released,

Upon ungrab, the force stop using on the controller should also be called.

thestonefox:
fixed in #342

PR #342, opened 2016-08-02
fix(Interaction): prevent controller swap breaking use functionality - resolves #343

### LINKED ISSUE #343: Using items doesn't work as expected when the object is swapped between hands
If you pick up an object in one hand and start using it, then without releasing the use button on that controller, grab the item with the other controller

It will continue to be used until the button on the original controller is released,

Upon ungrab, the force stop using on the controller should also be called.

thestonefox:
Previously, when using an object and swapping controllers (either
grab or touch state) the use state on the object would not be reset
correctly and therefore the previous using controller would still
be considered the active use controller.

This fix ensures that when a controller stops touching or grabbing
an object, that the use state on the controller is also reset to
no longer using the object.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractUse.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractUse.cs
@@ -57,6 +57,14 @@ namespace VRTK
             }
         }
 
+        public void ForceResetUsing()
+        {
+            if (usingObject != null)
+            {
+                UnuseInteractedObject(false);
+            }
+        }
+
         private void Awake()
         {
             if (GetComponent<VRTK_InteractTouch>() == null)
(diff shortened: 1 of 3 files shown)


thestonefox:
resolves #335

thestonefox:
fixes #335

thestonefox:
fixes #335

thestonefox:
resolves #343

Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractUse.cs
```diff
@@ -57,6 +57,14 @@ public void ForceStopUsing()
             }
         }
 
+        public void ForceResetUsing()
```
thestonefox: need to add docs for this method

PR #345, opened 2016-08-03
feat(Pointer): Customized compound teleport

gpvigano:
VRTK_SimplePointer: get the material from customPointerCursor to be set
on all the objects composing the pointerTip.

CurveGenerator: using a custom tracer the first segment must be
displayed, otherwise the curve seems to start in the air and not emitted
by the controller.

VRTK_BezierPointer: Define a custom teleport pointer composed by two
different types of game objects.
A set of objects is like the previous custom pointer, with the materials
changed when the destination is allowed or not.
Another game object (e.g. teleport beam) can be defined to appear only
if the teleport is allowed (its material will not be changed ).
Custom teleport tracer in VRTK_BezierPointer: rescale each pointer
tracer element according to the length of the Bezier curve.
Add member rescalePointerTracer both in VRTK_BezierPointer and
CurveGenerator, add an extra parameter in CurveGenerator.Create()
(default=false).
Affect also emission color if present.

Add example scene  and related resources.
Add as prefabs (with their related 3D models and textures) a custom
teleport pointer, a custom pointer tracer and a custom object to be
attached as teleport beam.
Add a new example scene 036_Controller_CustomCompoundPointer.unity to
test the new teleport features.

Add documentation to DOCUMENTATION.md


gpvigano:
Some screenshots from the example added within this branch:
customized_compound_teleport_1
customized_compound_teleport_2
customized_compound_teleport_3

thestonefox:
Looked over the code, it all looks good, I'll give it test soon and hopefully merge it today!

Nice work :)

PR #351, opened 2016-08-04
fix(ControllerEventsUI): Fix bug when using SteamVR Keyboard and UIPointer

thacio:
If SteamVR Keyboard is shown, SteamVR disables both controllers in the
scene. If the UIPointer is currently active (will happen if using the
UI, for example if input field is clicked), VRTK_ControllerEvents will
never change state  because controllers are disable. so UIPointer won't
act accordingly, This causes unwanted behaviours as even if no Pointer
is active, you can still interact with the UI.


GitCop:

 * Commit: f79f1a198bd5618e8650d6d80f9f2bb3ccc67e82

---

PR #354, opened 2016-08-05
fix(Teleporter): add blink y threshold when changing height.

thestonefox:
A `Blink Y Threshold `new parameter has been added to the
Height Adjust Teleporter that allows a custom y threshold to be given
to determine when the blink transition should be activated when
the user's height over the object they are standing on changes.

Any height difference that is above the threshold will cause the blink
transition to process, but anything under the threshold and the
`[CameraRig]` will just adjust the `y` position without blinking.

This is useful for climbing up slopes such as terrains or meshes.


thestonefox:
Superseded by https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/330

PR #355, opened 2016-08-05
fix(Interaction): make force of drawer more reasonable

10FingerArmy:
One needed to pull too strong to open the drawer so far.


thestonefox:
lowercase the git commit subject like `Make` should be `make`, then it's good to go! :D

PR #356, opened 2016-08-05
fix(Interaction): allow touch highlight to affect all object materials resolves #310

### LINKED ISSUE #310: On multiple material objects, highlight on touch only changes the first material colour.
This makes it difficult for the player to notice, if the first material is a small accent piece which isn't largely visible.

Ideally, the option to choose a single material #, or all material to highlight would be ideal. 

However, 'all material' would be a better default than the current.

Sample project with a dual material asset: https://dl.dropboxusercontent.com/u/85141892/Sample_Highlight_On_Multiple_Material.rar

thestonefox:
Previously, the touch highlighting would only change the first material
on the Interactable Object. This fix ensures that all the materials on
an object get updated by the hightlight colour.


thestonefox:
resolves #310

arkhint:
Tested on an object with more than 1 material. Results in all material being highlighted. Works as expected now.

PR #357, opened 2016-08-06
fix(Interaction): ensure grab state is reset for tracked objects

HenryGarle:
Minor fix for tracked objects, the grab state was not being reset when
"Hold Button To Grab" was disabled after dropping an item. When grabbing it for a second time you have to hold the button despite it being disabled. Moving the state
reset into InitUngrabbedObject() fixes the issue as it is called no
matter which path is used to drop an item.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractGrab.cs
@@ -381,13 +381,13 @@ namespace VRTK
                 controllerActions.ToggleControllerModel(true, grabbedObject);
             }
 
+            grabEnabledState = 0;
             grabbedObject = null;
         }
 
         private void ReleaseObject(uint controllerIndex, bool withThrow)
         {
             UngrabInteractedObject(controllerIndex, withThrow);
-            grabEnabledState = 0;
         }
 
         private GameObject GetGrabbableObject()


GitCop:

 * Commit: a41e1c291a8c896d1bde9c97f7413169ec6d0025

---

thestonefox:
Seems like a wise fix.

You'll need to update the commit subject line in your commit message (not just on github)

so it reads

`fix(Interaction): ensure grab state is reset for tracked objects`

Reason is, the scope is more of the generic theme and this theme is `Interaction` (it looks better when they're grouped like this in the changelog)

Also, the message part should be all lower case, and no point in saying `fix ....` in the message because the type is `fix` so the second mention is redundant somewhat.


To amend your git commit message you need to `git commit --amend` on your branch and change the message, then git force push it back up `git push origin fix/tracked-object -f`

HenryGarle:
Sorry about that, message should now be updated :)

thestonefox:
So close! :) your message has a spelling error :) `stats` should be `state`

HenryGarle:
Sausage fingers ...

thestonefox:
I've just noticed your commit message body has gone too :(

do a `git commit --amend` again and add the body in, so it should be this:

```
fix(Interaction): ensure grab state is reset for tracked objects

Minor fix for tracked objects, the grab state was not being reset when
"Hold Button To Grab" was disabled after dropping an item.
When grabbing it for a second time you have to hold the button
despite it being disabled. Moving the state reset into
InitUngrabbedObject() fixes the issue as it is called no
matter which path is used to drop an item.
```

Directly copy and paste the above and it will be the correct line length so gitcop won't moan :)

HenryGarle:
I used -m to edit the subject line thinking I was being clever not realizing it had cleared it.

thestonefox:
i would say it's ready ❤️ 💚 💛 💙

PR #359, opened 2016-08-06
feat(UI): add ability to ignore certain world canvases - resolves #309

### LINKED ISSUE #309: UIPointer - Issues with canvas attached to controller
This is more for the information of others who will have canvas's attached to their controllers (for example, ammo read outs on guns or health bars on controllers).

If using the UI Pointer it grabs all the world canvas and adjusts them to suit it's needs, when it grabs one (or more) canvas's on the controller it causes the rotation logic to be wrong (I'm not sure why it's wrong but, for example, my gun seems to be hitting something on certain axis causing it to be bend at an odd angle).

My solution was to simply add a tag called "IgnoredByPointer", set this on all controller attached canvas's and have the SetWorldCanvas do a quick check:

        public static void SetWorldCanvas(Canvas canvas)
        {
            if (canvas.renderMode != RenderMode.WorldSpace)
            {
                return;
            }

			if (canvas.gameObject.CompareTag ("IgnoredByPointer"))
				return;

thestonefox:
A new inspector parameter has been added that provides the ability
to ignore certain world canvases that match a given tag or contain
a specific named class.

The `Ignore Canvas With Tag Or Class` parameter accepts a string and
any canvas that is tagged with the same string or contains a script
that is named the same as the string will not be interactable with
the UI Pointers.

The UI example scene has been updated to demonstrate this exclusion.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Examples/034_Controls_InteractingWithUnityUI.unity
+++ b/Assets/SteamVR_Unity_Toolkit/Examples/034_Controls_InteractingWithUnityUI.unity
@@ -3233,6 +3233,80 @@ CanvasRenderer:
   m_PrefabParentObject: {fileID: 0}
   m_PrefabInternal: {fileID: 0}
   m_GameObject: {fileID: 428040042}
+--- !u!1 &434534480
+GameObject:
+  m_ObjectHideFlags: 0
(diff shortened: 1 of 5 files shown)


thestonefox:
resolves #309

PR #361, opened 2016-08-07
feat(Interaction): service pack for 3D controls

10FingerArmy:
This patch is the result of adapting the controls further to a real game
scene.

The lever has been reworked, coding aligned with the other controls and
the detection is improved.

Auto-detection of door anchor points has been greatly improved by a new
calculation method. The nested door in 025 does not need the manual
hingejoint anymore.

All controls now support returning normalized values that range between
0-100%.

In progress but stuck due to one last Unity issue: Buttons now support
being moved at run-time (gizmos and activation point already fixed,
only configurable joint left which does not seem updatable at run-time).
To try out the issue simply activate the animation on "ButtonWall".

Scene 025 has been cleaned up a bit and the texts now also show the
normalized value.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Button.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Button.cs
@@ -18,6 +18,8 @@
 
         public ButtonEvents events;
 
+        [Tooltip("An optional game object that when it moves also moves the button.")]
+        public GameObject connectedTo;
         public ButtonDirection direction = ButtonDirection.autodetect;
         public float activationDistance = 1.0f;
         public float buttonStrength = 5.0f;
@@ -25,14 +27,16 @@
         private static float MAX_AUTODETECT_ACTIVATION_LENGTH = 4; // full hight of button
 
         private ButtonDirection finalDirection;
-        private Vector3 initialPosition;
-        private Vector3 activationPoint;
+        private Vector3 restingPosition;
+        private Vector3 jointDelta;
+        private Vector3 activationDir;
 
         private Rigidbody rb;
         private ConfigurableJoint cj;
         private ConstantForce cf;
+        private int forceCount = 0;
 
-        public override void OnDrawGizmos()
+        protected override void OnDrawGizmos()
         {
             base.OnDrawGizmos();
             if (!enabled || !setupSuccessful)
@@ -41,12 +45,12 @@
             }
 
             // visualize activation distance
-            Gizmos.DrawLine(bounds.center, activationPoint);
+            Gizmos.DrawLine(bounds.center, bounds.center + activationDir);
         }
 
         protected override void InitRequiredComponents()
         {
-            initialPosition = transform.position;
+            restingPosition = transform.position;
 
             if (!GetComponent<Collider>())
             {
@@ -66,6 +70,16 @@
             {
                 cf = gameObject.AddComponent<ConstantForce>();
             }
+
+            if (connectedTo)
+            {
+                Rigidbody rb2 = connectedTo.GetComponent<Rigidbody>();
+                if (rb2 == null)
+                {
+                    rb2 = connectedTo.AddComponent<Rigidbody>();
+                }
+                rb2.useGravity = false;
+            }
         }
 
         protected override bool DetectSetup()
@@ -73,12 +87,12 @@
             finalDirection = (direction == ButtonDirection.autodetect) ? DetectDirection() : direction;
             if (finalDirection == ButtonDirection.autodetect)
             {
-                activationPoint = transform.position;
+                activationDir = Vector3.zero;
                 return false;
             }
             if (direction != ButtonDirection.autodetect)
             {
-                activationPoint = CalculateActivationPoint();
+                activationDir = CalculateActivationDir();
             }
 
             if (cf)
@@ -89,12 +103,35 @@
             if (Application.isPlaying)
             {
                 cj = GetComponent<ConfigurableJoint>();
-                if (cj == null)
-                {
-                    // since limit applies to both directions object needs to be moved halfway to activation before adding joint
-                    transform.position = transform.position + (activationPoint - transform.position).normalized * activationDistance * 0.5f;
 
-                    cj = gameObject.AddComponent<ConfigurableJoint>();
+                bool recreate = false;
+                Rigidbody oldBody = null;
+                Vector3 oldAnchor = Vector3.zero;
+                Vector3 oldAxis = Vector3.zero;
+
+                if (cj)
+                {
+                    // save old values, needs to be recreated
+                    oldBody = cj.connectedBody;
+                    oldAnchor = cj.anchor;
+                    oldAxis = cj.axis;
+                    DestroyImmediate(cj);
+                    recreate = true;
+                }
+
+                // since limit applies to both directions object needs to be moved halfway to activation before adding joint
+                transform.position = transform.position + activationDir.normalized * activationDistance * 0.5f;
+                cj = gameObject.AddComponent<ConfigurableJoint>();
+
+                if (recreate)
+                {
+                    cj.connectedBody = oldBody;
+                    cj.anchor = oldAnchor;
+                    cj.axis = oldAxis;
+                }
+                if (connectedTo)
+                {
+                    cj.connectedBody = connectedTo.GetComponent<Rigidbody>();
                 }
 
                 SoftJointLimit sjl = new SoftJointLimit();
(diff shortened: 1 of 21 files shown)


Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Lever.cs
```diff
@@ -10,46 +10,33 @@ public enum LeverDirection
         }
 
         public LeverDirection direction = LeverDirection.y;
-        public float min = 0f;
-        public float max = 100f;
-        public float stepSize = 1f;
+        public float minAngle = 0f;
```
thestonefox: Docs require updating to reflect new variable names

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Control.cs
```diff
@@ -42,7 +52,16 @@ public float getValue()
             return value;
         }
 
-        public virtual void Start()
+        /// <summary>
+        /// Returns the current value mapped onto a range between 0% and 100%.
+        /// </summary>
+        /// <returns>Normalized Value</returns>
+        public float getNormalizedValue()
```
thestonefox: new method will need documenting

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Control.cs
```diff
@@ -42,7 +52,16 @@ public float getValue()
...
+        {
+            return Mathf.Abs(Mathf.Round((value - cvr.controlMin) / (cvr.controlMax - cvr.controlMin) * 100));
+        }
+
+        public virtual void Awake()
```
thestonefox: Does Awake need to be public? Could it be protected? I'm guessing the only things that would need to access it would be things that inherit the Control class? Or do things outside of the scope of the inherited classes need to access it?

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Control.cs
thestonefox: Actually i dont think it will, we don't have docs for the VRTK_Control do we? I'm guessing you never use this class, so it doesn't need documenting?

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Control.cs
```diff
@@ -72,7 +92,7 @@ public virtual void Update()
                 if (value != oldValue)
                 {
                     HandleInteractables();
-                    defaultEvents.OnValueChanged.Invoke(getValue());
+                    defaultEvents.OnValueChanged.Invoke(getValue(), getNormalizedValue());
```
thestonefox: Is this an event we should document in all of the other inherited classes as I'm guessing they all throw this event?

10FingerArmy:
Thanks a lot for the comments. You actually detected that the generic event was never described at all so far. It is now! I also changed the visibility of all Awake/Start/Gizmo to their minimum (protected).

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Button.cs
```diff
@@ -179,6 +187,19 @@ protected override void HandleUpdate()
             }
         }
 
+        /*
```
thestonefox: Is this the chunk of code that's being left in but commented out to see if other people can figure out the button moving?

DOCUMENTATION.md
```diff
@@ -1686,6 +1686,10 @@ The following UI controls are available:
   * [VRTK_Lever](#vrtk_lever)
   * [VRTK_Slider](#vrtk_slider)
 
+All controls emit the following events:
```
thestonefox: As this overview page isn't on the readme.io site, it may be better to have the `### Class Events` section on each of the controls docs (plus the `#### Event Payload` they emit)

Then if people jump to a specific control they can read the events for that specific control (even though it's just a copy and paste)

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Control.cs
```diff
@@ -42,7 +52,27 @@ public float getValue()
...
+        /// <summary>
+        /// Returns the current value mapped onto a range between 0% and 100%.
+        /// </summary>
+        /// <returns>Normalized Value</returns>
+        public float getNormalizedValue()
+        {
+            return Mathf.Abs(Mathf.Round((value - cvr.controlMin) / (cvr.controlMax - cvr.controlMin) * 100));
+        }
+
+        public void SetContent(GameObject content, bool hideContent)
```
thestonefox: I know we don't document the VRTK_Control file. But I'm thinking maybe we should (like we document the WorldPointer which is an abstract class). Then all of these public methods can be documented (as well as the public event).

Then in each of the controls we can just say they extend this class (like the SimplePointer and BezierPointer docs show)

I think that would be more consistent and means (as I mentioned in the other comment) that there isn't the need to copy and paste the same docs between all of the controls

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Button.cs
10FingerArmy: Exactly, I would like to leave it in for others to continue. I got pretty far already.

DOCUMENTATION.md
10FingerArmy: I don't think we should copy paste it. What I could imagine is some sort of back reference stating "In addition to the default control events and methods, the following are exposed in addition:".

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Control.cs
10FingerArmy: The more I think about it the more I think we should actually document the control. It would also make sense from a user perspective to make it easy for them to create new controls. The syntax is too much for me though right now so late in the evening.

thestonefox: Sure no worries, but yeah I think this is better than the copy/paste idea I had. gets out of sync too easily with the old copy/pasta approach :)

DOCUMENTATION.md
thestonefox: agreed, ignore this idea :)

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Button.cs
thestonefox: cool, at least it's a starting point if someone else gets round to looking at it first

10FingerArmy:
I went through the comments and answered. I would currently not judge the two missing public methods as a stopper for this commit, especially since they were already in before, but I think we should document the VRTK_control class. That would be the place for common methods and events and would make it easy for others to write their own controls.

Unfortunately I cannot get it done right now. My proposal is to still merge this and then either I do the doc next week or you help me out and add the VRTK_control doc basics which I can later extend. I would for sure get the syntax incorrect I think :-/

thestonefox:
Ok I'll take a stab at writing the docs up for you to copy paste in to the `DOCUMENTATION.md` file and push with this PR. doesn't have to be today though.

If the 2.1.0 release is delayed a bit then it's not the end of the world, after all this is a free toolkit! people need to realise we're doing this stuff out of the kindness of our hearts :)

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Control.cs
```diff
@@ -42,7 +52,27 @@ public float getValue()
             return value;
```
thestonefox: The method should be `GetValue()` not `getValue()`

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Control.cs
```diff
@@ -42,7 +52,27 @@ public float getValue()
...
         }
 
-        public virtual void Start()
+        /// <summary>
+        /// Returns the current value mapped onto a range between 0% and 100%.
+        /// </summary>
+        /// <returns>Normalized Value</returns>
+        public float getNormalizedValue()
```
thestonefox: should be `GetNormalizedValue()` (upper case method name)

Assets/SteamVR_Unity_Toolkit/Scripts/Controls/3D/VRTK_Control.cs
thestonefox: Right I've created a draft Documentation:

The whole file is at this gist ->https://gist.github.com/thestonefox/48f9d887cfcf5b804a70f0dddad3e7ea

It includes your doc changes as well so you should just be able to get the raw file from the gist and overwrite your local one and repush the PR.

PR #364, opened 2016-08-08
fix(UIPointer): get pointer id from specified controller

thacio:
Previously, the UIPointer would attempt to assign the pointer id
based on the provision that a Tracked Object script was present on
the same component as the UIPointer script. This meant it could
only be applied to the controllers or the headset by default.

The pointer id is now determined by the index of the controller
passed to the UI Pointer and therefore the UI pointer can be
attached to any game object that can also emit a world pointer.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_UIPointer.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_UIPointer.cs
@@ -76,12 +76,12 @@
 
         private void Start()
         {
-            ConfigureEventSystem();
-            ConfigureWorldCanvases();
             if (controller == null)
             {
                 controller = GetComponent<VRTK_ControllerEvents>();
             }
+            ConfigureEventSystem();
+            ConfigureWorldCanvases();
         }
 
         private void ConfigureEventSystem()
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: e23f2ec1728618c2a5c59e49415a49de0196f1f5

---

thacio:
oh you, git cop :(
What's the problem? I thought it could be the space after the ":", so I changed it.

thestonefox:
I'd update the commit message (subject and body) to something like:

```
fix(UIPointer): get pointer id from specified controller

Previously, the UIPointer would attempt to assign the pointer id
based on the provision that a Tracked Object script was present on
the same component as the UIPointer script. This meant it could
only be applied to the controllers or the headset by default.

The pointer id is now determined by the index of the controller
passed to the UI Pointer and therefore the UI pointer can be
attached to any game object that can also emit a world pointer.
```

Bit of reasoning of why this is more appropriate: Having a subject that has the type of `fix` doesn't need the word "fix" mentioning in the subject text because it's implicitly a fix, plus it's best to say what the fix actually does. The body should try and explain the rationale behind the change rather than the implicit changes to the code (as that can be seen from looking at the code diff in the commit)

PR #369, opened 2016-08-09
fix(Interaction): allow pause collisions on tracked objects

thestonefox:
Previously, tracked objects did not pause collisions on grab due
to the code that calls the pause being inaccessible. This code has
now been moved so it is accessible for all grab mechanics.


HenryGarle:
Far better, I missed the common call to InitGrabbedObject().

PR #380, opened 2016-08-11
fix(Climbing): fix collision teleport bug, rope grab and usePlayerScale when jumping

mattboy64:


thestonefox:
not sure if this has gone wrong, but your commit is actually a merge pull request..

Did you commit amend onto the last commit?

PR #381, opened 2016-08-11
feat(Controller): optionally disable the pointer on grab or on use

gpvigano:
Add to Interactable Object a **Pointer Disable Mode** option
(`pointerDisableMode` in `VRTK_InteractableObject`) of type
`PointerDisableMode`:
 * `DontDisable` means not disabling the pointer.
 * `OnGrab` means disabling the pointer when grabbed.
 * `OnUse` means disabling the pointer when using.

Update methods `StartUsing` `StopUsing` `Grabbed` `Ungrabbed` to
support this option.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractableObject.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_InteractableObject.cs
@@ -46,6 +46,13 @@ namespace VRTK
             OverrideDontHide,
         }
 
+        public enum PointerDisableMode
+        {
+            DontDisable,
+            OnGrab,
+            OnUse,
+        }
+
         [Header("Touch Interactions", order = 1)]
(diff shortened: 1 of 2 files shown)


gpvigano:
I just pulled out from my previous PR #367 I feature that could be useful in general, then I think I can close the PR #367 (leaving that stuff for a future possible showcase repo).

thestonefox:
I can see the purpose of this, i'm not sure i agree with the approach.

Feels a bit coupled relating directly to the pointers.

I would probably do it by having a 3rd party helper script (something not specifically in the toolkit) that listens for the on grab event and the disables the pointers and re-enables them on the ungrab event.

For whatever reason, someone may be using a different pointer which isn't a world pointer and then this wouldn't work.

gpvigano:
You're right, it's coupled because of code like this:
`grabbingObject.GetComponent<VRTK_WorldPointer>()`
OK, I close this PR.

PR #401, opened 2016-08-14
chore(HeadsetCollisionFade): Independent blink times and base class

sblancodiez:
Simple changes to make HeadsetCollisionFade extendable/modifiable. Also took the chance to add another variable to get independent fade times.

Yesterday I implemented (not in this commit) a different version of the HeadsetCollisionFade script that created a gameobject with a collider just in the last good position and checked that you were inside this collider and outside others before fading back. Also, if you kept moving further it resetted the position of the CameraRig so just backing a little bit always faded you back to a good position. This way the player can't go through walls and cheat.

If this commit is good for you I will implement cleanly this different method in a new script that inherits from the new base so users can choose or create other methods.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_BasicTeleport.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_BasicTeleport.cs
@@ -198,7 +198,7 @@ namespace VRTK
 
         private void InitHeadsetCollisionListener(bool state)
         {
-            var headset = FindObjectOfType<VRTK_HeadsetCollisionFade>();
+            var headset = FindObjectOfType<VRTK_HeadsetCollisionFadeBase>();
             if (headset)
             {
                 if (state)
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_HeadsetCollisionFade.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_HeadsetCollisionFade.cs
@@ -10,40 +10,16 @@
 
     public delegate void HeadsetCollisionEventHandler(object sender, HeadsetCollisionEventArgs e);
 
-    public class VRTK_HeadsetCollisionFade : MonoBehaviour
+    public abstract class VRTK_HeadsetCollisionFadeBase : MonoBehaviour
     {
-        public float blinkTransitionSpeed = 0.1f;
+        public float blinkTransitionSpeedEnter = 0.1f;
+        public float blinkTransitionSpeedExit = 0.1f;
         public Color fadeColor = Color.black;
-        public string ignoreTargetWithTagOrClass;
 
         public event HeadsetCollisionEventHandler HeadsetCollisionDetect;
         public event HeadsetCollisionEventHandler HeadsetCollisionEnded;
 
-        public virtual void OnHeadsetCollisionDetect(HeadsetCollisionEventArgs e)
-        {
-            if (HeadsetCollisionDetect != null)
-            {
-                HeadsetCollisionDetect(this, e);
-            }
-        }
-
-        public virtual void OnHeadsetCollisionEnded(HeadsetCollisionEventArgs e)
-        {
-            if (HeadsetCollisionEnded != null)
-            {
-                HeadsetCollisionEnded(this, e);
-            }
-        }
-
-        private HeadsetCollisionEventArgs SetHeadsetCollisionEvent(Collider collider, Transform currentTransform)
-        {
-            HeadsetCollisionEventArgs e;
-            e.collider = collider;
-            e.currentTransform = currentTransform;
-            return e;
-        }
-
-        private void Start()
+        protected virtual void Start()
         {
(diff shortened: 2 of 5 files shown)


thestonefox:
Hi,

I'm not sure i understand the reason to rename the HeadsetCollisionFade to HeadsetCollisionFadeBase and make it an abstract class.

If you are hoping to extend this but because of the private methods it's preventing you, then the PR should probably just update the class to change the relevant private methods to protected.

sblancodiez:
Hi,

Its not a rename, there is still a HeadsetCollisionFade class. I did it that way so the base had what I think it would be mandatory for any possible implementation (basically, event dispatching and initialization) and then the HeadsetCollisionFade class builds from that the collision and ValidTarget thing.

If you prefer I can merge everything into a single class as you say, I just thought that having ValidTarget and the OnCollision handlers protected virtual instead forced other implementations to override them (maybe even leaving them empty) if they wanted to do things differently. Having two classes allow people to either override the base and do their thing or override the normal HeadsetCollisionFade if they just want to build from there.

Either way will work in the end, so please confirm if you want me to join everything under a single class or not and I will change whatever needed.

thestonefox:
Right, I see now the abstract class and concrete class are in the same file.

But if your concept is to have a base class that just deals with headset collision then I think it would be better to have a class in the Abstractions directory called `VRTK_HeadsetCollision` that purely deals with the headset collider trigger and throwing the events.

Then a concrete class `VRTK_HeadsetCollisionFade` inherits that and deals with fading the headset view on collision.

Then you could write other classes to do other things on headset collision (without ever needing to fade)

sblancodiez:
Hi,

It is the opposite of your statement; the base class does not do anything with actual collisions nor does it know how that is done, it just has the functionality of fading and event dispatching so other parts of the toolkit can react to the fade (Teleport, for example). In other words, the base class has the minimum functionality relating to other parts of the toolkit and the fading.

Then, the particular inherited class detects collisions or calculates in any way wether the headset should fade in or fade out and uses the base class functionality to do the actual fade and send the events.

I don't know if you can look at the full source in the pull request, but I think it is seen more clearly than using the diff. As an example, right now I could inherit from the base class and do a different implementation based on raycasting or positions or other flags specific to my project and would not have to actually override all the detection stuff.

Separating to another file in abstractions the base class is a good idea.

Sorry if I don't explain properly, English is not my mother tongue so maybe I am not explaining properly.

thestonefox:
Which idea do you think is better moving forward?

* A generic HeadsetCollision base class that is purely used for knowing when the headset has collided. Then another class that extends this to fade the headset on collision.

* A generic fade headset class that can be triggered to fade the headset, then extended into a class that calls the fade when a collision happens?

My preference is the first one, because the headset fade is just a call to SteamVR.Fade anyway.

sblancodiez:
I do think using a base class is the way to go since that will allow for better developer experience. But I still think that the collision detection per se should not be in the base class since that could potentially be achieved in different manners with different constraints, filters or philosophies. Normally using a collider and checking the Enter and Exit is the way to go, but there may be different cases.

If the user just needs to do some modification on top of the current implementation, they can inherit from the the child class (HeadsetCollisionFade) instead of the base class. But if they want to do something different, they can inherit from the base class so they will be compatible with the rest of the toolkit (because of the standardized event handling in the base class) and they can do whatever they want without having to break/modify anything and without needing to override with empty methods.

Anyway, surely I am just nitpicking. If you feel it would be better to have it unified and leave everything protected virtual that is enough so anyone can do their thing.

thestonefox:
If the user wants to make the headset fade without collision, then they can just call `SteamVR.fade()`- it's built straight into steamVR, that's why I think I'm not seeing the benefit of a core class that just does headset fading, because it would probably just be a wrapper over `SteamVR.fade()` ,unless there are fade/unfade events, and in which case I wonder if it's better to not use inheritance but to use composition to just include a headset action on collision (one could be headset fade) then the headset fade action could be included on other classes too (if you didn't want collision to cause the fade).

Are you in the slack channel? This is probably worth discussing in there to get more people joining in the discussion.

sblancodiez:
I will investigate this slack channel.

The composition idea is good too. The thing is, it is not just fading, because as you say that is just a call to the SteamVR API. The idea is fading and deactivating Teleport and maybe other things while faded, so the events are very important to get that. 

Also, I emphasize the possibility of different implementations regarding the 'collision' detection. Forcing it to be in one way and just leaving other scripts the possibility to react on this is not enough. Putting the idea in simple terms, I think that there is a condition that makes the screen fade and all fade side-effects and other condition that makes the screen fade back and remove side-effects. These conditions may be different in nature, but notifying other scripts via events and the actual fading (for now, since in the future making something to fade to a texture, for example, would be great) are mandatory.

thestonefox:
For completeness (following the Slack discussion) I will take a stab at implementing the new changes to the headset collision stuff and then link to my PR in this PR.

(so let's leave this PR open for now!)

thestonefox:
I'm going to close this as I have an issue open to resolve this now

PR #412, opened 2016-08-17
fix(Controller): enhance AutoGrab script

10FingerArmy:
AutoGrab would fail if the controllers initially already touched
something. This is now fixed.

Also stronger typing has been introduced and the script flow fixed so
that now error messages also appear in the console.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_ObjectAutoGrab.cs
+++ b/Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_ObjectAutoGrab.cs
@@ -5,7 +5,7 @@
 
     public class VRTK_ObjectAutoGrab : MonoBehaviour
     {
-        public GameObject objectToGrab;
+        public VRTK_InteractableObject objectToGrab;
         public bool cloneGrabbedObject;
 
         private VRTK_InteractGrab controller;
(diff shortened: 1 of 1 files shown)


Assets/SteamVR_Unity_Toolkit/Scripts/VRTK_ObjectAutoGrab.cs
```diff
@@ -5,36 +5,38 @@
...
         private VRTK_InteractGrab controller;
 
         private IEnumerator Start()
         {
             controller = GetComponent<VRTK_InteractGrab>();
-
             if (!controller)
             {
                 Debug.LogError("The VRTK_InteractGrab script is required to be attached to the controller along with this script.");
+                yield break;
```
thestonefox: Do you want to return out of this function here? As far as I know Debug.LogError will stop/pause the scene with the error, but you then in theory at this paused state add the interact grab script to the controller and resume the scene for it to carry on working.

If it's returned at this point then unpausing after the break won't actually carry on if you've applied the fix?

10FingerArmy: This is not how we should handle this IMO. This is a setup issue you should fix once. The old state was plain wrong since when the code continued to run (or you simply did not have errorpause on like most have) it would work on null objects. Also at least for me the error did not even appear in the console, it was simply hidden by Unity for some reason. This solution should be much cleaner and informative IMO.

PR #414, opened 2016-08-17
feat(Teleport): add budget cuts teleporter

mimminito:
Added in a budget cuts style teleporter. You can use a pointer to set a destination which in turn spawns a SpyCam. The SpyCam then allows you to look around through a viewport attached to the controller. You can then press the PointerSet button to teleport to that location.

A new example scene 038_BudgetCutsTeleport has been added to showcase the teleporter.

### DIFF
--- a/Assets/SteamVR_Unity_Toolkit/Examples/038_BudgetCutsTeleport.unity
+++ b/Assets/SteamVR_Unity_Toolkit/Examples/038_BudgetCutsTeleport.unity
@@ -1,0 +1,2697 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!29 &1
+SceneSettings:
+  m_ObjectHideFlags: 0
+  m_PVSData: 
+  m_PVSObjectsArray: []
+  m_PVSPortalsArray: []
+  m_OcclusionBakeSettings:
+    smallestOccluder: 5
+    smallestHole: 0.25
+    backfaceThreshold: 100
+--- !u!104 &2
+RenderSettings:
+  m_ObjectHideFlags: 0
+  serializedVersion: 7
+  m_Fog: 0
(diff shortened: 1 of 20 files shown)


xEvoGx:
Damn son! Awesome! :smile:

mimminito:
Doesnt include the fancy transitions like Budget Cuts and Vertigo do, but its a base to work from.

OmegaJak:
This is awesome! I've got a few thoughts/questions though:

- I couldn't find a way to cancel the teleport once it was initiated. Am I missing something or was this not included?
- Initially I didn't notice, but when doing frequent teleporting the load time of the viewport gets a bit prohibitive. I understand there might be limitations to the rendertexture or whatever, but it should be possible to teleport before the viewport has loaded. It should be possible pretty much instantly imo
- The camera view in the viewport seemed to clip through objects I was teleporting near pretty frequently, so I tried reducing the clipping plane to 0.1 and that helped a lot. Maybe this change should be made for the default prefab?

Otherwise, as far as being closer to the actual Buget Cuts teleporter, the biggest diffrences I noticed were the lack of being able to shoot a ball (which should probably be a separate script & PR), as well as the camera not moving in space when the controller does, only rotating. Again, though, that could be a separate PR (hell, maybe I'll add it even). This is a great starting point for sure though.

OmegaJak:
Also maybe the example should be named "038_Controller_BudgetCutsTeleport" for consistency? Though really idk why they're all named that way.

thestonefox:
@OmegaJak it would tecnically be `038_CameraRig_BudgetCutsTeleporter` but also I think perhaps best to come up with another name instead of BudgetCuts (otherwise people may draw comparisons and then discount it if it's not the same as the BudgetCuts one).

The reason they have the `Controller` `CameraRig` etc, is to show where the scripts the scene is largely about are being applied to :)

thestonefox:

PR #458, opened 2016-08-24
feat(Interaction): add helper for interaction unity events

Blueteak:
This adds a helper script to use on Interactable Objects. It will connect all delegates to UnityEvents to use inside the inspector, but is not integrated directly into the Interactable Object script for performance and clutter reasons.

### DIFF
--- a/Assets/VRTK/Scripts/Helper/UnityEvents.meta
+++ b/Assets/VRTK/Scripts/Helper/UnityEvents.meta
@@ -1,0 +1,9 @@
+fileFormatVersion: 2
+guid: bc66491294ded4eb49bc4e7882357d67
+folderAsset: yes
+timeCreated: 1472142091
+licenseType: Pro
+DefaultImporter:
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/Assets/VRTK/Scripts/Helper/UnityEvents/VRTK_InteractableObject_UnityEvents.cs
+++ b/Assets/VRTK/Scripts/Helper/UnityEvents/VRTK_InteractableObject_UnityEvents.cs
@@ -1,0 +1,70 @@
+﻿using UnityEngine;
+using System.Collections;
+using UnityEngine.Events;
+using VRTK;
+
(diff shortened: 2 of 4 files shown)


Assets/VRTK/Scripts/Helper/VRTK_InteractionEvents.cs
```diff
@@ -0,0 +1,68 @@
...
+	[System.Serializable]
+	public class UnityObjectEvent: UnityEvent<GameObject> {};
+	public UnityObjectEvent OnTouch;
+	public UnityObjectEvent OnUntouch; 
+	public UnityObjectEvent OnGrab; 
+	public UnityObjectEvent OnUngrab; 
+	public UnityObjectEvent OnUse; 
+	public UnityObjectEvent OnUnuse;  
+
+	void Awake () {
```
thestonefox: should this be in the `OnEnable` and then a `OnDisable` to clean them up?

Assets/VRTK/Scripts/Helper/VRTK_InteractionEvents.cs
```diff
@@ -0,0 +1,68 @@
...
+		OnUse.Invoke(e.interactingObject);
+	}
+
+	void Unuse(object o, InteractableObjectEventArgs e)
+	{
+		OnUnuse.Invoke(e.interactingObject);
+	}
+	
+	// Update is called once per frame
+	void OnDestroy () {
```
thestonefox: If this was done in OnDisable then you wouldn't need it done in Destroy

thestonefox:
I like this concept. However I think it shouldn't be hidden away in a "helper" folder. I think a new folder should be created in the `Scripts` folder called something like `UnityEventHelpers` and then follow a naming convention of `UnityEventHelpers/VRTK_InteractableObject_UnityEvents` then there should be a point in the documentation perhaps in the script section after `Class Events` called `Unity Events` that describes where the script is to add unity events and a brief explanation of each exposed event.

Do you wanna do this? Or shall I do it as it's a bit more involved work wise :)

PR #475, opened 2016-08-28
fix(Interaction): improve door gizmos

10FingerArmy:
Improves visualizing the opening direction of doors a lot. There were
quite some situations so far where gizmos would point into the wrong
direction. These should be much fewer now.

In addition when a hinge joint is added manually this is now taken as
the input for the gizmos and the direction is also derived from it
resulting in a much better editor experience.


10FingerArmy:
tested in all my 3 projects, all doors now work for me (which didn't before) and old ones didn't break.

PR #483, opened 2016-08-30
feat(Teleport): add dash teleport

mbbmbbmm:
Dashing Teleport inspired by the implementation of realities.io.
When dashing through objects their renderers can be turned off.
An example scene has been added to show it in action.

### DIFF
--- a/Assets/VRTK/Scripts/VRTK_DashTeleport.cs
+++ b/Assets/VRTK/Scripts/VRTK_DashTeleport.cs
@@ -1,0 +1,94 @@
+namespace VRTK
+{
+    using UnityEngine;
+    using System.Collections;
+
+    public class VRTK_DashTeleport : VRTK_HeightAdjustTeleport
+    {
+        public delegate void DashAction(RaycastHit[] hit);
+        public static event DashAction OnWillDashThruObjects;
+        public static event DashAction OnDashedThruObjects;
+        public float normalLerpTime = 0.1f; // 100ms for every dash above minDistanceForNormalLerp
+        public float minSpeedMps = 50.0f; // clamped to minimum speed 50m/s to avoid sickness
+        public float capsuleTopOffset = 0.2f;
+        public float capsuleBottomOffset = 0.5f;
+        public float capsuleRadius = 0.5f;
+
+        // The minimum distance for fixed time lerp is determined by the minSpeed and the normalLerpTime
+        // If you want to always lerp with a fixed mps speed, set the normalLerpTime to a high value
+        private float minDistanceForNormalLerp;
+        private float lerpTime = 0.1f;
+
+        protected override void Awake()
+        {
+            base.Awake();
+            minDistanceForNormalLerp = minSpeedMps * normalLerpTime; // default values give 5.0f
+        }
+
+        protected override void SetNewPosition(Vector3 position, Transform target)
+        {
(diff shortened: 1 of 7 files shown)


thestonefox:
Looks good. I'll have a proper play when i'm back from vacation (on Saturday) so hopefully it will be a quick merge :)

Assets/VRTK/Scripts/VRTK_DashTeleport.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        private IEnumerator lerpToPosition(Vector3 targetPosition, Transform target)
+        {
+            base.enableTeleport = false;
+            bool gameObjectInTheWay = false;
+
+            // Find the objects we will be dashing through and broadcast them via events
+            Vector3 eyeCameraPosition = eyeCamera.transform.position;
+            Vector3 eyeCameraPositionOnGround = new Vector3(eyeCameraPosition.x, this.transform.position.y, eyeCameraPosition.z);
```
thestonefox: don't need to write `this.transform` should just be `transform`

Assets/VRTK/Scripts/VRTK_DashTeleport.cs
```diff
@@ -0,0 +1,94 @@
...
+                lerpTime = (1 / minSpeedMps) * maxDistance; // clamped to speed for small dashes
+            }
+
+            Vector3 startPosition = new Vector3(transform.position.x, transform.position.y, transform.position.z);
+            float elapsedTime = 0;
+            float t = 0;
+
+            while (t < 1)
+            {
+                this.transform.position = Vector3.Lerp(startPosition, targetPosition, t);
```
thestonefox: don't need to write `this.transform` should just be `transform`

mbbmbbmm:
Yes I had used it intentionally for better understanding but looking at it again I'm not sure why I thought that it was helpful. It's pretty clear without it. Anyway I took it out now.

thestonefox:
I used to use the `this` keyword as well (you'll see in earlier versions of the toolkit code it's there) but I was told not to put it in, so I took it out :)

mbbmbbmm:
Oh I see. :) It's interesting, I hadn't really thought about how questions like this need to be resolved with an open source project.

thestonefox:
It's pretty much ready to go, the one thing i'd say is in the example scene perhaps use the Bezier Pointer rather than the simple pointer (or at least on one of the controllers) what do you think?

mbbmbbmm:
Good idea, I added the Bezier to one of the controllers!

thestonefox:
One last thing, and this is totally optional.

But would you consider doing a bit of extra work on adding a unity event helper as in this PR:

https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/509

Basically, it's a script that you attach to the same object as your delegate event script and it allows you to use unity events.

It's something I'm trying to support as standard.

If it's not making sense or you don't really have time then no worries I'll merge this and add the Unity Event helper afterwards :)

mbbmbbmm:
I looked at it, but unfortunately I am quite new to events in general and don't really understand how to deal with the event args. I'd love to learn it but at the moment the problem is the time. But I promise to examine how you add it so I'll be able do it myself in the future!

thestonefox:
ok no worries, then lets get this merged in!

mbbmbbmm:
Very cool! Thank you!

PR #496, opened 2016-09-01
feat(Interaction): Added script to highlight objects on touch

Mishkun:
VRTK_HighlightOnInteract adds outline to VRTK_InteractableObject's that changes its colour on touch. That outline done IKEA VR App style. It also can show/hide associated tooltip on touch. Easy to use and customize

This is reopenning of my previous pull request, because thing got messy...

### DIFF
--- a/Assets/VRTK/Materials/Resources/OutlineHighlighter.mat
+++ b/Assets/VRTK/Materials/Resources/OutlineHighlighter.mat
@@ -1,0 +1,146 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!21 &2100000
+Material:
+  serializedVersion: 6
+  m_ObjectHideFlags: 0
+  m_PrefabParentObject: {fileID: 0}
+  m_PrefabInternal: {fileID: 0}
+  m_Name: OutlineHighlighter
+  m_Shader: {fileID: 4800000, guid: 98d735b00d428fe4ba5e9858cecfc0c5, type: 3}
+  m_ShaderKeywords: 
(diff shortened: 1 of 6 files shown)


thestonefox:
This has got me thinking about how to re-do object highlighting.

I may not merge this PR (yet anyway) but let's not close it as I think it will form a good basis of the new highlighting solution.

Stick around and we can discuss ideas :)

Mishkun:
@thestonefox what do you think about deprecating old highlighting system and simply integrate this one into VRTK_InteractableObject? 
Or let people choose which one they prefer. Then I think we need to split highlighting from VRTK_Interactable, because it is already a god-class

thestonefox:
@Mishkun I don't think i want to move anything else into the interactable object (there are already issues with things breaking unity batching)

i like your idea of having a separate script that deals with highlighting, but i think i wanna do some ground work on how that could work as a base first then your stuff could go on top of that base.

interactable object is a horribly large class, it needs to go on a diet! :D

Mishkun:
@thestonefox okay, you can call me anytime to work on this issue.
Offtopic:
 Why not to split VRTK_Interactable in three basic scripts "Touchable" "Grabbable" "usable" like it was done with ones for controllers?

thestonefox:
@Mishkun i've considered the split of the interactable object, but they're so damn coupled

i.e.

use requires touch
grab requires touch
use mostly requires grab

it's just easier to handle those sorts of things in one pass and keep one consistent state on the object.

I may give it some more thought at some time though

thestonefox:
The new object highlighters supersede this PR

PR #499, opened 2016-09-02
feat(Tracking): add head-rotation independent hips

Blueteak:
Adds a VRTK_Hips script that allows for head-rotation-independent body tracking for things like gun holsters. The hips will rotate with the players Y rotation, but ignore X and Z rotation which is non-trivial when the HMD is looking directly up/down.

### DIFF
--- a/Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
+++ b/Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
@@ -1,0 +1,62 @@
+﻿// Hip Tracking|Scripts|0135
+namespace VRTK
+{
+	using UnityEngine;
+	using System.Collections;
+
+	/// <summary>
+	/// Hip Tracking attempts to reasonably track hip position in the absence of a hip position sensor.
+	/// </summary>
+	/// <remarks>
+	/// The Hip Tracking script is placed on an empty GameObject which will be positioned at the estimated hip position.
+	/// </remarks>
+
+	public class VRTK_Hip_Tracking : MonoBehaviour
+	{
+		[Tooltip("Distance underneath Player Head for hips to reside.")]
+		public float HeadOffset = -0.35f;
+		[Header ("Optional")]
+		[Tooltip("Optional Transform to use as the Head Object for calculating hip position. If none is given one will try to be found in the scene.")]
+		public Transform headOverride;
+		[Tooltip("Optional Transform to use for calculating which way is "Up" relative to the player for hip positioning.")]
+		public Transform ReferenceUp;
+		private Transform playerHead;
+
+		private void Awake ()
+		{
+			if (headOverride != null) {
+				playerHead = headOverride;
+			} else {
+				playerHead = VRTK_DeviceFinder.HeadsetTransform ();
+			}
+		}
+
+		private void Update ()
+		{
+			if (playerHead == null) {
+				return;
+			}
+			Vector3 up = Vector3.up;
+			if (ReferenceUp != null) {
+				up = ReferenceUp.up;
+			}
+
+			transform.position = playerHead.position + (HeadOffset * up);
+
+			Vector3 forward = playerHead.forward;
+			Vector3 forwardLeveld1 = forward;
+			forwardLeveld1.y = 0;
+			forwardLeveld1.Normalize ();
+			Vector3 mixedInLocalForward = playerHead.up;
+			if (forward.y > 0) {
+				mixedInLocalForward = -playerHead.up;
+			}
+			mixedInLocalForward.y = 0;
+			mixedInLocalForward.Normalize ();
+
+			float dot = Mathf.Clamp (Vector3.Dot (forwardLeveld1, forward), 0f, 1f);
+			Vector3 finalForward = Vector3.Lerp (mixedInLocalForward, forwardLeveld1, dot * dot);
+			transform.rotation = Quaternion.LookRotation (finalForward, up);
+		}
+	}
+}
--- a/Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs.meta
+++ b/Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs.meta
@@ -1,0 +1,12 @@
+fileFormatVersion: 2
+guid: e36dc0fc691424ac6aecdb5cdde2ed13
+timeCreated: 1472831354
+licenseType: Pro
+MonoImporter:
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 


Assets/VRTK/Scripts/Helper/VRTK_Hips.cs
```diff
@@ -0,0 +1,42 @@
...
+	public Transform playerHead;
+	public float HeadOffset = -0.35f;
+	[Header("Optional")]
+	public Transform ReferenceUp;
+
+	void Update()
+	{
+		if(playerHead == null)
+		{
+			return;
```
mattboy64: Consider setting the head in Start/Awake instead of forcing the user to set it themselves. If the user has explicitly set it then obviously you wouldn't want to override it in that case. This would still allow the customization, but removes one bug prone step if the use forgets to manually set it themselves.

Assets/VRTK/Scripts/Helper/VRTK_Hips.cs
```diff
@@ -0,0 +1,55 @@
+﻿using UnityEngine;
```
thestonefox: Should this be in the VRTK namespace?

thestonefox:
You know this is in the `Helper` folder, would it be considered a main script?

I'm guessing you have to add it to a component to get the benefit of it? Similar to the `Headset Collision` script.

Also, it will need the docs adding to the code, either look at some existing code or ask and we'll go through it.

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,73 @@
+﻿namespace VRTK
+{
+	using UnityEngine;
+	using System.Collections;
+
+	public class VRTK_Hip_Tracking : MonoBehaviour {
+
+		/// <summary>
```
thestonefox: this should go above the class (before line 6)

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,73 @@
...
+    	/// Hip Tracking attempts to reasonably track hip position in the absence of a hip position sensor
+    	/// </summary>
+   	 	/// <remarks>
```
thestonefox: bad indentation

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,73 @@
...
+		///
+		/// The Hip Tracking script is placed on an empty GameObject which will be positioned at the estimated hip position
+		/// <para>&#160;</para>
+		/// Variables
+		/// <list type="bullet">
+    	/// <item>Head Offset<description>Distance underneath Player Head for hips to reside.</description> </item>
+		/// <item>Head Override<description>Optional Transform to use as the Head Object for calculating hip position. If none is given one will try to be found in the scene.</description> </item>
```
thestonefox: I dont think these `description` tags would work. Check out the AdaptiveQuality script to see how the lists were done in those comments

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,73 @@
...
+		/// <item>Reference Up<description>Optional Transform to use for calculating which way is "Up" relative to the player for hip positioning.</description> </item>
+    	/// </list>
+		/// </remarks>
+
+		Transform playerHead;
```
thestonefox: need to determine the accessibility of the parameter (and if it's private it needs to go below the public then protected params)

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,73 @@
...
+		public float HeadOffset = -0.35f;
```
thestonefox: need tooltips on the inspector params

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,73 @@
...
+				mixedInLocalForward = -playerHead.up;
+			}
+			mixedInLocalForward.y = 0;
+			mixedInLocalForward.Normalize();
+
+			float dot = Mathf.Clamp(Vector3.Dot(forwardLeveld1, forward), 0f, 1f);
+			Vector3 finalForward = Vector3.Lerp(mixedInLocalForward, forwardLeveld1, dot*dot);
+			transform.rotation = Quaternion.LookRotation(finalForward, up);
+		}
+
```
thestonefox: prolly remove this blank line

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,73 @@
+﻿namespace VRTK
```
thestonefox: Need to include the header comment line, stating script name, category and list position (check another script for an example)

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,73 @@
...
+	public class VRTK_Hip_Tracking : MonoBehaviour {
+
+		/// <summary>
+    	/// Hip Tracking attempts to reasonably track hip position in the absence of a hip position sensor
+    	/// </summary>
+   	 	/// <remarks>
+		///
+		/// The Hip Tracking script is placed on an empty GameObject which will be positioned at the estimated hip position
+		/// <para>&#160;</para>
+		/// Variables
```
thestonefox: You don't need to do the variables here, you can just add tooltips, the inspector params gets built off of those.

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Hip Tracking|Scripts|0135
+namespace VRTK
+{
+	using UnityEngine;
+	using System.Collections;
+
```
thestonefox: remove additional line (only needs 1 line break)

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,66 @@
...
+
+	/// <summary>
+	/// Hip Tracking attempts to reasonably track hip position in the absence of a hip position sensor
+	/// </summary>
+	/// <remarks>
+	///
```
thestonefox: Don't need this blank line, a paragraph is already added when splitting summary and remarks

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Hip Tracking|Scripts|0135
+namespace VRTK
+{
+	using UnityEngine;
+	using System.Collections;
+
+
+	/// <summary>
+	/// Hip Tracking attempts to reasonably track hip position in the absence of a hip position sensor
```
thestonefox: Add a full stop. (period)

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,66 @@
...
+	/// </summary>
+	/// <remarks>
+	///
+	/// The Hip Tracking script is placed on an empty GameObject which will be positioned at the estimated hip position.
+	/// </remarks>
+
+	public class VRTK_Hip_Tracking : MonoBehaviour
+	{
+
+		private Transform playerHead;
```
thestonefox: private variable should come at the end of public, then protected variables (also remove the preceeding line break)

Assets/VRTK/Scripts/VRTK_Hip_Tracking.cs
```diff
@@ -0,0 +1,66 @@
...
+				mixedInLocalForward = -playerHead.up;
+			}
+			mixedInLocalForward.y = 0;
+			mixedInLocalForward.Normalize ();
+
+			float dot = Mathf.Clamp (Vector3.Dot (forwardLeveld1, forward), 0f, 1f);
+			Vector3 finalForward = Vector3.Lerp (mixedInLocalForward, forwardLeveld1, dot * dot);
+			transform.rotation = Quaternion.LookRotation (finalForward, up);
+		}
+
```
thestonefox: remove this line break too

PR #507, opened 2016-09-03
fix(Teleport): add floor height tolerance parameter - resolves #489

### LINKED ISSUE #489: Height Adjust Teleport triggers drop to floor excessively
While testing a script in the climbing example, I had an issue where I was being teleported to the floor instantly while climbing. I traced the issue back to an OnTeleporting callback getting triggered in VRTK_HeightAdjustTeleport.cs. It appears the floorY variable can vary slightly due to noise in either the Raycast or the eyeCamara tracking position. 

The comparison on line 155 needs a small tolerance in order to potentially avoid this issue. I was regularly seeing a previousFloorY = 0.0 and an extremely small floorY ~= 1.0E-7. This means a calculated height difference of a micrometer was enough to trigger a teleport.
https://github.com/thestonefox/SteamVR_Unity_Toolkit/blob/master/Assets/VRTK/Scripts/VRTK_HeightAdjustTeleport.cs#L155

```
                Ray ray = new Ray(eyeCamera.transform.position, -transform.up);
                RaycastHit rayCollidedWith;
                bool rayHit = Physics.Raycast(ray, out rayCollidedWith);
                float floorY = eyeCamera.transform.position.y - rayCollidedWith.distance;

                if (rayHit && ValidLocation(rayCollidedWith.transform, rayCollidedWith.point) && !FloorIsGrabbedObject(rayCollidedWith))
                {
                    var floorDelta = currentRayDownY - floorY;
                    currentFloor = rayCollidedWith.transform.gameObject;
                    currentRayDownY = floorY;

                    float fallDistance = transform.position.y - floorY;
                    bool shouldDoGravityFall = useGravityFall && (playerPresence.IsFalling() || fallDistance > gravityFallHeight);

                    if (withBlink && !shouldDoGravityFall && (floorDelta > blinkYThreshold || floorDelta < -blinkYThreshold))
                    {
                        Blink(blinkTransitionSpeed);
                    }

                    if (floorY != previousFloorY) // Could be replaced with (Mathf.Abs(floorY - previousFloorY) < minHeightThreshold)
                    {
                        if (shouldDoGravityFall)
                        {
                            playerPresence.StartPhysicsFall(Vector3.zero);
                        }
                        else // teleport fall
                        {
                            Vector3 newPosition = new Vector3(transform.position.x, floorY, transform.position.z);
                            var teleportArgs = new DestinationMarkerEventArgs
                            {
                                destinationPosition = newPosition,
                                distance = rayCollidedWith.distance,
                                enableTeleport = true,
                                target = currentFloor.transform
                            };
                            OnTeleporting(gameObject, teleportArgs);
                            SetNewPosition(newPosition, currentFloor.transform);
                            OnTeleported(gameObject, teleportArgs);
                        }
                    }
                }
```

thestonefox:
The `Floor Height Tolerance` parameter is used to determine how much
of a change between the current floor `y` and the previous floor `y`
has happened before considering it valid to drop to the nearest floor.

This setting can be used to prevent unnecessary teleport events when
the floor below the user is uneven but should not be considered
different enough to force the user to drop to a new location on such
a small variation change in their `y` position.

### DIFF
--- a/Assets/VRTK/Scripts/VRTK_HeightAdjustTeleport.cs
+++ b/Assets/VRTK/Scripts/VRTK_HeightAdjustTeleport.cs
@@ -18,6 +18,7 @@ namespace VRTK
         public bool useGravity = true;
         public float gravityFallHeight = 1.0f;
         public float blinkYThreshold = 0.1f;
+        public float floorHeightTolerance = 0.001f;
 
         private float currentRayDownY = 0f;
         private GameObject currentFloor = null;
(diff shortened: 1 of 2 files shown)


Assets/VRTK/Scripts/VRTK_HeightAdjustTeleport.cs
```diff
@@ -152,7 +159,7 @@ private void DropToNearestFloor(bool withBlink, bool useGravityFall)
                         Blink(blinkTransitionSpeed);
```
thestonefox: Should this be done inside the `FloorHeightChanged` conditional?

What's the point in blinking if you're not going to change position?

PR #517, opened 2016-09-05
feat(Interaction): allow to throw kinematic objects

TMaloteaux:
add an option to allow to throw kinematic objects
instead of letting them floating in the air.


thestonefox:
I'm not sure this is the best way to solve this issue.

would it not be better if you had a kinematic object that you wanted to then throw, to extend the Interactable Object script and override the `LoadPreviousState` method and then force the rigidbody kinematic to false?

There is the potential that I may want to drop an object that is kinematic by default and have it return back to its kinematic state, but if the controller is moving at a certain velocity then i may want it not to be kinematic so i can throw it, but then have the object go back to being kinematic when it's velocity is zero.

All of the above is quite custom to a certain use-case though and therefore is better done by extending the interactable object rather than adding more options to the base class.

TMaloteaux:
I agree with you. 
It's what i wanted to do in the first place but the protection level of almost all variables used in the method `LoadPreviousState` are preventing me to implement this.
I can modify theses variables if it's ok for you.

thestonefox:
What private variables do you need access to? they may be better being protected if you think they need accessing in an extended class.

TMaloteaux:
All of them. 
`previousParent`
`forcedDropped`
`previousKinematicState`
`previousIsGrabbable`

I can't call the base method and do my stuff after that because the base class is setting the kinematic value and it seems that is breaking the force added. So the objet fall instead of been throwed.
So i must re implement all the method.

I come from the java world and i'm not used yet to the C# good practices. 
Should not be all the needed fields of a virtual method protected?

thestonefox:
yeah makes sense to have these protected, i'll update them, give me a mo :)

thestonefox:
https://github.com/thestonefox/SteamVR_Unity_Toolkit/commit/1384687c5cd782dcdb58d5421a0f120759ffdb68 done :)

TMaloteaux:
Super thank you :)

PR #536, opened 2016-09-08
feat(Climbing): Added Climbing Stamina

sidcitris:
Add climbing stamina class that extends the existing VRTK_PlayerClimb
class. The VRTK_PlayerClimbStamina script is added to the CameraRig in
of place the previous climbing script.

The players min reach and max reach can be set where the distance in
the xz plane between the controller and the head determine the effect
on stamina. A distance less than min reach will not decrease stamina
while a distance beyone max reach will apply the maximum damge to
stamina.

The player can regenerate stamina when either no hands are holding a
climbable object, or when both hands are holding climbable objects.
The VRTK currently doesn't support grabbing one object with both hands
but should work when it is implemented.

Example 039_CameraRig_ClimbingStamina has been added to demonstrate
the climbing stamina as well as a useful grip stamina visualization
prefab which has been added.

### DIFF
--- a/Assets/VRTK/Examples/039_CameraRig_ClimbingStamina.unity
+++ b/Assets/VRTK/Examples/039_CameraRig_ClimbingStamina.unity
@@ -1,0 +1,16690 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!29 &1
+SceneSettings:
+  m_ObjectHideFlags: 0
+  m_PVSData: 
+  m_PVSObjectsArray: []
+  m_PVSPortalsArray: []
+  m_OcclusionBakeSettings:
+    smallestOccluder: 5
+    smallestHole: 0.25
+    backfaceThreshold: 100
+--- !u!104 &2
+RenderSettings:
+  m_ObjectHideFlags: 0
+  serializedVersion: 7
+  m_Fog: 0
+  m_FogColor: {r: 0.5, g: 0.5, b: 0.5, a: 1}
+  m_FogMode: 3
+  m_FogDensity: 0.01
+  m_LinearFogStart: 0
+  m_LinearFogEnd: 300
+  m_AmbientSkyColor: {r: 0.212, g: 0.227, b: 0.259, a: 1}
+  m_AmbientEquatorColor: {r: 0.114, g: 0.125, b: 0.133, a: 1}
+  m_AmbientGroundColor: {r: 0.047, g: 0.043, b: 0.035, a: 1}
+  m_AmbientIntensity: 1
+  m_AmbientMode: 0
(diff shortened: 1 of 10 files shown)


sidcitris:
Attempt 2... I merged the latest changes and cleaned up the commit mess I made... Let me know if you find any issues.

mattboy64:
First of all great work and thank you for contributing! 

I've got a few suggestions that I think can help. There seems to be 2 features here, 2 handed functionality and stamina. The 2 handed system I think needs to be part of the base VRTK_PlayerClimb class. We'd want that regardless of the use of stamina. It might be nice to define a climbable object as a 2 handed object or a 1 handed one, but that can wait for another time.  

That leaves the stamina. As far as I can tell that is a simple distance based check and forces an ungrab. Extending the PlayerClimb class to do that is a little overkill. I'd use a component based design here as it seems perfect for this. There's no real dependency on the PlayerClimb other than the direction and distance. Initializing a GripStamina type object with a GetComponent<VRTK_PlayerClimb>() and a few queries will remove the tight coupling in the current implementation and reduce feature conflicts.

Let me know if you need help with any of that.

sidcitris:
Thanks for taking a look at it. I think you are right about it being more like 2 features. I didn't want to add the stamina as part of the VRTK_PlayerClimb script, but as I realized I needed to support both hands I didn't even think about adding that to the base script. I have only been using Unity/C# for about 2 weeks now so still learning a bit. I will take a shot at implementing it the way you put it as I think that makes sense. Should I break it into 2 pull requests when I get it working, or should I implement them both at the same time for simplicity?

mattboy64:
No problem @sidcitris. Thanks for being willing to contribute. A lot of people think of pull requests as a take it or leave it type thing, where in actuality it's a place to start a discussion. VRTK is partially such a great tool because of the high coding and design standards. It took a little while to get the initial climbing merged, but, even though it's not perfect, it's a much cleaner design than if my first version was taken. Don't worry about the inexperience with C#, it looks let good to me. Thank you again for helping make climbing better!

xEvoGx:
Has this ever made into the Master?

sidcitris:
I was going to separate the pull into two features, and implement the suggestions matt made above, but have been busy with work. I would like to finish this though...

BinaryLegend:
Any chance of this (both stamina and 2 handed climbing) getting into Master?

thestonefox:
@BinaryLegend For two handed grabbing what about this for a potential solution:

A secondary grab action that when engaged is just listening for the primary grabbed controller to stop grabbing, when the primary grab stops grabbing, if the secondary controller is in a grabbed state then it just takes over the grab action.

That would solve the problem of if you grab with both controllers and let go of the grab on the first controller, then the second controller would take over the grab action without having to re-press the grab button.

thestonefox:

PR #537, opened 2016-09-08
fix(Examples): added error checking if no gameobject named Bullet exi…

peroht:
…sts as child

When there is no gameobject named Bullet which is referenced by this script
the instantiation fails. Added a simple Debug.LogError() to notify user about this.


GitCop:

 * Commit: 5790441cd160353dad460a2b613ae9019f0229e3

---

PR #540, opened 2016-09-08
Chore/namespacing - resolves #521

### LINKED ISSUE #521: Namespace collisions
While trying to add VRTK to my existing game, i had to remove all examples because i had collisions with some classes (`Bow`, `BowAim`, `Sword`) but the more problematic one is the collision with `RadialMenu` because it's not part of the examples. Anyway, deleting stuff from the plugin is not a solution as i would have to do it again at every update.

Those classes should be namespaced like the rest of this library to avoid this kind of problems, don't you think ?

thestonefox:


jcrombez:
cool :)

PR #550, opened 2016-09-09
feat(Pointer): rotate bezier pointer cursor to hit surface

Blueteak:
Bezier Pointer base rotates to match normal direction of collision surface.


thestonefox:
superseded by https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/553

PR #562, opened 2016-09-12
fix(Controller): ensure controller colliders scale

TMaloteaux:
The colliders added automatically by the script
are now scaled properly according to the parent scale

### DIFF
--- a/Assets/VRTK/Scripts/VRTK_InteractTouch.cs
+++ b/Assets/VRTK/Scripts/VRTK_InteractTouch.cs
@@ -384,7 +384,7 @@ namespace VRTK
         {
             if (customRigidbodyObject == null)
             {
-                controllerCollisionDetector = Instantiate(defaultColliderPrefab, transform.position, transform.rotation) as GameObject;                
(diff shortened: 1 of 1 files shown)


Assets/VRTK/Scripts/VRTK_InteractTouch.cs
```diff
@@ -384,8 +384,9 @@ private void CreateTouchCollider()
         {
             if (customRigidbodyObject == null)
             {
-                controllerCollisionDetector = Instantiate(defaultColliderPrefab, transform.position, transform.rotation) as GameObject;
+                controllerCollisionDetector = Instantiate(defaultColliderPrefab, transform.position, transform.rotation) as GameObject;
```
thestonefox: seems to be a lot of added whitespace at the end of this line

TMaloteaux: Oops. will fix that. ^^

thestonefox:
For some reason github is saying this is 2 commits now.

Do you have 2 commits in your git log? if so you'll want to squash them into 1 commit and force push

PR #563, opened 2016-09-12
fix(Controller): ensure controller colliders scale

TMaloteaux:
fix(Controller): ensure controller colliders scale

The colliders added automatically by the script
are now scaled properly according to the parent scale


TMaloteaux:
Just 1 commit now.

PR #570, opened 2016-09-14
fix(Colliders): ensure colliders scale properly

mattboy64:
The custom rigid body collider object needs to be scaled to match
the parent object's scale.

The PlayerPresence body collider now scales properly.

### DIFF
--- a/Assets/VRTK/Scripts/VRTK_InteractTouch.cs
+++ b/Assets/VRTK/Scripts/VRTK_InteractTouch.cs
@@ -401,6 +401,7 @@ namespace VRTK
                 {
                     controllerCollisionDetector = Instantiate(customRigidbodyObject, transform.position, transform.rotation) as GameObject;
                     controllerCollisionDetector.transform.SetParent(transform);
+                    controllerCollisionDetector.transform.localScale = transform.localScale;
                     destroyColliderOnDisable = true;
                 }
             }
--- a/Assets/VRTK/Scripts/VRTK_PlayerPresence.cs
+++ b/Assets/VRTK/Scripts/VRTK_PlayerPresence.cs
@@ -297,7 +297,7 @@ namespace VRTK
         private void UpdateCollider()
         {
             var playAreaHeightAdjustment = 0.009f;
-            var newpresenceColliderYSize = (headset.transform.position.y - headsetYOffset) - transform.position.y;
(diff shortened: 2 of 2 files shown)


GitCop:

 * Commit: 3279d6f9b0f042e14fac3a3d69520e2eadf1366b

---

mattboy64:
Noticed my problem was fixed 2 days ago, but I think scaling needs to happen if you set the collider yourself. No good way for me to review this yet, so I'll leave it up to you if this is needed or not.

mattboy64:
fixed an issue with the PlayerPresence collider too. Appended it on here. Hopefully I didn't screw it up.

Assets/VRTK/Scripts/VRTK_InteractTouch.cs
```diff
@@ -401,6 +401,7 @@ private void CreateTouchCollider()
                 {
                     controllerCollisionDetector = Instantiate(customRigidbodyObject, transform.position, transform.rotation) as GameObject;
                     controllerCollisionDetector.transform.SetParent(transform);
+                    controllerCollisionDetector.transform.localScale = transform.localScale;
```
thestonefox: hummmmmmmmmmmm, if it's custom should you be doing anything automatic to it?

what if you wanted to provide a custom collider and specifically have it smaller (or the size you created it at)?

mattboy64: I believe once you set the parent it will try and rescale it down to match. This is why the controllers scale that was fixed 2 days ago broke.

Maybe it would be safer to store the localScale before setting the parent, and then setting it to that. What do you think?

thestonefox: if parenting rescales to match, is there a nee to specifically set the localScale to match the parent?

mattboy64: It scales it down. I guess it depends on how you set it up. If you think of it like the controllers that auto scale up and down with the camera scale then yes you'd want to. If you want it to stay the same size no matter what scale the CameraRig is then no. Personally I think it makes sense to do it.

I was thinking of adding this 
```controllerCollisionDetector = Instantiate(customRigidbodyObject, transform.position, transform.rotation) as GameObject;
                    Vector3 originalScale = controllerCollisionDetector.transform.localScale;
                    controllerCollisionDetector.transform.SetParent(transform);
                    controllerCollisionDetector.transform.localScale = originalScale;```

PR #583, opened 2016-09-18
fix(Controls): replace lever to use rotator track - resolves #582

### LINKED ISSUE #582: Consider changing lever to use Rotation Track rather than Track Object
Apparently it works better with Rotation Track

cameronoltmann:
The lever now uses the Rotator_Track as the Track_Object grab mechanic doesn't work as well for moving the lever.


GitCop:

 * Commit: a335c761ce40195f0064d6229e4848e9a90453c8

---

PR #589, opened 2016-09-20
Adaptive Quality - Service Pack 1

bddckr:
This cleans up the code and docs a bit, improves the performance a smaller bit and introduces dynamic adjusting of the render target resolution.


xEvoGx:
Does AdaptiveQuality only work in Forward? (I'm using deferred for image effects)

Thank you!

bddckr:
> Adjusting the render target resolution seems to be hitting a Unity bug:
> Sometimes when the resolution changes one can see the old frame flashing inside the Vive. I have no idea how to fix this.
> 
> Until someone tests and tells me that it's just the example scene dropping too many frames that causes this, **I'd say this PR shouldn't be merged yet**.
> There a some bugs filed with Unity regarding the render scales, so maybe this is just another one of those bugs.

This actually works fine for me with the latest commits on Unity 5.4.2f1 now. Someone else needs to test if it works for them, too. Then this PR should be ready to merge.

PR #592, opened 2016-09-21
feat(Pointer): pass RaycastHit struct on destination marker events

thestonefox:
The RaycastHit struct is now passed in the Destination Marker payload
and the RaycastHit.collider is also used to check if a new object is
being collided with as using RaycastHit.transform only checks the top
level rigidbody for changes and not the sub colliders.


getwilde:
That's great news! Thank you.

PR #593, opened 2016-09-23
feat(InteractGrab) add toggle to auto-scale throw velocity with player

mattboy64:
A boolean toggle is now used to determine if the user wants the throw
velocity to scale up and down with the camera rig. To get around this
issue before, you had to manually enter the scale in the mulitpier
slot which is error prone and a bit confusing. This toggle keeps all
velocities used to throw in true world space coordinates, so dynamic
size adjustments and throwing just work without any additional input
needed from the developer.

### DIFF
--- a/Assets/VRTK/Scripts/VRTK_InteractGrab.cs
+++ b/Assets/VRTK/Scripts/VRTK_InteractGrab.cs
@@ -37,6 +37,8 @@ namespace VRTK
         public float grabPrecognition = 0f;
         [Tooltip("An amount to multiply the velocity of any objects being thrown. This can be useful when scaling up the `[CameraRig]` to simulate being able to throw items further.")]
         public float throwMultiplier = 1f;
+        [Tooltip("Automatically scale the throw velocity up and down to match the scale of the `[CameraRig]`.")]
+        public bool throwScalesWithPlayer = true;
         [Tooltip("If this is checked and the controller is not touching an Interactable Object when the grab button is pressed then a rigid body is added to the controller to allow the controller to push other rigid body objects around.")]
         public bool createRigidBodyWhenNotTouching = false;
 
@@ -323,7 +325,9 @@ namespace VRTK
 
             if (origin != null)
             {
-                rb.velocity = origin.TransformDirection(velocity) * (throwMultiplier * objectThrowMultiplier);
+                Vector3 worldSpaceVelocity = throwScalesWithPlayer ? origin.TransformVector(velocity) : origin.TransformDirection(velocity);
+
+                rb.velocity = worldSpaceVelocity * (throwMultiplier * objectThrowMultiplier);
                 rb.angularVelocity = origin.TransformDirection(angularVelocity);
             }
             else


GitCop:

 * Commit: 6fab5ac9bfa6e2cdd2b875aa1b3656af76c08aec

---

Assets/VRTK/Scripts/VRTK_InteractGrab.cs
```diff
@@ -37,6 +37,8 @@ public class VRTK_InteractGrab : MonoBehaviour
         public float grabPrecognition = 0f;
         [Tooltip("An amount to multiply the velocity of any objects being thrown. This can be useful when scaling up the `[CameraRig]` to simulate being able to throw items further.")]
         public float throwMultiplier = 1f;
+        [Tooltip("Automatically scale the throw velocity up and down to match the scale of the `[CameraRig]`.")]
+        public bool throwScalesWithPlayer = true;
```
thestonefox: I'm wondering rather than just having a bool to do this on `velocity` we have 2 enum drop downs that provide options for both velocity and angular velocity so you can cater to whatever need you want.

Alternatively, we could look at doing this without additional parameters? You inherit the script and override the method? Or composition that you add a "throw strategy" script  into the mix?

what do you think about any of that?

mattboy64: I don't think doing it on your angular velocity is correct, since no matter the size, it should stay constant.

thestonefox: but is there a need where people may want it to be different? It used to be the other way before and it was changed because under certain circumstances it was preferred the current way.

I'm just thinking is it better to make it all optional now?

The alternative is we make a dropdown that has "throw strategies" and they basically point to a script that calculates whatever you want it to, we build some default ones, but people can write their own if they want.

I can try doing a PR with the idea if you want?

mattboy64: Eh, it's up to you, but that seems a bit much. Even having the boolean at all is a compromise for me. I get that hypothetically you might want to set an exact max velocity. TransformDirection() used with the multiplier gives you that. In reality however, I've found that a percentage based approach gives much better results. TransformVector() with the multiplier gives you that option (e.g. throw at 70% power).

Is there a project that is using this? Cause you really really don't want to change it 😩

thestonefox: so do you think it's worth then just doing TransformVector and tell people if they want it to work the way it does now then they set their multiplier to like 0.7 or 0.5 to replicate that feeling?

I'd rather do that and see how it goes, I'm just loathed to keep adding options to things, look at the interactable object and pointer scripts, they're full of options and it makes my eyes hurt :(

mattboy64:
Yeah, that sounds best to me. The nice thing about doing it that way is the behavior is exactly the same whether you are scaled or not. Do you want me to change to the PR?

Assets/VRTK/Scripts/VRTK_InteractGrab.cs
mattboy64: Yeah, that sounds best to me. The nice thing about doing it that way is the behavior is exactly the same whether you are scaled or not. Do you want me to change to the PR?

thestonefox: Do you want to close this PR and just do a new one with the one word change? I'm not fussed which way round it's done :)

mattboy64: A simplified version has been submitted in #599. Closing this one out accordingly.

PR #594, opened 2016-09-23
feat(ControllerEvents): get controller magnitude for swing speed

pargee:
Apologies if this isn't done right -- my first pull request on GH! This was brought up as a question on Slack (and one of my own yesterday). It might be too self-explanatory, but as someone who didn't know Vector3.Magnitude was available, I'm hoping to remove any headaches others who want to use this for items like sword swing detection and other motion-based gestures.

Also removed a few double semicolons in the existing code.

### DIFF
--- a/Assets/VRTK/Scripts/VRTK_ControllerEvents.cs
+++ b/Assets/VRTK/Scripts/VRTK_ControllerEvents.cs
@@ -579,7 +579,8 @@ namespace VRTK
         /// The GetMagnitude method is used to return the current magnitude of the physical game controller. This can be useful for determining the speed at which the controller is being swung.
         /// </summary>
         /// <returns>A float illustrating the length of the vector based on current real world physical controller velocity.</returns>
-        public float GetMagnitude() {
+        public float GetMagnitude()
+        {
             SetVelocity();
             return controllerMagnitude;
         }
--- a/Assets/VRTK/Scripts/VRTK_ControllerEvents.cs
+++ b/Assets/VRTK/Scripts/VRTK_ControllerEvents.cs
@@ -305,6 +305,7 @@ namespace VRTK
         private float hairTriggerDelta;
         private Vector3 controllerVelocity = Vector3.zero;
         private Vector3 controllerAngularVelocity = Vector3.zero;
+        private float controllerMagnitude = 0;
 
         public virtual void OnTriggerPressed(ControllerInteractionEventArgs e)
         {
@@ -572,6 +573,15 @@ namespace VRTK
         {
             SetVelocity();
             return controllerAngularVelocity;
+        }
+
+        /// <summary>
+        /// The GetMagnitude method is useful for getting the current magnitude of the physical game controller. This can be useful for determining the speed at which the controller is moving.
+        /// </summary>
+        /// <returns>A float illustrating the length of the vector based on current real world physical controller velocity.</returns>
+        public float GetMagnitude() {
+            SetVelocity();
+            return controllerMagnitude;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: e421af72843e6bf449084a6e0080d12b8b69fbb0

---

GitCop:

 * Commit: e421af72843e6bf449084a6e0080d12b8b69fbb0

 * Commit: 2eedea656bb37a614dd7c01facb0e435723f1b8b

---

mattboy64:
Looks like GitCop got ya again ;)

Assets/VRTK/Scripts/VRTK_ControllerEvents.cs
```diff
@@ -575,6 +576,15 @@ public Vector3 GetAngularVelocity()
         }
 
         /// <summary>
+        /// The GetMagnitude method is used to return the current magnitude of the physical game controller. This can be useful for determining the speed at which the controller is being swung.
+        /// </summary>
+        /// <returns>A float illustrating the length of the vector based on current real world physical controller velocity.</returns>
+        public float GetMagnitude() {
```
mattboy64: curly bracket needs to be on it's own line.

GitCop:

 * Commit: e421af72843e6bf449084a6e0080d12b8b69fbb0

 * Commit: 2eedea656bb37a614dd7c01facb0e435723f1b8b

---

Assets/VRTK/Scripts/VRTK_ControllerEvents.cs
pargee: Thanks, good catch! I had it set but VS snuck it back to inline.

thestonefox:
What's the benefit of having this method as an explicit thing in the script?

Can you not just do this yourself by getting the Velocity and then calling magnitude on the return in your own code?

pargee:
It would just be a quality of life change - feel free to discard if necessary. I've always needed the magnitude rather than raw velocity for the interactions I've built so far (the same with using a Rigidbody's velocity.magnitude).

I went through and originally did the formula for getting magnitude from velocity sqrt(x² + y² + z²); not knowing there was a vec3 operation (since mag has always been fed to me :P), so I thought this might be helpful to others and save them 10 mins of poking how to calc mag from vel, heh.

It's simple enough as you said, so I can just keep it in my own  helper if you'd rather not expand with those types of features :) Thanks for lookin!

thestonefox:
I'll ask in the slack channel, we'll get an idea of what people think about the feature, see if it's something people would want

mattboy64:
Square root isn't the fastest thing in the world, so calculating this often could have some performance considerations (see https://docs.unity3d.com/ScriptReference/Vector3-sqrMagnitude.html).

If this stays, I'd pull it out of the SetVelocity() function....actually you can just do this and avoid the step completely:

```
public float GetMagnitude()
{
      return controllerVelocity.magnitude;
}

public float GetMagnitudeSqr()
{
    return controllerVelocity.sqrMagnitude;
}
``` 

Looking at it know, it seems a little unnecessary since it's only one line though......either way.

thestonefox:
@mattboy64 is it worth having those additional methods though that literally just call the same method again?

mattboy64:
Probably not since it's kind of redundant.

thestonefox:
Is it worth just closing this PR then?

mattboy64:
The only situation where I could see this being useful is if the user didn't know that the magnitude of the velocity is what they need. That's going to be useful for some people, but I don't how many, or if it's worth spelling out. What did slack say?

thestonefox:
I didn't ask yet, i should I suppose.

mattboy64:
My vote is to just close it out. I appreciate the contribution, but keeping the code base lean and simple feels more important to me, especially since this would just be a one line function.

Issue #453, opened 2016-08-22
[Bug] 037 ClimbingFalling - Can Climb Down Into Geometry...


xEvoGx:
...and fall out of the level and into the void. Pretty self explanatory, but if you'd like to see, grab any _climbable'_ and drag yourself into the ground...voila bye bye cruel world.

Never mind that I feel this extends Issue #407 , and would like some collision safety whilst mantling about

Cheers

xEvoGx:
For a possible solution, we could **Raycast** downward from what would be the _Player Height_(the HMD) during a climb to determine if the player was pushing himself into the surface he is currently standing on (mind you we're not just talking about the ground, but any solid surface the player may stand on). 

We also have to take into account if the player is crouching, so doing cheap and dirty math in my head, If we're climbing, and there's a surface below us (Raycast), Y can only be a positive number. If Y is less than the starting value (player height upon initiating climbing), then the player is trying to go into the ground...reset position. 

Now this does bring up the issue of descending, such as down a ladder. Either you educate the player to lean over the edge a bit and then initiate the climb, or we have a threshold before the reset occurs; a grace period if you will before we force the position back to safety. This was players can start a descent, with the raycast likely detecting a -Y value initially before then being over open air (this is good) and they can climb down.

Thoughts?

xEvoGx:
**Small Update:** 

Taking a quick look at the **VRTK_Player_Climb** script, during the initialization of the required scripts, this one...

`controllerManager = FindObjectOfType<SteamVR_ControllerManager>();
            var headCamera = VRTK_DeviceFinder.HeadsetTransform();
            collisionFade = headCamera.GetComponent<VRTK_HeadsetCollisionFade>();
            if (collisionFade == null)
            {
                collisionFade = headCamera.gameObject.AddComponent<VRTK_HeadsetCollisionFade>();
            }`

...clearly looks for the HeadsetTransform, yet when the scene is played, this component is added to the _Camera (eye)_. If you look at **017 ...TouchpadWalking**, _VRTK_Headset Collision Fade_ is assigned (correctly?) to _Camera (head)_. 

I'll try to test all this later today (I'm only perusing the scripts, away from my headset), but this sounds like a good reason why both this issue and #407 are having problems.

Cheers

HyppoNymous:
`...and fall out of the level and into the void. Pretty self explanatory, but if you'd like to see, grab any climbable' and drag yourself into the ground...voila bye bye cruel world.`

[...]

PR #595, opened 2016-09-23
feat(Climbing): climb release can height adjust or physics fall

mattboy64:
Instead of assuming a physics based fall is required, the climbing
object now checks if we at or below ground level. If it is, it will
allow the VRTK_HeightAdjustTeleport object to move you up rather
than assuming you must enter a physics falling state.

The ground plane of example 37 (ClimbingFalling) now uses a box
collider on the ground. This should remove the falling through the
floor bug that was reported.


mattboy64:
This should fix #453

mattboy64:
@thestonefox let me know when you have a change to take a look at this. It should fix most problems with climbing onto ledges as well as falling through the world.

thestonefox:
Just lowercase the commit subject line so it should be:

`feat(Climbing): climb release can height adjust or physics fall`

Then looks good to merge

mattboy64:
Done. Thanks fox

PR #607, opened 2016-09-29
fix(Simulator): inconsistent vector addition.

insominx:
Fixed inconsistency in vector addition caused by adding a vector in world space (the camera forward) to a vector in local space (the camera rig's basis vectors).  Doing the math in the same space gives the intended result.


thestonefox:
Gitcop should have jumped in here:

the git commit subject line needs the full stop removing so it should read:
`fix(Simulator): inconsistent vector addition`

Also, it seems that the git commit body doesn't have any line breaks. No line should exceed 72 characters.

If you can `git commit --amend` this commit message and then force push your branch back up (don't make a new commit) then I can look to merge it.

PR #614, opened 2016-09-30
Feat/interactions with pointers - resolves #597

### LINKED ISSUE #597: Highlighting with Simple Pointer
Is there a way to use the simple pointer to highlight interactable objects? The examples only contain ways to highlight an object upon touching the object.

thestonefox:


getwilde:
Cool, I'll have to take a look at this. Thanks @thestonefox.

PR #628, opened 2016-10-06
feat(Locomotion): translate play area jump into game jump - resolves #393

### LINKED ISSUE #393: Jumping with Headset Movement
I was playing around with the **Climbing/Falling** demo and absolutely loved it. The sensation of dropping off the tower is exhilarating, especially when you have to take a moment to remember the height is not real despite being rather unnatural to '_want_' to drop off from a great height. :smile: 

Coupled with practicing **Jumping** from grabbable surface to grabbable surface, that sensation too feeling great, I was thinking it might be quite cool to have a simple jump mechanic built into the headset that doesn't require actual jumping, nor even a great deal of effort (Threshold parameter). Essentially, if you were to quickly bend your knees and then stand up quickly, you would jump. 

 Mind you in VR you can easily be standing and crouching regularly in some games, so you'd have to test for the magnitude of change along the Y axis in _both directions_. Quick down-to-up movements of the HMD would result in a jump. Quick down movements (dodging incoming fire) without the equally quick upward movement would do nothing. You would have to slightly yet quickly crouch down the quickly stand back up to initiate the jump. 

All these parameters would be tweakable, such as **Jump Window**(if you quickly crouch but then don't quickly stand within the window, you'll be considered crouching and won't jump), **Required Crouch Depth** and both the **Required Magnitude** of the crouch and the subsequent speed of the standing (perhaps these could be tied to a single magnitude, but I was considering that people may do one more quickly than the other).

Thoughts? Could be a fun addition.

ajaxlex:
Adds the player jump script, and demo scene 040 to exercise it.

### DIFF
--- a/Assets/VRTK/Examples/041_CameraRig_PlayerJump.unity
+++ b/Assets/VRTK/Examples/041_CameraRig_PlayerJump.unity
@@ -1,0 +1,2711 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!29 &1
+SceneSettings:
+  m_ObjectHideFlags: 0
+  m_PVSData: 
+  m_PVSObjectsArray: []
+  m_PVSPortalsArray: []
+  m_OcclusionBakeSettings:
+    smallestOccluder: 5
+    smallestHole: 0.25
+    backfaceThreshold: 100
+--- !u!104 &2
+RenderSettings:
+  m_ObjectHideFlags: 0
+  serializedVersion: 7
+  m_Fog: 0
+  m_FogColor: {r: 0.5, g: 0.5, b: 0.5, a: 1}
+  m_FogMode: 3
+  m_FogDensity: 0.01
+  m_LinearFogStart: 0
+  m_LinearFogEnd: 300
+  m_AmbientSkyColor: {r: 0.212, g: 0.227, b: 0.259, a: 1}
+  m_AmbientEquatorColor: {r: 0.114, g: 0.125, b: 0.133, a: 1}
+  m_AmbientGroundColor: {r: 0.047, g: 0.043, b: 0.035, a: 1}
+  m_AmbientIntensity: 1
+  m_AmbientMode: 0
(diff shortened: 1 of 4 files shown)


thestonefox:
You've been beaten to the finish line, the other PR with example scene 40 is now merged.

You'll have to go to 41 :(

ajaxlex:
Ha! Ok - will pick it up when I get home.

ajaxlex:
Hopefully I will not miss the boat again - modified to 41 in comments and scene number

ajaxlex:
!!  I have some repairs to apply to this ( and which I am putting through their paces in the example scene ) - you should delay pulling until I can get them integrated ( probably wed the 12th )

thestonefox:
Let me know when this is ready to review

ajaxlex:
Jump is fairly reliable, and somewhat simplified ( i think there is plenty to do there ), but there is a nagging issue that remains.  While jumping from column to column in the demo,  I keep teleporting to the ground, I suspect because of code in HeightAdjustTeleport ( Update method ).  I can add code to jump and to heightadj. teleport to listen for jumping, and add that to the growing list of things in IsExternalSystemManipulatingPlaySpace(), but this feels awkward.  I am wondering whether playerPresence has a part to play in negotiating among all these competing behaviors.  I have been trying to think of a nice pattern that would allow for extensibility ( I am contemplating a glide behavior, for example ) without cluttering up the height adj or player presence with a bunch of listeners.

External perspectives are welcome!

ajaxlex:
Maybe an interface IModifiesPlayspace which expects ChangingPlayspace and DoneChangingPlayspace?

xEvoGx:
**HeightAdjust** doesn't prevent you from climbing and then launching yourself into the air (this is how I "jump" from rope to rope in the demo), or simply falling off of high points (you fall, not snap immediately to the ground below), so clearly this is doable within the existing framework?

ajaxlex:
Ok, issue is def. related to HeightAdjust - it goes away when that is not configured.  I will run it down.

PR #629, opened 2016-10-06
feat(Object): show snap indicators for interactable objects - resolves #472

### LINKED ISSUE #472: Easier way of positioning snap handles
I have virtual hands in place of my controllers models, and they can take on a variety of poses.

I want to be able to precisely position the snap handles on an object so that it appears in the correct location (e.g. a key would be between the fingertips of a hand in a "pinch" pose).

Accurately positioning the snap handles to put the object in the right place is kind of tedious right now, involving a lot of trial and error, and of course it has to be done for each object and for each hand.

It would be great if there were some sort of tool (or tutorial) to simplify the process.

ajaxlex:
Allows the user to turn on and off snap handle indicators

### DIFF
--- a/Assets/VRTK/Editor/VRTK_InteractableObjectEditor.cs
+++ b/Assets/VRTK/Editor/VRTK_InteractableObjectEditor.cs
@@ -100,6 +100,7 @@
                     targ.precisionSnap = EditorGUILayout.Toggle(VRTK_EditorUtilities.BuildGUIContent<VRTK_InteractableObject>("precisionSnap", "Precision Grab(Snap)"), targ.precisionSnap);
                     if (!targ.precisionSnap)
                     {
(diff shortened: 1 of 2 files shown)


spaggi:
This is great, but it is not clear to me how this can be used. I can see the Boxes around my handle, but how exactly can they help me to align my handle? See attached pic

[https://i.imgur.com/qo1Tjzy.png](url)

ajaxlex:
Thanks spaggi.  The boxes represent the wrist, the palm area, and a finger, but I think they are deforming in rotation.  I will see if I can get a handle on the controller mesh - a little tricky for me, but easier to understand

thestonefox:
I think we should close this PR as using the scene gizmos doesn't work very well as they don't update position

Issue #642, opened 2016-10-11
Something strange in VRTK_BasicTeleport_UnityEvents.OnDisable()


gpvigano:
I found, by chance, that in VRTK_BasicTeleport_UnityEvents.OnDisable() event handlers are added again (+=) instead of being removed (-=), is this a bug?
In case this can help, I also checked every OnDisable() method in VRTK and none of them include a "+=".

thestonefox:
What line are they getting added in again?

thestonefox:
Oh in the UnityEvents Helper for it. Yeah that will be a bug they should be `-=`

gpvigano:
OK, anyway I checked all the other OnDisable() in VRTK and they all have `-=`

gpvigano:
I can't see any commit after this issue, so I made a PR ( #644 ).

PR #644, opened 2016-10-12
fix(UnityEvents): resolves issue #642

gpvigano:
Changed `+=` to `-=` in `VRTK_BasicTeleport_UnityEvents.OnDisable()`.

Event handlers were added again instead of being removed #642 .


thestonefox:
Thanks, I'll merge this later

Issue #494, opened 2016-09-01
VRTK_Simulator directions don't match the current camera orientation


mattboy64:
With no rotation on the playspace left/right rotate left/right were reversed. Rotating the playspace 180 degrees (before hitting play) fixed that problem, but the forward/backwards movement was then reversed. This is probably a problem with the movement directions not taking camera orientation into consideration when applying the movement offset.

I attached it to the CameraRig and the Camera(head). Both seemed wrong. It's also possible that I just don't know what I'm doing. :smile:

thestonefox:
Fixed with https://github.com/thestonefox/SteamVR_Unity_Toolkit/pull/495

maybe?

mattboy64:
I don't think so, he fixes the left/right thing. But there's something still inherently wrong with the rotation. Rotate the camera 180 degrees and I think the forward/backward becomes reversed.

thestonefox:
re-opening then :)

ridoutb:
I wonder if this is related to the Box Collider that gets added as a child of the Camera Rig by PlayerPresence. I noticed it does not follow the rotation of the headset or eye camera (it seems it should follow Y, at least). I was adding some momentum to the player and depending on the initial orientation of the rig, sometimes the force was incorrect by 90, 180 or 270 degrees (based on the initial Rig rotation).

sblancodiez:
Perhaps not totally related, but wouldn't it be better to move and rotate the whole CameraRig (basically the "physical" room space) with the simulator instead of moving the camera? The best solution would be perhaps to map the camera direction as forward first so controlling the rig could be like an FPS controller.

Or is there an specific reason for working with the camera?

mattboy64:
@sblancodiez I think the point of the script is to simulate headset movement without having to pick up the HMD and physically put it on your head. If that's accurate then it would make sense to only do it on the camera.

sblancodiez:
@mattboy64 I get that, but since most of the interaction is done with the
controllers and they have a position relative to the play area, it makes
some sense to move the CameraRig (since with this we still move around and
see in the pc screen what we are doing). This way you can use the
controllers to test things without putting on the headset.

Still, since the controllers have their position tracked inside the play
area, some means to have the headset camera stay in a given position inside
the play area would make testing faster. I.E. having it "behind" your chair
pointing towards the screen so you can just grab the controllers and watch
the screen to test.

mattboy64:
[...]

PR #650, opened 2016-10-14
fix(Presence): adjust physics fall to correct for local rotation

ajaxlex:
Fixes bug with climb where ungrab sends you in the wrong direction


ajaxlex:
See also Bug #494 
hattip -> MattBoy

PR #660, opened 2016-10-16
feat(Interaction): support moving knobs

10FingerArmy:
Knobs have been converted to use configurable joints. This enables
specifying connected rigidbodies like in many other 3d controls already.
Scene 25 contains a new knob on the small door now showcasing this.

While at it also the axis has been switch from world to local so
existing knobs will have to be adapted.

Also the value calculation was fixed which was working more by accident
than by intention. It now uses a heuristics to determine the correct
axis.

### DIFF
--- a/Assets/VRTK/Scripts/Controls/3D/VRTK_Knob.cs
+++ b/Assets/VRTK/Scripts/Controls/3D/VRTK_Knob.cs
@@ -19,6 +19,8 @@ namespace VRTK
             x, y, z // TODO: autodetect not yet done, it's a bit more difficult to get it right
         }
 
+        [Tooltip("An optional game object to which the knob will be connected. If the game object moves the knob will follow along.")]
+        public GameObject connectedTo;
         [Tooltip("The axis on which the knob should rotate. All other axis will be frozen.")]
         public KnobDirection direction = KnobDirection.x;
         [Tooltip("The minimum value of the knob.")]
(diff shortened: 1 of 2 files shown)


Assets/VRTK/Scripts/Controls/3D/VRTK_Knob.cs
```diff
@@ -163,8 +196,23 @@ private KnobDirection DetectDirection()
 
         private float CalculateValue()
         {
+            if (!subDirectionFound)
```
pvh: should be able to just use subDirection and drop subDirectionFound, right?

Assets/VRTK/Scripts/Controls/3D/VRTK_Knob.cs
```diff
@@ -163,8 +196,23 @@ private KnobDirection DetectDirection()
...
+            {
+                float angleX = Mathf.Abs(transform.localRotation.eulerAngles.x - initialLocalRotation.x) % 90;
+                float angleY = Mathf.Abs(transform.localRotation.eulerAngles.y - initialLocalRotation.y) % 90;
+                float angleZ = Mathf.Abs(transform.localRotation.eulerAngles.z - initialLocalRotation.z) % 90;
+                angleX = (Mathf.RoundToInt(angleX) > 89 && Mathf.RoundToInt(angleX) < 91) ? 0 : angleX;
```
pvh: What does this do? You're already % 90 above so shouldn't it always be 0 <= angle <= 90?

Assets/VRTK/Scripts/Controls/3D/VRTK_Knob.cs
```diff
@@ -163,8 +196,23 @@ private KnobDirection DetectDirection()
...
+                    subDirectionFound = true;
+                }
+            }
+
             float angle = 0;
-            switch (finalDirection)
+            switch (subDirection)
             {
                 case KnobDirection.x:
                     angle = transform.localRotation.eulerAngles.x - initialLocalRotation.x;
```
pvh: A few lines below (GitHub won't let me comment outside the patch) you round * 1000 then divide back by 1000. Math.Round has a two argument form where you can provide a precision. Math.Round(angle, 3), for example.

Assets/VRTK/Scripts/Controls/3D/VRTK_Knob.cs
10FingerArmy: Good catch I'll try to simplify that

Assets/VRTK/Scripts/Controls/3D/VRTK_Knob.cs
10FingerArmy: Math.Round only supports doubles, Unity works with floats, that's why there is the Mathf class, which unfortunately does not have the necessary rounding to digits functionality. Would you judge it's better to cross-cast to use the other method?

10FingerArmy:
Should be ready now for merge

Assets/VRTK/Scripts/Controls/3D/VRTK_Knob.cs
pvh: Nah, this is janky but fine as long as it's always done the same.

Assets/VRTK/Scripts/Controls/3D/VRTK_Knob.cs
pvh: Gotcha - C# is not my main language.

PR #661, opened 2016-10-16
feat(Headset): remove need for scripts on headset

thestonefox:
> **Breaking Changes**

Previously, the scripts that required to be on the Headset object would
require placing on the correct object that the headset would finally
become. As SteamVR places the headset on a different game object in
Unity 5.3 to Unity 5.4 and above, it meant that the headset scripts
were cumbersome to use.

This change now just expects any headset related script to be placed
on the `[CameraRig]` prefab as it will look up the headset object
from the DeviceFinder and then apply any logic to the headset
game object rather than needing to directly be on that object.

This change makes it much easier to apply headset scripts as there is
no need to know which object to apply the script to.

The example scenes have been updated to use this new scheme and the
previous `Add_HeadsetCollisionFade` script that was present in some
example scenes has been removed as it no longer requires to do any
clever logic within the editor.


Assets/VRTK/Examples/039_CameraRig_AdaptiveQuality.unity
```diff
@@ -37,7 +37,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
   m_CustomReflection: {fileID: 0}
   m_Sun: {fileID: 0}
-  m_IndirectSpecularColor: {r: 0.12758532, g: 0.13438144, b: 0.121170156, a: 1}
+  m_IndirectSpecularColor: {r: 0.12758517, g: 0.1343812, b: 0.121169955, a: 1}
```
bddckr: Looks like Unity wants to introduce some floating point changes...

PR #662, opened 2016-10-16
feat(Interaction): add option to interact without grab

10FingerArmy:
Controls now provide a native option to be interactable without the
need to hold the grab button. This will automatically setup the
required trigger volume based on the bounding box of the control.


thestonefox:
Works very nicely.

I think if the new way of dealing with rigidbodies on the controller (by tracking a separate object) will make this even better because at the moment if you push your controller through something then the force is applied as the rigidbody passes through it which is spangy.

PR #664, opened 2016-10-17
feat(Locomotion): add locomotion based on arm swing / head movement - resolves #602

Highsight:
This PR is to add in the Immersive Movement system to the VRTK.  Immersive Movement has been renamed Real Movement to fall more in line with the VRTK brand and to feel more generic and self-explanitory.

PR #667, opened 2016-10-17
feat(Locomotion): translate play area jump into game jump - resolves #393

### LINKED ISSUE #393: Jumping with Headset Movement

ajaxlex:
Adds the player jump script, and demo scene 041 to exercise it.

### DIFF
--- a/Assets/VRTK/Examples/041_CameraRig_PlayerJump.unity
+++ b/Assets/VRTK/Examples/041_CameraRig_PlayerJump.unity
@@ -1,0 +1,2704 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!29 &1
+SceneSettings:
+  m_ObjectHideFlags: 0
+  m_PVSData: 
+  m_PVSObjectsArray: []
+  m_PVSPortalsArray: []
+  m_OcclusionBakeSettings:
+    smallestOccluder: 5
+    smallestHole: 0.25
+    backfaceThreshold: 100
+--- !u!104 &2
+RenderSettings:
+  m_ObjectHideFlags: 0
+  serializedVersion: 7
+  m_Fog: 0
+  m_FogColor: {r: 0.5, g: 0.5, b: 0.5, a: 1}
+  m_FogMode: 3
+  m_FogDensity: 0.01
+  m_LinearFogStart: 0
+  m_LinearFogEnd: 300
+  m_AmbientSkyColor: {r: 0.212, g: 0.227, b: 0.259, a: 1}
+  m_AmbientEquatorColor: {r: 0.114, g: 0.125, b: 0.133, a: 1}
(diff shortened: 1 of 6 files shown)


xEvoGx:
Hey alex, here's my snip from Slack:

The only way to go straight up and not forward is to reduce _Forward_ to 0. Easy enough perhaps through a separate script to modify Forward at run-time so you can do either or, but perhaps it's a good idea to include it in your script (such as holding a button on the vive controller to jump up and forward, or even placing the controller in front of you.

Ultimately it would be nice to have locomotion influence jumping up or up/forward/left/right/back. Now that we do in fact have an _arm swinger_ addition to VRTK #664 , that too might not be a bad idea to integrate.

Cheers brother

ajaxlex:
( from slack ) Those are good ideas.  I played Redswood recently, and they use a mechanic where a button is pressed and the controller is swung through the air to cause a jump.  I prefer the visceral association of a real jump motion though.  I had considered incorporating controller orientation into the mix ( amount controllers are moved backward during jump translates into forward motion )

xEvoGx:
Whatever happened to this as it doesn't appear to be a part of the Release?

ajaxlex:
Hey xEvoGx - there was some instability in the jump code that I think is finally fixed.  i have been concentrating on getting my product out the door, then will clean up the more mature code and submit a new pull request.  

The new code does some things like:

- better ground detection 
- dynamic player height detection 
- a public function to force a jump ( use with care! ) 
- a jump inhibitor 
- a lateral offset test that discourages jumps when a player is just picking something up off the ground

Its also considerably more broadly user tested, though it would be good to see it in other application contexts.

xEvoGx:
Awesome, no idea it was even still being worked on!

thestonefox:

PR #669, opened 2016-10-18
fix(Interaction): use relative instead of world force when snapping - resolves #518

### LINKED ISSUE #518: VRTK_Door Snapping option seem to be orientation dependant
-Open scene "025_Controls_Overview". Move "NormalDoor" where you want on the floor.
-Press play. It's OK.
-Stop.
-set rotation of "NormalDoor" to 0,-120,0.
-Press play. The door is opening without user interaction.
door orientation 0 120 0

-Stop. Set VRTK_Door Snapping to false.
-Play. The door is working normally again.

I don't find a simple way to fix this. Same issue with another door 3d model.

10FingerArmy:
Fixes issue of force affecting objects after rotation

### DIFF
--- a/Assets/VRTK/Scripts/Controls/3D/VRTK_Door.cs
+++ b/Assets/VRTK/Scripts/Controls/3D/VRTK_Door.cs
@@ -286,7 +286,7 @@ namespace VRTK
             }
             if (doorCfCreated)
             {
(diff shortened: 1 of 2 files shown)


10FingerArmy:
Fixes https://github.com/thestonefox/VRTK/issues/518

PR #678, opened 2016-10-21
feat(SDK): GearVR support sample implementation

gpvigano:
Derive a class `SDK_Default` from `SDK_Base`, overriding each abstract
method (mostly with an empty body). As default implementation get the
main camera as headset.

Derive a class `SDK_GearVR` from `SDK_Default`, overriding only specific
methods (not yet implemented).

Automatically create the correct SDK in `VRTK_SDK_Bridge` (if Unity 5.4
or above is available, checking `VRSettings.loadedDeviceName`).

Derived a class `SDK_GearVR_Events` from `VRTK_ControllerEvents`, with a
new `Update()` method, and other members related to the GearVR.

Because VRTK is currently bound to `VRTK_ControllerEvents` for now
`SDK_GearVR_Events` derives from it and to defines the GearVR specific
behaviour, cancelling in this way the base class behaviour.

Add `SDK_DeviceSelector` MonoBehaviour to switch between camera rigs for
the different supported devices.

Add an example working both with both Vive and GearVR
(`042_SDK_MultiDeviceSupport`).

### DIFF
--- a/Assets/VRTK/Examples/042_SDK_MultiDeviceSupport.unity
+++ b/Assets/VRTK/Examples/042_SDK_MultiDeviceSupport.unity
@@ -1,0 +1,4032 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!29 &1
+SceneSettings:
+  m_ObjectHideFlags: 0
+  m_PVSData: 
+  m_PVSObjectsArray: []
+  m_PVSPortalsArray: []
+  m_OcclusionBakeSettings:
+    smallestOccluder: 5
+    smallestHole: 0.25
+    backfaceThreshold: 100
+--- !u!104 &2
+RenderSettings:
+  m_ObjectHideFlags: 0
+  serializedVersion: 7
+  m_Fog: 0
+  m_FogColor: {r: 0.5, g: 0.5, b: 0.5, a: 1}
+  m_FogMode: 3
+  m_FogDensity: 0.01
+  m_LinearFogStart: 0
+  m_LinearFogEnd: 300
+  m_AmbientSkyColor: {r: 0.212, g: 0.227, b: 0.259, a: 1}
+  m_AmbientEquatorColor: {r: 0.114, g: 0.125, b: 0.133, a: 1}
+  m_AmbientGroundColor: {r: 0.047, g: 0.043, b: 0.035, a: 1}
+  m_AmbientIntensity: 1
+  m_AmbientMode: 0
+  m_SkyboxMaterial: {fileID: 10304, guid: 0000000000000000f000000000000000, type: 0}
+  m_HaloStrength: 0.5
+  m_FlareStrength: 1
+  m_FlareFadeSpeed: 3
+  m_HaloTexture: {fileID: 0}
+  m_SpotCookie: {fileID: 10001, guid: 0000000000000000e000000000000000, type: 0}
+  m_DefaultReflectionMode: 0
+  m_DefaultReflectionResolution: 128
+  m_ReflectionBounces: 1
+  m_ReflectionIntensity: 1
+  m_CustomReflection: {fileID: 0}
+  m_Sun: {fileID: 0}
+  m_IndirectSpecularColor: {r: 0.44692492, g: 0.4967869, b: 0.57508546, a: 1}
+--- !u!157 &3
+LightmapSettings:
+  m_ObjectHideFlags: 0
+  serializedVersion: 7
+  m_GIWorkflowMode: 0
+  m_GISettings:
+    serializedVersion: 2
+    m_BounceScale: 1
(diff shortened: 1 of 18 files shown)


thestonefox:
Quite a few good ideas here, definitely food for thought.

Quite like the SDK_Default class that just returns defaults for everything that you then override.

In fact there is a SDK_Base class now which is an abstract class, this could be updated to do the same thing as what you're doing with the defaults.

I'm not sure if the SDK chooser would work though, for instance if you are building and don't have the SteamVR plug in installed (or the Oculus SDK plugin) then you're going to get a bunch of errors in scripts.

My solution to this (while I was messing with the Oculus SDK Bridge) was to wrap things like

```
#if SDK_SteamVR
   SDK_SteamVR Bridge code here
#endif
```

Then use Unity `Scripting Define Symbols` in the Player Settings to denote which SDK you are currently building for. This has the bonus that all of that code gets ignored unless you define that value.

Quite like the idea of extending the Controller_Events script too.

I definitely think it's a good idea to split up the SDK Bridge into files that are relevant like Headset, Controller, Play Area.

I'll do some of this work and raise a PR then post here for further discussion.

gpvigano:
Thank you for your feedback. Just one remark about the SDK chooser.

I understand the need to build for a device without having it connected, but it is also uncomfortable to continuosly change the Player Settings when you are working with more devices in the same project.
I am testing a multi-user environment with the GearVR connected via wi-fi to my PC. To avoid continuous (time consuming) platform switching, I usually set Android as target platform, running a client on the GearVR and running  a server in the Unity Player on the Windows PC.

In this way I can just build the Android APK, then run the server side on the player without any change, with or without the Vive connected. If the Vive is not connected I can use keyboard and mouse to emulate the GearVR SDK (this is done by `SDK_DeviceSelector` script, setting a defalut camera rig).

thestonefox:
You also have to take into consideration building without even owning that device.

E.g. if there is a bunch of GearVR specific calls or Oculus specific calls and someone doesn't own that headset then it will constantly fail until you install their SDK, which isn't a good solution imho

gpvigano:
I didn't install anything, with Unity 5.4 the GearVR support is native with Unity, I am using a Unity standard camera.
I am not using neither Oculus SDK or VR Samples, my goal is to develop everything upon VRTK (for now I created auxiliary scripts and prefabs externally to VRTK).
The problem is with the older versions, but in that case the code can be excluded if `UNITY_5_4_OR_NEWER` is not defined.
I think we better exploit the Unity 5.4 device management instead of reinvent the wheel.
I know that VRTK must work also with 5.3, but hopefully things can be arranged to let 5.4 users exploit new features without breaking the code for 5.3 users.

gpvigano:
To be more precise: I didn't install anything on my Windows PC to build for GearVR, but obviously I installed the Oculus GearVR runtime on the Android smartphone.

thestonefox:
Yeah for your specific use case it probably works. But in the greater scheme of things when it comes to the toolkit being SDK agnostic, using unity defines means you don't need to have other SDKs installed to develop.

I'm talking from a purely what's more appropriate for the toolkit perspective rather than what works for you perspective.

gpvigano:
As I wrote, for Oculus and OpenVR, since Unity 5.4, there is no need to have an SDK installed (I also hope that the Unity native support will be extended for other devices in future).
I think that it is anyway useful to have a default configuration that runs (possibly without errors) when nothing is connected.

thestonefox:
You need to use an SDK for both Oculus and SteamVR for unity 5.4 and 5.5 as Unity 5.4 only supports native headset detection it does not support controllers (Unity 5.5 is bringing basic controller support though but as far as I can tell it's not full controller support).

gpvigano:
You're right, it works for me because I just use GearVR controllers, not for Oculus Touch or HTC Vive (and we still depend on SteamVR SDK). Let's postpone this for a possible future release of Unity...

I try to point out some idea about a possible implementation...

About the controller events: maybe a base class with virtual methods (button aliases and events for examples), to be overridden in more specific classes written for each device, paying attention to expose (as public or protected) those members that can be useful for the derived classes. I would not use an external abstract SDK class, whose methods are called by controllers (like the current implementation).

A default SDK for testing in Unity Player when no device is connected could be useful (I am using the main camera and providing a controller that can be both attached to the camera or moved with keyboard and mouse).

thestonefox:
Having a default Unity SDK bridge would be the ideal solution. If Unity do continue to implement a standard SDK that devices can support will be excellent.

I think splitting the SDK bridges into the separate sections will mean a Unity Headset SDK Bridge will be possible with 5.4.

As for the controller stuff, yeah you're right, it's very `Vive` centric at the moment, it needs to migrate away from that so it's more generic.

thestonefox:
Potentially superseded by https://github.com/thestonefox/VRTK/pull/944

PR #680, opened 2016-10-21
feat(UIPointer): add ability to choose click on button down or up - resolves #525 - resolves #639 - resolves #675

### LINKED ISSUE #525: Change VRTK_UIPointer click handlers to Down instead of Release
I've run into an interesting issue and a possible inconsistency. InteractableObject events fire on down/click, while GUI events fire on up/release. 

Background:  In my app, an Always_On pointer can be aimed at a gameObject.  Upon trigger press, that gameobject displays a popup GUI over it.  Then, the pointer can be used to interact with GUI buttons.  But what's happening is that on trigger _press_, the GameObject (correctly) shows the GUI, but then on trigger _release_, whichever GUI button happens to be under the popup menu is inadvertently fired.  Within a sub-second press/release of the trigger, two actions occur, when ideally only one should.

More generally, how should users interact with elements? In the desktop/mouse environment, mouseUp is typically when actions execute, and users are familiar with that convention.  But in VR... does it seem more appropriate to perform actions on trigger _click_, _grip_, etc? That would more closely mirror real life.  It would also make GUI interaction consistent with InteractableObject interaction.  (Consider this: If I had modeled buttons as meshes / InteractableObjects, they would have fired on press. But because I'm using GUI buttons, they fire on release. Devs know the difference... but users don't, and when they're standing in a VR environment with solid-looking objects at waist height, the distinction feels arbitrary.)

Anyway, just wondered if this issue is worth considering? 

As an aside, I'm making the transition over to VRTK from NewtonVR, where I never had this issue (but ran into several others, heh).  In NVRCanvasInput, I see ExecuteEvents.ExecuteHeirarchy (which I don't pretend to understand), with the following explanation:
```
// we want to do click on button down at same time, unlike regular mouse processing
// which does click when mouse goes up over same object it went down on
// reason to do this is head tracking might be jittery and this makes it easier to click buttons
```

As always, thanks for the great toolkit!

### LINKED ISSUE #639: Allow Controller to directly interact with Unity UI elements
Allow users to touch, press, and interact with GUI elements (such as buttons) via controller meshes... as an alternative to laser pointer.

Background info: 
As discussed on Slack, the emerging trend seems to be that if elements are within arms length, the user ought to be able to poke at and otherwise interact with them using their finger/hand (or similar custom controller mesh), rather than a twitchy laser pointer.  (I've seen laser pointer cause confusion in my own usability testing.)  Also, it was mentioned how hover events perhaps ought to be triggered when the controller is within a couple of centimeters, and @thestonefox suggested that a spherecast or capsulecast be used around the controller.

### LINKED ISSUE #675: Implement Drag & Drop mechanic for Canvas UI elements.
Provision of drag & drop functionality with Canvas UI elements to allow:

- Selection of a GUI element (e.g. an image) with the controller by touching the element (to highlight it).
- Holding the trigger could then select the element and allow it to be dragged with the controller.
- Releasing the trigger would drop the element at that position in the canvas.
- Snap functionality would then re-position the element to the nearest snap point (e.g. another GUI element).

The above could perhaps be an extension to the functionality between the controller and GUI elements discussed in issue #639.

thestonefox:
The UI Pointer click action can now be chosen per UI Pointer script.

Previously, the UI Click action would fire when the UI Click alias
button was pressed and the released (on click up) however it is also
a good idea to provide an option to fire the click action when the
UI Click alias button is pressed down.

The On Click Up works like a mouse click, if a mouse button is pressed
then the action on a clicked button is only triggered when the mouse
button is released.

However, the On Click Down works like a keyboard, if a key is pressed
then the action to type is done as the key is pressed down not when
it is released.

### DIFF
--- a/Assets/VRTK/Scripts/Helper/VRTK_EventSystemVRInput.cs
+++ b/Assets/VRTK/Scripts/Helper/VRTK_EventSystemVRInput.cs
@@ -25,6 +25,7 @@
                     {
                         results = CheckRaycasts(pointer);
                     }
+
                     //Process events
                     Hover(pointer, results);
                     Click(pointer, results);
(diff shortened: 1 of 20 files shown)


Assets/VRTK/Scripts/Helper/VRTK_EventSystemVRInput.cs
```diff
@@ -153,13 +154,70 @@ private void Hover(VRTK_UIPointer pointer, List<RaycastResult> results)
 
         private void Click(VRTK_UIPointer pointer, List<RaycastResult> results)
         {
-            pointer.pointerEventData.eligibleForClick = pointer.controller.uiClickPressed;
+            switch (pointer.clickMethod)
+            {
+                case VRTK_UIPointer.ClickMethods.Click_On_Button_Up:
+                    ClickOnUp(pointer, results);
```
thestonefox: is this needed if this is implemented:

https://github.com/thestonefox/VRTK/issues/685

thestonefox: #685 is a no-go, so yes

PR #673, opened 2016-10-20
feat(Interaction): change velocity of thrown object

TorrqTB:
…ce from controller

Based on article at http://www.normalvr.com/blog/throwing-throw-down/

Currently when throwing grabbed object, the velocity is based on center of steam controller.
Change to increase velocity based on distance object is attached to controller.
Example: Ball at end of stick will move at greater speed, compared to where stick is being held.

PR #684, opened 2016-10-22
feat(Interaction): alter velocity of thrown object based on distance

TorrqTB:
Replace PR #673

Currently when throwing grabbed object, the velocity is based on center
of steam controller. Change to increase velocity based on distance
object is attached to controller.

Example: Ball at end of stick will move at greater speed,
compared to where stick is being held.

Based on article at http://www.normalvr.com/blog/throwing-throw-down/


TorrqTB:
Thanks much!

I also have code that calculates on the RigidBody AttachPoint.  And another set that reproduces the articles implementation if anyone is interested.

The one in the branch is the simplest solution though.

PR #699, opened 2016-10-28
fix(Climbing): change VRTK_PlayerClimb to allow moving objects.

### LINKED ISSUE #697: Climbing does not support moving InteractableObjects
When you grab a moving Interactable Object with the climb interaction it does not move the player alongside the IO. As a player I expect to move alongside the the IO as if I was attached to it.

JoshGalvin:
Allow modeling of Zip Lines, Helicopters, etc...
Fix #697.

### DIFF
--- a/Assets/VRTK/Scripts/VRTK_PlayerClimb.cs
+++ b/Assets/VRTK/Scripts/VRTK_PlayerClimb.cs
@@ -48,6 +48,7 @@ namespace VRTK
         private Transform headCamera;
         private Transform controllerTransform;
         private Vector3 startControllerPosition;
+        private Vector3 startClimbingObjectPosition;
         private Vector3 startPosition;
         private Vector3 lastGoodHeadsetPosition;
         private bool headsetColliding = false;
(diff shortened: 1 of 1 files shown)


mattboy64:
Did you test the safe zone teleport with this? 

Also could you include an option for rotation?

JoshGalvin:
I haven't tested out the safe zone teleport. I'll do that and add rotation at the same time.

lucboudreau:
Hey guys, I have some code here which I think works better. I've also added support for large rotating objects with precision snap. I've tested the shit out of it and it works awesome. Please consider this code as I find it invaluable to be able to precision snap without needing to add thousands of colliders.

See my comments on the issue.

https://github.com/thestonefox/VRTK/issues/716

mattboy64:
I created a version with movement but also added rotation for the 3.0 version of VRTK here: #903

It would be great to merge all of these climbingMove PRs together and get these finally committed, so please take a look and tell me what you think.

AlbertMN:
Greatly appreciated :)

thestonefox:
Superseded by https://github.com/thestonefox/VRTK/pull/903

PR #718, opened 2016-11-08
fix(DnapDropZone): fix `ValidSnapObject()` not searching parents for `InteractableObject`

quintesse:
Right now the snap to drop zone functionality only works if the `Collider`
generating the event is on the same object that has the `InteractableObject`
script. If your Colliders are only on the children things won't work.
This fix makes sure that the `ValidSnapObject()` method also looks in the
parent objects.

### DIFF
--- a/Assets/VRTK/Prefabs/Resources/Scripts/VRTK_SnapDropZone.cs
+++ b/Assets/VRTK/Prefabs/Resources/Scripts/VRTK_SnapDropZone.cs
@@ -281,6 +281,9 @@ namespace VRTK
         private VRTK_InteractableObject ValidSnapObject(GameObject checkObject, bool grabState)
         {
             var ioCheck = checkObject.GetComponent<VRTK_InteractableObject>();
+            if (ioCheck == null) {
+                ioCheck = checkObject.GetComponentInParent<VRTK_InteractableObject>();
+            }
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 8b8441d99984dc1ef88f1f852c2d90a3b48998ba

---

quintesse:
Sorry about the line length issues, but the contributing docs specifically say "The commit message lines should never exceed 72 **lines**", it didn't say "72 characters in length" :)
(I was already thinking: who on earth writes commit messages *that* long?)
Anyway, very later here already, if it's really a problem I'll do a new PR tomorrow.

thestonefox:
Superseded by https://github.com/thestonefox/VRTK/pull/722

Also, there were more occurrences of GetComponent that needed updating, plus doing GetComponentInParent is the same as doing GetComponent, it checks the current object first the goes up the hierarchy looking for the next valid component.

quintesse:
> plus doing GetComponentInParent is the same as doing GetComponent

Ah, you're absolutely right, but I copied that way of working from `VRTK_InteractTouch.IsObjectInteractable` so there might be more code around that does that (unnecessarily).

thestonefox:
Thanks for the heads up, I've fixed those now in Interact Touch

quintesse:
I will :)

Btw, is there a mailing list or chat for VRTK somewhere? I couldn't find that information.

PR #731, opened 2016-11-18
fix(Throwing): calc throw velocity from center of collision

pargee:
Throwing objects with a non-centered pivot would cause undesired results for objects like spheres with their pivot at the bottom. Suggesting using the center of the collider instead.

This is a proposed fix from #684


thestonefox:
I've actually made that additional point calculation optional in an upcoming build due to the issues it can cause.

thestonefox:
Resolved with https://github.com/thestonefox/VRTK/commit/6dbc7124b1c6728af3205005398b69af0010f978

PR #752, opened 2016-12-16
feat(menu): add setup helper

virror:
Add a new menu item to help with setting up a new Intractable Object.

### DIFF
--- a/Assets/VRTK/Editor/VRTK_ObjectSetup.cs
+++ b/Assets/VRTK/Editor/VRTK_ObjectSetup.cs
@@ -1,0 +1,172 @@
+﻿namespace VRTK
+{
+    using UnityEngine;
+    using UnityEditor;
+    using VRTK.GrabAttachMechanics;
+    using VRTK.SecondaryControllerGrabActions;
+
+    public class VRTK_ObjectSetup : EditorWindow
+    {
+        private enum PrimaryGrab
+        {
+            ChildOfController,
+            FixedJoint,
+            Climbable,
+            CustomJoint,
+            RotatorTrack,
+            SpringJoint,
+            TrackObject
+        }
+        private enum SecondaryGrab
+        {
+            SwapControllers,
+            ControlDirection,
+            AxisScale
+        }
+        private bool useGrab = true;
+        private bool holdGrab = false;
+        private bool useUse = false;
+        private bool useIfGrabbed = false;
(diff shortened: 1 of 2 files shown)


Assets/VRTK/Editor/VRTK_ObjectSetup.cs
```diff
@@ -0,0 +1,135 @@
...
+        [MenuItem("Window/VRTK/Setup Interactable Object")]
+        private static void Init()
+        {
+            VRTK_ObjectSetup window = (VRTK_ObjectSetup)EditorWindow.GetWindow(typeof(VRTK_ObjectSetup));
+            
+            window.minSize = new Vector2( 300f, 300f );
+            window.maxSize = new Vector2( 400f, 300f );
+            
+            window.autoRepaintOnSceneChange = true;
+            window.titleContent.text = "Setup Object";
```
thestonefox: worth naming it "Setup Interactable Object" to make it more obvious what the tab is?

Assets/VRTK/Editor/VRTK_ObjectSetup.cs
```diff
@@ -0,0 +1,135 @@
...
+        private bool holdGrab = false;
+        private bool useUse = false;
+        private bool useIfGrabbed = false;
+        private bool holdUse = false;
+        private PrimaryGrab primGrab;
+        private SecondaryGrab secGrab;
+        private bool disableIdle = true;
+        private bool addrb = true;
+
+        [MenuItem("Window/VRTK/Setup Interactable Object")]
```
thestonefox: I quite like the idea of having this as a main menu option

"VRTK/Setup Interactable Object"

thoughts?

virror:
I have some other small things i want to change, so i will move the menu to main instead.
 "Setup Interactable Object" dont fit the small Unity tabs, that's why i used a shorter name instead.

getwilde:
For this usage, "Set up" should be two words.  But grammer police are annoying, so you can just ignore me.   :)

Assets/VRTK/Editor/VRTK_ObjectSetup.cs
getwilde: For this usage, "Set up" should be two words. But grammer police are annoying, so you can just ignore me. :)

virror:
Ok, this is ready to be merged now.

PR #754, opened 2016-12-17
feat(Interaction): Added simple script for unify scaling as second gr…

smygarn:
Added a realy simple second grab action script to scale the current object, works just as expected.

### DIFF
--- a/Assets/VRTK/Examples/043_Controller_SecondaryControllerActions.unity
+++ b/Assets/VRTK/Examples/043_Controller_SecondaryControllerActions.unity
@@ -37,7 +37,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
   m_CustomReflection: {fileID: 0}
   m_Sun: {fileID: 0}
-  m_IndirectSpecularColor: {r: 0.37429237, g: 0.42147171, b: 0.46712714, a: 1}
+  m_IndirectSpecularColor: {r: 0.18053626, g: 0.22601849, b: 0.30718803, a: 1}
 --- !u!157 &3
 LightmapSettings:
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: 11069acbcdcaea95d6b3137bd4f1882de04824ea

---

thestonefox:
There is an option in the existing scale secondary grab script for uniform scaling.

https://github.com/thestonefox/VRTK/blob/master/Assets/VRTK/Scripts/Interactions/SecondaryControllerGrabActions/VRTK_AxisScaleGrabAction.cs#L23

smygarn:
I know, but this works more as you expect and is much easier implemented. Please try it out.

thestonefox:
Yes, but if you have improvements or additions to an existing script then do a pr on amending an existing script rather than have 2 scripts that have very similar functionality

smygarn:
Unfortunately I didn't get the original uniform scaling to scale as expected and the original script was too complicated to refactor.

thestonefox:
The uniform scale is only a couple of lines in the existing script. I'd recommend looking at it again and if there are improvements then raise an issue, ask for help in slack and look to improve the existing script.

It's better to have one good working script than 2 scripts where they do the same thing and one works and the other one doesn't work as expected.

smygarn:
I think it ends here as I can still use the script in my application as it extends the base action script.

thestonefox:
@smygarn take a look at this pull request: https://github.com/thestonefox/VRTK/pull/782

It attempts to implement your solution in the existing script for uniform scale.

So rather than using distance, it uses magnitude like your solution and provides a smoother uniform transition.

PR #788, opened 2017-01-05
fix(Interaction): ensure collision checks are done in physics loop

thestonefox:
Previously, there was a check to see if a collider on the controller
had been disabled by checking in the `LateUpdate` method which worked
but only because SteamVR locked the physics time step meaning the
physics routine (running the OnTrigger stuff) was in sync with the
Update/LateUpdate routines.

However, if the physics time step was not locked (e.g. if using the
simulator) then it would be possible to get the LateUpdate routine
to run during the physics loop and cause objects to constantly be
touched and untouched.

As FixedUpdate is a physics routine, it's best to do some of the
checks in there then the different time steps won't cause any
issues.


virror:
LGTM

PR #793, opened 2017-01-07
perf(ControllerEvents): don't create garbage on vector comparison

bddckr:
Two `Vector2`s were compared by using a string based method which
creates garbage on every call which decreases performance. The solution
is to do it without using strings and only relying on stack allocated
structs.
Thanks to @reznovvr for this performance fix.


thestonefox:
5KB down to 0B for GC ❤️ 💚 💛 💙

PR #796, opened 2017-01-08
feat(smoothing): add generic smoothing and pointer smoothing settings

bddckr:
Most humans aren't able to hold controllers completely still. To
improve the user experience it is therefore a good idea to smooth the
position and rotation of tracked objects like controllers.

This is enabled by improving on the existing `VRTK_ObjectFollow`. This
change adds two new subclasses to support smoothing any combination of
position, rotation and scale of a game object's transform or rigidbody.

One of the best candidates for smoothing are the various pointers in
VRTK. That's why this change also adds smoothing settings to the base
pointer so both the simple and bezier pointer benefit of a smoothing
option.

Thanks to Nathan Rowe from SculptrVR for the pseudocode!

# TODO

- [x] #795
- [x] Merge smoothing into `VRTK_ObjectFollow`.
- [x] Make sure all the example scenes still work after merging the classes.
- [x] Allow updating in `OnPreRender`.
- [x] Allow using a rigidbody instead of the transform.
- [x] For pointer smoothing: Figure out a way to only smooth the pointer's tip, not the origin.
- [x] Document everything.
- [x] Create new issue/PR  to replace `Follow` with the new merged `VRTK_ObjectFollow` (#839).

Easily done by using the new subclasses of `VRTK_ObjectFollow`:
- [x] ~~Add a script that allows to smooth controllers?~~
- [x] ~~Add a script that allows to smooth the headset?~~

Scrapped for now:
- [x] ~~Allow smoothing grabbed objects.~~
- [x] ~~Allow smoothing two-handed grabbed objects.~~

---

The following is fixed already:
> # Problems with current implementation
> 
> This introduced some serious lag to the pointer beam when moving the controller. It looks like if I switch from `Update` to `OnPreRender` I get the latest transform, so the lag is minimized.
> 
> `Update`:
> update
> `OnPreRender`:
> onprerender
> 
> I'm currently looking into providing some sort of "late transform updating" utility for VRTK. This utility will go through the SDK since SteamVR is probably doing their pose updating at another point in time than the Oculus SDK.
> The utility should then be used wherever VRTK creates game objects whose transform is updated instead of attaching directly to the appropriate tracked object.


xEvoGx:
Nice and could be great for two-handed weapons (the second hand would activate smoothing)

bddckr:
This introduced some serious lag to the pointer beam when moving the controller. It looks like if I switch from `Update` to `OnPreRender` I get the latest transform, so the lag is minimized.

`Update`:
update
`OnPreRender`:
onprerender

I'm currently looking into providing some sort of "late transform updating" utility for VRTK. This utility will go through the SDK since SteamVR is probably doing their pose updating at another point in time than the Oculus SDK.
The utility should then be used wherever VRTK creates game objects whose transform is updated instead of attaching directly to the appropriate tracked object.

---

# TODO for this PR:

- [ ] Document the `VRTK_TransformSmoothing` class itself.
- [ ] Add a comment marker as the first line to `VRTK_TransformSmoothing.cs`.
- [ ] Add a new example scene. (Or maybe not since the examples will be changed soon I think?! @thestonefox)
- [ ] Add and use said utility.

bddckr:
# Basic Pointer smoothing
Before:
before
After:
after

bddckr:
# Controller smoothing
Only the controller:
controlleroriginonly
Controller and pointer tip:
controlleroriginandpointertip

thestonefox:
You wanna see if you can implement this in the new pointer stuff now?

bddckr:
@thestonefox Done!
Should I remove the smoothing from the old, deprecated pointers? Given that the work is already done I think we can leave that in.

thestonefox:
@bddckr meh, just leave it in there.

PR #802, opened 2017-01-10
refactor(SDK): use OpenVR provided invalid index

bddckr:
The VRTK SteamVR SDK used `uint.MaxValue` to check if a controller
index is out of bounds. This change makes sure the OpenVR provided
constant for the first invalid tracked device index is used instead.
`uint.MaxValue` is still used for everything that is VRTK API since it
is used to signal an invalid or missing index.


bddckr:
This is a `refactor` rather than a `fix` because nothing is broken yet.

Issue #804, opened 2017-01-11
Improve documentation of enums


bddckr:
If we name enums `LikeThis` instead of `Like_This` they show up way better in the inspector:
`Like This` instead of `Like_This`.
To document enum cases the documentation should be done on the case itself. This probably means to update the doc generator for this. We get no documentation in IDEs otherwise. :disappointed:
The contributing document should probably be updated to name the new enum convention.

From
```C#
/// <summary>
/// Allowed controller type.
/// </summary>
/// <param name="Both">Both controllers are allowed to interact.</param>
/// <param name="Left_Only">Only the left controller is allowed to interact.</param>
/// <param name="Right_Only">Only the right controller is allowed to interact.</param>
public enum AllowedController
{
	Both,
	Left_Only,
	Right_Only
}
```

To
```C#
/// <summary>
/// Allowed controller type.
/// </summary>
public enum AllowedController
{
	/// <summary>
	/// Both controllers are allowed to interact.
	/// </summary>
	Both,
	/// <summary>
	/// Only the left controller is allowed to interact.
	/// </summary>
	LeftOnly,
	/// <summary>
	/// Only the right controller is allowed to interact.
	/// </summary>
	RightOnly
}
```

I'll take a stab at creating a PR for this soon. 😄

thestonefox:
PR with fix https://github.com/thestonefox/VRTK/pull/1465

PR #808, opened 2017-01-11
feat(Locomotion): add improved touchpad movement script

reznovVR:
Adds a new touchpad locomotion script that supports multiple configurable movement options for horizontal and vertical axis.

### DIFF
--- a/Assets/VRTK/Scripts/Locomotion/VRTK_TouchpadMovement.cs
+++ b/Assets/VRTK/Scripts/Locomotion/VRTK_TouchpadMovement.cs
@@ -1,0 +1,493 @@
+﻿// Touchpad Movement|Locomotion|20062
+namespace VRTK
+{
+    using UnityEngine;
+
+    /// <summary>
+    /// Adds the ability to move and rotate the play area and the player by using the touchpad. 
+    /// </summary>
+    /// <remarks>
+    /// The Touchpad Movement script requires VRTK_BodyPhysics script to be present in one of the scene GameObjects for collision detection. 
+    /// 
+    /// Vertical axis movement types include:
+    /// - regular smooth sliding (walking)
+    /// - warping, which instantly moves the player forward at a fixed distance
+    /// 
+    /// Horizontal axis movement types include:
+    /// - smooth sliding (strafing)
+    /// - smooth rotation
+    /// - snap rotation, which instantly rotates the player at a fixed angle
+    /// - warping, which instantly moves the player sideways (instant strafing)
+    /// 
+    /// Additionally it's possible to enable direction flip feature which allows the user to do an instant 180 degree turn by pressing the touchpad down.
+    /// 
+    /// It's also possible to define a button to multiply any type of movement (speed, range, angle) when the set button is pressed. Values above one will give a boost effect
+    /// and values below one will do the opposite. All movement values are public properties and can be set from other script at runtime.
+    /// 
+    /// Different movement types can be split across the controllers by having one script per hand side and with the desired options. 
+    /// Warp and snap rotate options may provide more comfortable experience for some and blink effect can be used to soften the movement. 
+    /// Snap rotate and flip direction options can be useful with teleport scripts for seated experiences and for people using front facing camera setups(Oculus default, PSVR). 
+    /// </remarks>
+    public class VRTK_TouchpadMovement : MonoBehaviour
+    {
+        /// <summary>
+        /// Movement types that can be performed by the vertical axis.
+        /// </summary>
+        /// <param name="None">No movement is performed.</param>
+        /// <param name="Slide">Performs smooth movement (walk).</param>
+        /// <param name="Warp">Performs an instant warp movement.</param>
+        /// <param name="WarpWithBlink">Performs an instant warp movement with a blink effect.</param>
+        public enum VerticalAxisMovement
+        {
+            None,
+            Slide,
+            Warp,
+            WarpWithBlink
+        }
+
+        /// <summary>
+        /// Movement types that can be performed by the horizontal axis.
+        /// </summary>
+        /// <param name="None">No movement is performed.</param>
+        /// <param name="Slide">Performs smooth movement (strafe).</param>
+        /// <param name="Rotate">Performs smooth rotation.</param>
+        /// <param name="SnapRotate">Performs fixed angle rotation.</param>
+        /// <param name="SnapRotateWithBlink">Performs fixed angle rotation with a blink effect.</param>
+        /// <param name="Warp">Performs an instant warp movement.</param>
+        /// <param name="WarpWithBlink">Performs an instant warp movement with a blink effect.</param>
+        public enum HorizontalAxisMovement
+        {
+            None,
+            Slide,
+            Rotate,
+            SnapRotate,
+            SnapRotateWithBlink,
+            Warp,
+            WarpWithBlink
+        }
+
+        /// <summary>
+        /// Which type axis movement did occur.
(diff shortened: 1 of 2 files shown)


bddckr:
I didn't test it but I just needed to underline how clean that code looks. Good work 👍

thestonefox:
I'll do a full review, but you'll also need a git description explaining the solution:

best thing to do is a `git commit --amend` and just add in the git body (if you look at some other commits in the repo, you'll see they have fairly lengthy commit messages explaining the commit)

When you've updated the commit just force push your branch back to github and it will auto update the pull request.

If you need any help with the gits then just hollar

Assets/VRTK/Scripts/Locomotion/VRTK_TouchpadMovement.cs
```diff
@@ -0,0 +1,480 @@
...
+    /// Snap rotate and flip direction options can be useful with teleport scripts for seated experiences and for people using front facing camera setups (Oculus default, PSVR). 
+    /// </remarks>
+    public class VRTK_TouchpadMovement : MonoBehaviour
+    {
+        /// <summary>
+        /// Movement types that can be performed by the vertical axis.
+        /// </summary>
+        public enum VerticalAxisMovement
+        {
+            /// <summary>
```
thestonefox: Whilst this is the "correct" way of doing the enum docs. The current doc generator doesn't work like this.

There is a github issue to update all enums this way -> https://github.com/thestonefox/VRTK/issues/804

But this means we can either wait to merge this pull request until the doc generator has been done, or you can do the docs in the old (wrong) way and then when the pull issue above is done it will also update this script to be correct.

Assets/VRTK/Scripts/Locomotion/VRTK_TouchpadMovement.cs
```diff
@@ -0,0 +1,480 @@
...
+            /// </summary>
+            WarpWithBlink
+        }
+
+        /// <summary>
+        /// Movement types that can be performed by the horizontal axis.
+        /// </summary>
+        public enum HorizontalAxisMovement
+        {
+            /// <summary>
```
thestonefox: same as above for this enum docs

Assets/VRTK/Scripts/Locomotion/VRTK_TouchpadMovement.cs
```diff
@@ -0,0 +1,480 @@
...
+            /// Performs an instant warp movement with a blink effect.
+            /// </summary>
+            WarpWithBlink
+        }   
+
+        /// <summary>
+        /// Which type axis movement did occur.
+        /// </summary>
+        public enum AxisMovementType
+        {
```
thestonefox: same as above

Assets/VRTK/Scripts/Locomotion/VRTK_TouchpadMovement.cs
```diff
@@ -0,0 +1,480 @@
...
+            /// User snap rotated.
+            /// </summary>
+            SnapRotate
+        }
+
+        /// <summary>
+        /// Which direction did the axis movement occur.
+        /// </summary>
+        public enum AxisMovementDirection
+        {
```
thestonefox: same as above

Assets/VRTK/Scripts/Locomotion/VRTK_TouchpadMovement.cs
```diff
@@ -0,0 +1,480 @@
...
+        public delegate void AxisMovementEventHandler(VRTK_TouchpadMovement sender, AxisMovementType movementType, AxisMovementDirection direction);
+
+        /// <summary>
+        /// Emitted when a warp, a flip direction or a snap rotate movement has successfully completed.
+        /// </summary>
+        public event AxisMovementEventHandler AxisMovement;
+
+        public bool LeftController
+        {
+            get { return leftController; }
```
thestonefox: as @bddckr pointed out, this requires to go on separate lines:

```
get
{
  return leftController;
}
```

Assets/VRTK/Scripts/Locomotion/VRTK_TouchpadMovement.cs
```diff
@@ -0,0 +1,480 @@
...
+            set
+            {
+                leftController = value;
+                SetControllerListeners(controllerLeftHand);
+            }
+        }
+
+        public bool RightController
+        {
+            get { return rightController; }
```
thestonefox: as @bddckr pointed out, this requires to go on separate lines:

```
get
{
  return rightController;
}
```

Assets/VRTK/Scripts/Locomotion/VRTK_TouchpadMovement.cs
```diff
@@ -0,0 +1,480 @@
...
+        private float lastFlip = 0;
+        private float lastSnapRotate = 0;
+        private CapsuleCollider bodyCollider;
+        private Transform headset;
+        private bool leftSubscribed;
+        private bool rightSubscribed;
+        private ControllerInteractionEventHandler touchpadAxisChanged;
+        private ControllerInteractionEventHandler touchpadUntouched;
+
+        private void Awake()
```
thestonefox: methods should follow their accessibility, so public methods first, then protected then private.

So these private methods need moving below the protected methods.

Assets/VRTK/Scripts/Locomotion/VRTK_TouchpadMovement.cs
```diff
@@ -0,0 +1,480 @@
...
+            SetControllerListeners(controllerRightHand);
+
+            bodyCollider = playArea.GetComponent<CapsuleCollider>();
+            if (!bodyCollider)
+            {
+                Debug.LogError("No body collider could be found in the play area. VRTK_BodyPhysics script is required in one of the scene GameObjects.");
+            }
+        }
+
+        protected void OnAxisMovement(AxisMovementType movementType, AxisMovementDirection direction)
```
thestonefox: Actually this is the only protected method, so it can be probably just moved up (or can it just be private?)

thestonefox:
I also have no idea if this is possible, but how viable is it to have this class just inherit the touchpad walking class and override and add to it to provide these additional features?

GitCop:

 * Commit: d285d55a31c5cbc2d7c215dd18f4b2253dbbd9be

---

thestonefox:
close :)

just needs line breaks in the description to keep it under 72 characters long :D

thestonefox:
This all looks good to me now, anything else you wanna change on it or are you happy for me to do a final play test and then merge it?

reznovVR:
It's all good now for an initial version so you can test and merge if no problems occur.

thestonefox:
Ok cool I'll try a bit later today

thestonefox:
I just added a speed multiplier to touchpad walking: https://github.com/thestonefox/VRTK/commit/a57af77675291ceee5f4eaf682c70e7deabbbdc3

Do you wanna add the feature into this PR as well?

reznovVR:
Yeah, did :)

thestonefox:
@reznovVR I tested it, all works fine.

The one thing i'm wondering though is.... should the speed multiplier be for more than just the slide speed?

Would it make sense also to have the warp go more steps if you have the speed button held down?

And the rotation go faster with the button down? (rotation could potentially be a separate setting maybe)

What do you think?

reznovVR:
Yeah it could work and is an option anyway. I'll also split the modifier so each axis has it's own since for horizontal actions the same modifier value could be too intense in some cases.

PR #809, opened 2017-01-11
feat(Editor): add quick select sdk option

reznovVR:
Adds buttons to quickly select the desired SDK in the SDK manager inspector.


thestonefox:
Needs a git commit description as well (`git commit --amend` then force push). Again hollar if you want help with the gitness

thestonefox:
I like the idea, but what do we do when there are more SDKs to choose from?

reznovVR:
4th SDK should fit nicely when the buttons are split in to two rows (2x2). I guess by the time 5th full feature set SDK is done the SDK manager has probably evolved to something else like maybe a single selection by default and overrides for single components when necessary :)

thestonefox:
Could the quick select work as a drop down menu rather than buttons?

virror:
I also think a drop-down would be better in the long run.

reznovVR:
Changed to drop down menu.

PR #818, opened 2017-01-13
docs(FAQ): add frequently asked questions document

thestonefox:
The new frequently asked questions document contains a collection
of commonly asked questions and answers to make it easier for users
to find solutions to common issues.


thestonefox:
@bddckr I've created a PR to bring FAQs into the main repo, lemme know what you think

(Also, I know the links can be put at the bottom and don't all need to be repeated and inline on github, but when I copy/paste it into readme.io then all the links get lost unless they're inline so it's pure laziness on my behalf why they're all inline :) )

PR #824, opened 2017-01-15
fix(Interaction): ensure a valid using object is present before haptics

StewMcc:
Brings it in line to the other AttemptHaptic functions (i.e on grab). Stops the object reference null error that can occur.

### DIFF
--- a/Assets/VRTK/Scripts/Interactions/VRTK_InteractUse.cs
+++ b/Assets/VRTK/Scripts/Interactions/VRTK_InteractUse.cs
@@ -144,13 +144,16 @@ namespace VRTK
             }
         }
 
-        private void AttemptHaptics()
-        {
-            var doHaptics = usingObject.GetComponentInParent<VRTK_InteractHaptics>();
-            if (doHaptics)
-            {
-                doHaptics.HapticsOnUse(controllerActions);
-            }
+		private void AttemptHaptics()
+		{
+			if (usingObject) 
+			{ 
+				var doHaptics = usingObject.GetComponentInParent<VRTK_InteractHaptics>();
+				if (doHaptics)
+				{
+					doHaptics.HapticsOnUse(controllerActions);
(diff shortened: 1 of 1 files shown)


thestonefox:
I'm not sure how you're even getting this error as `AttemptHaptics` is only triggered in the `UseInteractedObject` which checks to see if the touchedObject is usable.

If you have no touched object then it can't be usable, and therefore the Using Object is not set which means AttemptHaptics would never be attempted to run in the first place.

Any chance you can provide instructions on how to recreate the issue?

StewMcc:
public override void StartUsing(GameObject usingObject)
{
    base.StartUsing(usingObject);
    spinSpeed = 360f;
    gameObject.SetActive(false);    // problem line
}
In WhirleyGig.cs will cause the error message to appear.

thestonefox:
Why would you disable the object as soon as you start using it?

I'm happy to merge your PR though, the extra check doesn't hurt.

Can you just improve the git commit message though to have a description as well. something like:

```
fix(Interaction): ensure a valid using object is present before haptics

If the using object is set to inactivate at the point StartUsing is
called and using haptics have also been set then the AttemptHaptics
script will throw an error as the usingObject is null.
```

You will need to `git commit --amend` your pushed commit and then force push to change the git commit message (changing it just on github isn't enough to actually change the git history).

StewMcc:
I just disable the object, instead of disabling everything on it individual to ensure nothing else gets called when its used. Seemed simplest approach for the simple Research project a client is doing. Doubt it would just disappear in a game admittedly, but project needs as little variation as possible.

thestonefox:
@StewMcc if you wanna just update the commit message (to what I put in the above comment) then I'll merge it in.

GitCop:

 * Commit: fbf1f8d5c5456e50d633329907fbe1d235343a2b

---

PR #826, opened 2017-01-15
Set highlight's parent before setting transf

BenWoodford:
I don't know if there was a specific reason it was being done in the other order, but I just ran into this one - it loses the local scale of the model because Unity will re-calculate scales when you set the parent. So you need to set the parent first.

Case where it was uses: Gameobject with scale of 0.015 holding mesh of scale 1 which was the interactable. When touching it, spawned highlighted mesh was at scale 66x because it would recalculate its scale to match the previous scale of 1x the root, rather than 1x the new mesh.


GitCop:

 * Commit: 253976b55a2853cc389a0dba851215f60f1f08a3

---

BenWoodford:
SourceTree managed to foobar my last attempt to amend the commit, so going to try again.

That'll teach me not go use the command line.

PR #827, opened 2017-01-15
fix(Highlighters): ensure cloned object parent is set first

BenWoodford:
I don't know if there was a specific reason it was being done in the other order, but I just ran into this one - it loses the local scale of the model because Unity will re-calculate scales when you set the parent. So you need to set the parent first.

Case where it was uses: Gameobject with scale of 0.015 holding mesh of scale 1 which was the interactable. When touching it, spawned highlighted mesh was at scale 66x because it would recalculate its scale to match the previous scale of 1x the root, rather than 1x the new mesh.

### DIFF
--- a/Assets/VRTK/Scripts/Interactions/Highlighters/VRTK_OutlineObjectCopyHighlighter.cs
+++ b/Assets/VRTK/Scripts/Interactions/Highlighters/VRTK_OutlineObjectCopyHighlighter.cs
@@ -145,10 +145,10 @@ namespace VRTK.Highlighters
             }
 
             highlightModel = new GameObject(name + "_HighlightModel");
+            highlightModel.transform.SetParent(transform);
             highlightModel.transform.position = copyModel.transform.position;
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 8ef46dbad79b7df06a568724408dc7f5c52796c7

---

thestonefox:
The commit message needs changing to adhere to the contribution guidelines.

you'll need to `git commit --amend` your commit and the force push the change.

A git message like:

```
fix(Highlighters): ensure cloned object parent is set first

The OutlineObjectCopyHighlighter cloned object needs to have the
parent set before the scale/rotation/position are done otherwise
the scale isn't set correctly.
```

BenWoodford:
Done

BenWoodford:
I have now realised I missed a bit to this, as it does not set the scale to 1,1,1 /after/ setting the parent then a parent that does not have a scale of 1,1,1 will see the copy be smaller/bigger. Following up with #828

PR #831, opened 2017-01-18
fix(SDK): controller retrieval and touch events in simulator

burnedikt:
When trying to develop without any VR-Hardware, the simulator comes in really handy. However, I noticed that some events on the simulator (e.g. touchdown, touchup, touched on any buttons) are never fired. While it makes sense to not fire those events from a technical perspective (since we cannot distinguish between press and touch on a keyboard) it makes developing a whole lot harder, since a developer potentially has to use different types of events during development (Simulator) and in production (HMD). 

Furthermore, the Simulator SDK right now does not allow for finding script alias controllers. Whenever the calls to get a controller (e.g. `GetControllerLeftHand` or `GetControllerByIndex`) are used, the boolean `actual` parameter is simply ignored and the actual controllers are always returned. This is not in line with the base implementation or the implementation for SteamVR.

This PR mitigates both the above problems.

- Touch Events are now always fired alongside with pressed / released events
- Controller Retrieval / Finding logic has been reworked to be more consistent with other implementations

### DIFF
--- a/Assets/VRTK/SDK/Simulator/SDK_SimController.cs
+++ b/Assets/VRTK/SDK/Simulator/SDK_SimController.cs
@@ -21,6 +21,9 @@ namespace VRTK
             {"StartMenu", KeyCode.F }
         };
 
+        protected const string RIGHT_HAND_CONTROLLER_NAME = "RightHand";
+        protected const string LEFT_HAND_CONTROLLER_NAME = "LeftHand";
+
         /// <summary>
         /// The ProcessUpdate method enables an SDK to run logic for every Unity Update
         /// </summary>
@@ -141,11 +144,12 @@ namespace VRTK
         /// <returns>The GameObject containing the left hand controller.</returns>
         public override GameObject GetControllerLeftHand(bool actual = false)
         {
-            GameObject controller = null;
-            GameObject simPlayer = SDK_InputSimulator.FindInScene();
-            if (simPlayer != null)
+            // use the basic base functionality to find the left hand controller
+            var controller = GetSDKManagerControllerLeftHand(actual);
+            // if the controller cannot be found with default settings, try finding it below the InputSimulator by name
+            if (!controller && actual)
             {
-                controller = simPlayer.transform.FindChild("LeftHand").gameObject;
+                controller = GetActualController(ControllerHand.Left);
             }
 
             return controller;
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: ffd2c10e6cfc6bffd1598bfa25161f900007f24d

 * Commit: ec004478e3379c2abfddcfe5b2628b8977ee0fa0

---

thestonefox:
@burnedikt  is it the best thing to do to map the touch just to the press? would that help at all? or would it be better to map different buttons to the touch instead?

burnedikt:
@thestonefox Intuitively I'd say another key would be overkill, since naturally, a touch inevitably has to occur before a press can ever be detected. *One cannot press a button without first touching it.*

However, I see your point of having a possibility to trigger the touch events separately during development to test specific or timed behaviour (e.g. an Animation which starts `onTouchStart` and ends `onPressDown` would be hard to test, if the events occurred in such quick succession).

So in my opinion:
1. Touch-Events should occur naturally together when a button pressed (Principle of least astonishment)
2. You're right in that a way of just having the touch event is required as well. If someone just lightly touches a button, it'll never be pressed.

In order to avoid adding separate Touch Keys for all mapped buttons, I'd suggest we add a modifier key, which - when pressed together with the actual button - opts out of the pressing feature and just fires the touch events.

What do you think?

thestonefox:
@burnedikt yeah @virror also suggested a modifier, it could be the way to go..

What do you think @virror  ?

virror:
I like it. I think its a good addon to as you said to make the button trigger when you let go of the modifier to better emulate the real behavior. Im all for it.

burnedikt:
Just pushed the update: 

* When Pressing `T` (TouchModifier) + any Button (e.g. `Mouse0`), only touch events will be fired
* When Pressing `H` (HairTouchModifier) + any Button (e.g. `Mouse0`), only touch and "hair" events will be fired
* When Pressing none of the Modifiers and just a Button (e.g. `Mouse0`), all events will be fired

In any case, I also implemented event ordering, since a press cannot occur before a touch. The order right now (see https://github.com/thestonefox/VRTK/pull/831/commits/9732473409c1a7f3262e038cf94693d5ca8ee45e) is (where events in brackets are only fired for some buttons)

1. (touchDown)
2. (hairlineDown)
3. pressDown
4. (click)
5. pressUp
6. (hairlineUp)
7. (touchUp)

burnedikt:
One issue with the implementation of https://github.com/thestonefox/VRTK/pull/831/commits/bea4f0811ad995ee31ee3cc986d3f7f1c3cc2341 is: 

If someone wants to have the trigger event and holds down `Mouse0` accordingly, then notices midway that he actually just wanted touch events and forgot to hit the TouchModifier `T` as well, then hits and holds `T` while releasing `Mouse0`, he will get the `touchDown`, `hairDown`, `pressDown` events but only `touchUp` in the end. Vice versa, if the Modifier Key is released too early, the full event lifecycle will be fired as soon as the button (`Mouse0`) is released. 

So in general, the modifier keys need to be *pressed down before* the actual key and *released after* the actual key to get the expected results.

thestonefox:
Is this commit: 

https://github.com/burnedikt/VRTK/commit/11120f2bf49e3eaca41e3f3fcea4dc5a0efaab89

still required? as it's overriden by a newer commit anyway?

If you can rebase and remove it without breaking anything and still have a pull request that works then that would be cool.

If it's a lot of hassle then i'm sure it can be overlooked ;)

burnedikt:
rebased

virror:
Looks nice : )

PR #828, opened 2017-01-16
fix(Highlighters): Use local scale of 1,1,1 to match parent size

BenWoodford:
The copy highlighter needs a scale of 1,1,1 to match the size of the parent, otherwise if the parent doesn't have a scale of 1,1,1 it'll be smaller/bigger than the parent.

PR #850, opened 2017-01-20
fix(Interaction): parent of the outline highlighter cloned object

gpvigano:
Change the parent of the OutlineObjectCopyHighlighter cloned object to
the same parent as the original object.

Change the parent setting to not re-compute world position.

Change the transform assignments from world to local (faster), because
now the parent is the same.

The OutlineObjectCopyHighlighter cloned object needs to have the
parent set to the same parent of the customOutlineModel to exactly
match its shape. Even the PR #828 still causes problems with a hierarchy
of game objects with different scaling (e.g customOutlineModel with its
own scaling as child of the game object with its own scaling, holding
the OutlineObjectCopyHighlighter component).

### DIFF
--- a/Assets/VRTK/Scripts/Interactions/Highlighters/VRTK_OutlineObjectCopyHighlighter.cs
+++ b/Assets/VRTK/Scripts/Interactions/Highlighters/VRTK_OutlineObjectCopyHighlighter.cs
@@ -145,10 +145,10 @@ namespace VRTK.Highlighters
             }
 
             highlightModel = new GameObject(name + "_HighlightModel");
-            highlightModel.transform.SetParent(transform);
-            highlightModel.transform.position = copyModel.transform.position;
-            highlightModel.transform.rotation = copyModel.transform.rotation;
-            highlightModel.transform.localScale = Vector3.one;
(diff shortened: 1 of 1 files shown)


Assets/VRTK/Scripts/Interactions/Highlighters/VRTK_OutlineObjectCopyHighlighter.cs
```diff
@@ -145,10 +145,10 @@ private void CreateHighlightModel()
             }
 
             highlightModel = new GameObject(name + "_HighlightModel");
-            highlightModel.transform.SetParent(transform);
-            highlightModel.transform.position = copyModel.transform.position;
-            highlightModel.transform.rotation = copyModel.transform.rotation;
-            highlightModel.transform.localScale = Vector3.one;
+            highlightModel.transform.SetParent(copyModel.transform.parent, false);
```
thestonefox: This line has actually broken the Outline Highlighter. If you set it as the parent then the outline no longer follows the object around.

thestonefox:
I'm going to revert this commit as it's causing too many issues.

PR #861, opened 2017-01-21
Fix Panel Orientation on all platforms

burnedikt:
Fxies an issue where the PanelMenuController would only work for the SteamVR platform.

Instead of searching for hardcoded objects, use the SDK Manager to find the correct gameobject.

### DIFF
--- a/Assets/VRTK/Scripts/Controls/2D/PanelMenu/PanelMenuController.cs
+++ b/Assets/VRTK/Scripts/Controls/2D/PanelMenu/PanelMenuController.cs
@@ -1,4 +1,4 @@
-// Panel Menu Controller|Prefabs|0070
+﻿// Panel Menu Controller|Prefabs|0070
 namespace VRTK
 {
     using System.Collections;
@@ -106,7 +106,7 @@ namespace VRTK
             {
                 if (rotateTowards == null)
                 {
-                    rotateTowards = GameObject.Find("Camera (eye)");
+                    rotateTowards = VRTK_DeviceFinder.HeadsetCamera().gameObject;
                     if (rotateTowards == null)
                     {
                         Debug.LogWarning("The PanelMenuController could not automatically find an object to rotate towards.");


Assets/VRTK/Scripts/Controls/2D/PanelMenu/PanelMenuController.cs
```diff
@@ -106,7 +106,7 @@ protected virtual void Update()
             {
                 if (rotateTowards == null)
                 {
-                    rotateTowards = GameObject.Find("Camera (eye)");
+                    rotateTowards = VRTK_SDKManager.instance.GetHeadsetSDK().GetHeadsetCamera().gameObject;
```
thestonefox: Lol i knew i'd speak too soon.. Looking at this, you should be using `VRTK_DeviceFinder.HeadsetTransform().gameObject`

Nothing should really ever go through the SDK Manager if it can help it

burnedikt: I see. Updated, using the headset camera instead of the headset itself though.

Assets/VRTK/Scripts/Controls/2D/PanelMenu/PanelMenuController.cs
```diff
@@ -106,7 +106,7 @@ protected virtual void Update()
             {
                 if (rotateTowards == null)
                 {
-                    rotateTowards = GameObject.Find("Camera (eye)");
+                    rotateTowards = VRTK_SDK_Bridge.GetHeadsetCamera().gameObject;
```
thestonefox: This still should be going through the DeviceFinder. replace with `VRTK_DeviceFinder.HeadsetCamera().gameObject`

burnedikt: mea culpa

PR #863, opened 2017-01-21
chore(Structure): ensure MonoBehaviour magic methods are overridable - Resolves #835

### LINKED ISSUE #835: Ensure Unity magic methods are all `protected virtual` in VRTK scripts
https://docs.unity3d.com/ScriptReference/MonoBehaviour.html

If any VRTK script uses a MonoBehaviour method and is set to private, it then means it's not possible to override that method.

thestonefox:
All of the VRTK scripts that utilise MonoBehaviour magic methods
e.g. Awake, Update, OnCollisionEnter have all been changed to at least
protected virtual so any VRTK script can be inherited and any of the
magic methods can be overriden.


bddckr:
It looks like `SDK_FallbackBoundaries` still has a `private void Awake()`... No idea if that's intended.

Issue #842, opened 2017-01-19
Improve SDK button mapping


bddckr:
After some discussion on Slack about the current state of button mapping in VRTK, SteamVR, OpenVR and Unity it looks like there is a problem with the current button mapping somewhere because a button is mapped "wrong" when using SteamVR vs. Oculus SDK.

That discussion led to an idea by @BenWoodford:

> what like Button1 has a list (or mask so it uses Unity's multi select stuff?) you can add SDK buttons to? And each SDK in the project (I guess you'd have to have a list of SDKs in the VRTK manager script that you can select which ones you want to use from)
>
> At a higher level, VRTK_SDKManager would give you a list of SDKs (in an ideal world, it wouldn't be a hardcoded list so you could drop in something like a PS4 easily, idk how possible that is though)
> And you check the SDKs you want to use
> Then there's a list of Button Aliases, and you can select which buttons from each SDK apply to each alias
> So you could set Button1 to be App menu on Vive, X or whatever on Touch, X on PSVR
> etc
>
> Hell, you could make X and Square on PSVR the same button then
> Ah yeah, a section for each SDK may be a bit cleaner than having every button from every SDK under one Bitmask selector for each alias.
>
> Like I said, bonus points if you can make the whole SDK process plugin-based so that us PSVR registered folk could write a PSVR plugin that VRTK doesn't need to know about (and shouldn't, as you wouldn't be able to bundle it I suspect, worth checking with Sony though)
> It could just be a List<SDK_BaseSystem> for that though, and you add the relevant SDKs to that from what the user has in their library

Some ideas...
Change the base SDK:
- Add `abstract string name` that identifies the SDK.
- Add `abstract Type buttonEnumType` that returns the type of the custom SDK button enum (which uses an `int` as the underlying type).

On every editor script compile (and once in a build at runtime) get every type, search for subclasses of the base SDK. Cache the base SDK subclasses by name in a `Dictionary<string, Type>`.

Now for the button alias stuff in VRTK you replace it with something like `Dictionary<string, int>` which maps the buttons to use to the SDK name.
That will get a custom editor inspector that uses the above cached dictionary so you don't type in the SDK name. You select the SDK from a dropdown instead.

At runtime the button alias thing will be used by looking up the SDK and using the provided enum bitmask value...

Things to use to make this work:
https://docs.unity3d.com/ScriptReference/Callbacks.DidReloadScripts.html
[...]

PR #867, opened 2017-01-23
feat(SDK): allow for dynamic SDK discovery

bddckr:
This is a part of #842.

---

The list of supported SDKs in VRTK is fixed because they are hard coded
into the SDK manager (and its editor). This change modifies the SDK
manager so it looks up the available SDKs by checking the code base for
classes that inherit from one of the base SDK classes. This allows to
create and use additional SDKs without touching the existing VRTK
codebase at all.

Sharing of SDK implementations in a private group of people (e.g. where
everyone signed an NDA) is now possible without needing to share patches
to the VRTK codebase. There is no reason anymore to keep up with changes
to the parts of VRTK one changed to use their own SDK implementations.

To allow for easy specifying of SDK implementations a new "SDK
description" attribute is added. Another "predicate" attribute allows to
mark a static method as a predicate for SDK purposes. The method signals
whether to add or remove the scripting define symbol that is specified
by the attribute by returning true or false respectively. Marking the
same method with multiple predicate attributes is allowed.

The existing SDK implementations have been updated to make use of both
attributes. Usage of the SteamVR plugin in VRTK is now simplified by
using the predicate attribute to add a scripting define symbol that
defines what version of the plugin is currently used in addition to
defining the generic SteamVR SDK symbol.

If a set SDK doesn't exist anymore (e.g. the SDK class was removed) the
SDK manager will use the corresponding fallback SDK instead and log an
error so the problem is known about. The editor shows that error, too,
but won't remove the set SDK from its list of selectable SDKs until the
user overrides it with another SDK. It shows the name of the missing
class in that case instead.

The SDK manager's editor is updated to allow selecting one of the
discovered SDKs. Logic that previously handled the scripting define
symbols is now moved to the SDK manager class itself in form of public
methods. This allows to build scripts that handle automatic building for
different stores as Oculus Home doesn't allow to include SteamVR/OpenVR
libraries, for example. The automatic populating of the SDK object
references like the Linked Objects and Alias Objects is moved from the
editor to a new setting on the SDK manager itself. This has a public
method to trigger manually, too.

Additionally the editor was changed to support undo/redo for all the
possible changes. Proper error messages and hints are now shown in the
editor if needed, too.

The existing auto-managing of the scripting define symbols is now done
whenever needed. This means loading a scene with an active SDK manager
in it will make sure to manage the symbols if the SDK managers setting
says to do so. The behavior for the auto-populating of the SDK object
references is the same with this change. The idea is to always make sure
the current setup is correct for a used SDK manager.

---

SDKs themselves:
- Now inherit from the new `public abstract class SDK_Base : ScriptableObject { }` to give some type safety for the various kinds of SDKs (system, boundaries, headset, controller).
- SDKs specify their "description" by adding a C# attribute called `SDK_DescriptionAttribute` to their class. The existing SDKs are updated to have this attribute. Each SDK has their own attribute because VRTK allows to implement an SDK without implementing all the other SDK kinds with it (e.g. it's possible to provide a controller SDK without a System, Boundaries or Headset SDK).

`VRTK_SDKManager`:
- doesn't need to run in the editor anymore now:
`PopulateAvailableAndInstalledSDKInfos` is added and is called in the editor and on build startup to make sure the public static fields `available*SDKInfos` and `installed*SDKInfos` are **always** up to date.
- now is able to be used in the build pipeline easily with the new method `ManageScriptingDefineSymbols`. This will make sure the current scripting define symbols are set for the current SDKs. This method is also used whenever something is changed using the inspector on the SDK manager.
- The SDKs to use are adjustable using new `*SDKInfo` properties like `systemSDKInfo`. They are of a new type called `SDKInfo<BaseType>`. You create one by providing the type of an SDK using `SDKInfo<BaseType>.FromType<ActualType>`. An example would be `SDKInfo.FromType<SDK_SteamVRController>`.
- `SDKInfo<BaseType>` is serialized using the `ActualType`'s [`FullName`](https://msdn.microsoft.com/en-us/library/system.type.fullname(v=vs.90).aspx). That allows to uniquely identify the SDK that is set. That's what `AvailableSDKInfo<BaseType>.FromTypeName<FallbackType>` is for.
- `SDK_DescriptionAttribute.Fallback` represents a fallback SDK (for all kinds of SDK endpoints).
- If a set SDK isn't available anymore the fallback SDK is automatically used and an error is logged. Additional errors are only logged when the fallback SDK is actually used (by the instantiation of the current set SDK via the `Get*SDK` methods). The set SDK that was serialized isn't changed to the fallback SDK and left as is in that case to allow for fixing stuff without changing the setup.
- All the SDK infos and SDKs themselves are now cached. The cache is invalidated whenever needed of course.

`VRTK_SDKManagerEditor`:
- is updated to allow to set SDKs by providing a list of the available ones.
- allows proper usage of undo/redo for all the features now (it even automatically switches the scripting define symbols according to the undo/redo!).
- explains why only a subset of all the SDKs is available in the Quick Select dropdown. It only shows this explanation when it's actually needed, i.e. there are some SDKs not available for all the endpoints/kinds (System, Boundaries, Headset, Controller).
- explains when a set SDK is not available and therefore tells you that it's falling back to the fallback SDK instead. It leaves the setting for that missing SDK alone, though. So the missing SDK can actually be seen by type name and is only removed from the SDK list after setting another SDK in its place.

### DIFF
--- a/Assets/VRTK/Editor/VRTK_EditorUtilities.cs
+++ b/Assets/VRTK/Editor/VRTK_EditorUtilities.cs
@@ -22,9 +22,13 @@
             AddHeader(headerAttribute == null ? displayName : headerAttribute.header);
         }
 
-        public static void AddHeader(string header)
+        public static void AddHeader(string header, bool spaceBeforeHeader = true)
         {
-            EditorGUILayout.Space();
+            if (spaceBeforeHeader)
+            {
+                EditorGUILayout.Space();
+            }
+
             EditorGUILayout.LabelField(header, EditorStyles.boldLabel);
         }
     }
--- a/Assets/VRTK/Editor/VRTK_SDKManagerEditor.cs
+++ b/Assets/VRTK/Editor/VRTK_SDKManagerEditor.cs
@@ -1,50 +1,129 @@
 ﻿namespace VRTK
 {
-    using UnityEngine;
     using UnityEditor;
-    using System.Collections.Generic;
+    using UnityEngine;
     using System;
+    using System.Collections.Generic;
+    using System.Collections.ObjectModel;
     using System.Linq;
+    using System.Reflection;
 
     [CustomEditor(typeof(VRTK_SDKManager))]
     public class VRTK_SDKManagerEditor : Editor
     {
-        private SDK_BaseHeadset previousHeadsetSDK;
-        private SDK_BaseController previousControllerSDK;
-        private SDK_BaseBoundaries previousBoundariesSDK;
-        private VRTK_SDKManager.SupportedSDKs quicklySelectedSDK = VRTK_SDKManager.SupportedSDKs.None;
+        private const string SDKNotInstalledDescription = " (not installed)";
+        private const string SDKNotFoundAnymoreDescription = " (not found)";
 
         public override void OnInspectorGUI()
         {
             serializedObject.Update();
 
-            //Get actual inspector
-            VRTK_SDKManager sdkManager = (VRTK_SDKManager)target;
-
-            EditorGUILayout.BeginVertical("Box");
+            var sdkManager = (VRTK_SDKManager)target;
 
             EditorGUILayout.PropertyField(serializedObject.FindProperty("persistOnLoad"));
-            EditorGUILayout.PropertyField(serializedObject.FindProperty("autoManageScriptDefines"));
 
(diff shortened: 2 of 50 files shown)


bddckr:
Updated for SteamVR plugin version 1.2.1 and the Daydream SDK that got merged on master.

thestonefox:
Too risky to go into 3.1.0 :(

bddckr:
_shakes fist_
I'm OK with that. 😀

PR #888, opened 2017-01-29
fix(UnityEvents): ensure all unity event helpers initialise event.

thestonefox:
Previously, the Unity Event Helper scripts were not initialising the
event parameters meaning that if the events were used via script then
they would throw a null reference error. They would work in the editor
as Unity would provide a default state for them.

This fix ensures that all events set up a default event and are
no longer null.


thestonefox:
Radial menus don't use c# delegate/unity event helpers (yet)

so until they are updated they won't work just yet

brydaman:
I added in the UnityEvent() initialization to each of the Radial Menu events and it works for me. Previously I couldn't addListener to dynamically generated Radial Menu buttons, once I added in default initialization it worked.

thestonefox:
Yeah that will work as a temporary fix. But the unity events need replacing with c# delegates and a unity event helper provided.

PR #889, opened 2017-01-29
fix(Teleporter): ensure listener coroutine is not run if disabled - fixes #884

### LINKED ISSUE #884: VRTK_BasicTeleporter: destination marker registration timing issue
Due to the delay in registration with destination markers, if a teleporter is disabled during its first tick, it registers with all listeners at the end of the frame even though it's supposed to be deregistered.  This causes trouble with, for example, having multiple teleporters and swapping between them at runtime via enabling/disabling the components.  The timing issue can be resolved by changing the init coroutine as follows:

```c#
private IEnumerator InitListenersAtEndOfFrame()
{
    yield return new WaitForEndOfFrame();
    if (enabled)
    {
        InitDestinationMarkerListeners(true);
    }
}
```

thestonefox:
The Teleporter script runs a co-routine at the end of the frame to
register any destination markers. However, if the script is disabled
within the first frame then the co-routine is still executed and
registers the listeners even if they're not required.

This fix only registers the listeners if the teleporter script is
enabled and the co-routine is also cancelled when the script is
disabled.

### DIFF
--- a/Assets/VRTK/Scripts/Locomotion/VRTK_BasicTeleport.cs
+++ b/Assets/VRTK/Scripts/Locomotion/VRTK_BasicTeleport.cs
@@ -57,6 +57,7 @@ namespace VRTK
         private float fadeInTime = 0f;
         private float maxBlinkTransitionSpeed = 1.5f;
         private float maxBlinkDistance = 33f;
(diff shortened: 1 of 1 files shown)


cameronoltmann:
Looks good to me

PR #902, opened 2017-01-31
Fix/player motion - resolves #853

### LINKED ISSUE #853: Ensure momentum in Move In Place script is carried on when the activator button is released
At the moment if you have momentum using Move In Place and then release the activator button, you come to an abrupt halt which is jarring.

It should keep the momentum going and decelerate.

thestonefox:


SuperSonic68:
Woo! Can't wait to give this a go!

PR #903, opened 2017-02-01
fix(Climbing): add moving and rotating to VRTK_PlayerClimb

mattboy64:
Climbable move points will now move and rotate the player as they
do likewise around the game world. You can disable the rotation of
the player (for comfort mode) by the inspector toggle named
useGrabbedObjectRotation in the VRTK_PlayerClimb class. With it off
the player will only pivot around the grabbed location, but will
not rotate the view at all.


thestonefox:
@mattboy64 is this PR a combination of the other two referenced PRs as well? Or does this do something specific?

thestonefox:
@mattboy64 also is it worth updating the example scene demonstrating the feature?

mattboy64:
@thestonefox no, this is my take on it. It does a very similar thing to the others (having movable climb spots), but also throws rotational support in as well.

As far as updating the example scene, it might be worth it as long as you don't mind another helper script or 2 being added to the project.

thestonefox:
@mattboy64 where would the helper scripts go? If they're going in the `Examples/Scripts` folder then go nuts! :)

AlbertMN:
Never have I ever been so close to falling in VR - this is great. I've tested it for about an hour now, and I can't seem to spot anything that needs changing.
Leaving "Use Grabbed Object Rotation" on, swining in a rope consisting of configurable joints that moves/"reacts" to the controller force is the _ultimate_ drunken-simulator.

mattboy64:
I've updated example scene 37 with a hand held lift and zip line. I've also changed movement from Update() to LateUpdate(), move the rotational toggle from a global paremeter in PlayerClimb to a per object one in VRTK_ClimbableGrabAttach, and updated the documentation md file.

DOCUMENTATION.md
```diff
@@ -3592,6 +3592,10 @@ The ProcessFixedUpdate method is run in every FixedUpdate method on the interact
 
 The Climbable Grab Attach script is used to mark the object as a climbable interactable object.
 
+### Inspector Parameters
```
mattboy64: Do I need to revert the DOCUMENTATION.md change or will that get auto updated? Fixed all the rest.

AlbertMN:
This has one problem, that there should at least be an option for;
With "Use Grabbed Object Rotation" on, you keep the rotation when let you go of the object. So, if I swing in a vine, and let go while the vine is at ~90 degrees, it leaves me lying on the floor.

mattboy64:
@AlbertMN yeah, I noticed that too. That's why I default the setting to OFF now. It works fine for rotation around the Y axis, but there really needs to be a script attached to the player trying to right the player when not grabbing an object. Rotation like that is really, really discouraged though since it can cause motion sickness, but I thought I'd at least have it as an option so that you could have Y rotation. I'm hoping somebody that actually needs the feature will be willing to build on the idea *wink* *wink* 😄

Assets/VRTK/Examples/Resources/Scripts/MoveBlock.cs
```diff
@@ -0,0 +1,70 @@
...
+    public float waitTime = 5.0f;
+    public float rotateSpeed = 10.0f;
+
+
+    private float startY;
+    private bool goingUp = true;
+    private float stoppedUntilTime;
+    private float moveUpAmount;
+
+	void Start ()
```
thestonefox: Needs the `protected virtual` on it

Assets/VRTK/Examples/Resources/Scripts/MoveBlock.cs
```diff
@@ -0,0 +1,70 @@
...
+        if(moveYAmount < 0.0f)
+        {
+            startY -= moveYAmount;
+            goingUp = false;
+        }
+
+        stoppedUntilTime = Time.time + waitTime;
+	}
+	
+	void Update ()
```
thestonefox: Needs the `protected virtual` on it

Assets/VRTK/Examples/Resources/Scripts/Zipline.cs
```diff
@@ -0,0 +1,62 @@
...
+        private float currentSpeed;
+
+        public override void OnInteractableObjectGrabbed(InteractableObjectEventArgs e)
+        {
+            base.OnInteractableObjectGrabbed(e);
+
+            isMoving = true;
+        }
+
+        protected void Start()
```
thestonefox: Needs the `protected virtual` on it

DOCUMENTATION.md
thestonefox: Just revert it so it's not in the commit

mattboy64:
K, those changes should all be done.

Assets/VRTK/Examples/Resources/Scripts/MoveBlock.cs
```diff
@@ -0,0 +1,70 @@
...
+    public float moveYAmount = 20.0f;
+    public float moveSpeed = 1.0f;
+    public float waitTime = 5.0f;
+    public float rotateSpeed = 10.0f;
+
+
+    private float startY;
+    private bool goingUp = true;
+    private float stoppedUntilTime;
+    private float moveUpAmount;
```
thestonefox: there seems to be some crazy spacing here, my guess is it's a mix of tabs and spaces.

did you put tabs in front of the protected virtual? if so that's what's causing the crazy spacing :)

mattboy64: I don't have tabs turned on so not sure about that. I deleted the extra spaces though.

mattboy64: Oh, I see what you're saying. I started with the whirlygig.cs file so maybe there are some lingering tabs in there. All tabs should be removed from MoveBlock.cs now though.

thestonefox:
Looks ready to merge, i just need to clean some stuff up my end then I'll merge it!

PR #920, opened 2017-02-06
fix(Highlighters): Added option to merge submeshes on copy

BenWoodford:
Previously was only highlighting the first submesh found, rather than
the whole mesh. Added as a bool option to avoid any unforeseen
consequences for current users

Not to be confused with multiple MeshRenderers/objects,
that's a different problem altogether.

### DIFF
--- a/Assets/VRTK/Scripts/Interactions/Highlighters/VRTK_OutlineObjectCopyHighlighter.cs
+++ b/Assets/VRTK/Scripts/Interactions/Highlighters/VRTK_OutlineObjectCopyHighlighter.cs
@@ -1,4 +1,4 @@
-﻿// Outline Object Copy|Highlighters|40030
+// Outline Object Copy|Highlighters|40030
 namespace VRTK.Highlighters
 {
     using UnityEngine;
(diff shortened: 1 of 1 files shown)


Assets/VRTK/Scripts/Interactions/Highlighters/VRTK_OutlineObjectCopyHighlighter.cs
```diff
@@ -163,7 +165,25 @@ private void CreateHighlightModel()
...
+                        CombineInstance ci = new CombineInstance();
+                        ci.mesh = copyMesh.mesh;
+                        ci.subMeshIndex = i;
+                        ci.transform = copyMesh.transform.localToWorldMatrix;
+                        combine.Add(ci);
+                    }
+
+                    highlightMesh.mesh = new Mesh();
+                    highlightMesh.CombineMeshes(combine.ToArray(), true, false);
+                } else {
```
thestonefox: this needs to be on separate lines

```
}
else
{
```

Assets/VRTK/Scripts/Interactions/Highlighters/VRTK_OutlineObjectCopyHighlighter.cs
```diff
@@ -21,6 +21,8 @@ public class VRTK_OutlineObjectCopyHighlighter : VRTK_BaseHighlighter
         public GameObject customOutlineModel;
         [Tooltip("A path to a GameObject to find at runtime, if the GameObject doesn't exist at edit time.")]
         public string customOutlineModelPath = "";
+        [Tooltip("If your mesh has multiple submeshes you wish to highlight, you'll want to check this otherwise only the first mesh will be highlighted")]
```
thestonefox: Don't use words like `you` and `you'll` refer like:

`If the mesh has multiple sub-meshes to highlight then this should be checked, otherwise only the first mesh will be highlighted.`

BenWoodford:
Updated as per request.

PR #927, opened 2017-02-15
fix(UI): allow non-VR UI usage

bddckr:
This fix makes it possible to use Unity UI in the regular, non-VR way
while making sure the VR UI interactions provided by VRTK still work.

Since this introduces a new class `VRTK_EventSystem` the existing logic
to disable the auto-pausing behavior in Unity 5.5 and above is moved
into the new class. Possible confusion regarding the naming of the
existing `VRTK_EventSystemVRInput` and the new class is resolved by
changing the name of the existing class to `VRTK_VRInputModule`. An
unused constant of `VRTK_UIPointer` is also removed.


bddckr:
I only seem to get canvases that are set to "screen space - overlay" to work right now. Maybe I'm just doing it the wrong way to allow for world space canvases.

Let me know if this works for you or if you know how to make world space canvases work with the mouse. 😄

thestonefox:
I think @narfanator has done some work on this in the past. Worth hooking up and comparing notes

GloryFish:
I integrated these changes into my project and simultaneous UI interaction is now working where it wasn't before.

I have two canvases. Both are World Space. Mouse interaction works on the desktop canvas and interaction with the VR canvas works with the new VRTK_UIPointer script. I am not getting any errors running this and made no other changes to my project. 

I'll continue to run this code and if I notice any issues I'll report back but it seems very solid as is. Thanks kindly for the work on this, it's going to save me a lot of time.

bddckr:
Thanks for testing @GloryFish!
I just checked: If I actually create a second camera for desktop only and make sure to set that one as the "Event Camera" of the canvas it works for world space, too. :tada:

This means this should be ready to merge @thestonefox 😄

jeremedia:
This is awesome.

thestonefox:
Sorry :( i think my pointer PR created conflicts in here :( i shudda checked first :(

narfanator:
I can't tell what's going on this (yet). Still working my way through the Pointer codebase.

My best guess at  this time is that you've found a way to make the normal Unity event system / input module work while VR is active?

bddckr:
@narfanator Yes:
> This fix makes it possible to use Unity UI in the regular, non-VR way
while making sure the VR UI interactions provided by VRTK still work.

VRTK provides a way to use Unitys UI in VR using pointers or directly interacting with it via the controllers. This change just makes it possible to still be able to use Unity UI using mouse and keyboard.

It does this by making sure to call `Process` on both the `StandaloneInputModule` and the one VRTK provides. This only works correctly because the VRTK VR input handling is doing most of its work without using `EventSystem`. `EventSystem` is not made to allow multiple active input modules... So this solution is kinda hacky, but works :)


BTW: I will update this PR sometime next week so it works on the current master again. I'm currently having a terrible cold. :(

narfanator:
I could've sworn that just calling Process on Standalone didn't work :/

bddckr:
@narfanator Yeah you have to make sure to change the private field that references the current event system of all the input modules. Otherwise those input modules aren't able to correctly communicate their stuff with the event system.

As long as the VR input module VRTK has doesn't use the selection stuff of the event system everything is fine. The selection stuff is handled partially by he current event system which totally isn't written to support multiple input modules, so stuff would break.
That's the reason this fix works for the VRTK VR input module, but wouldn't allow to use any input modules at the same time.

bddckr:
@thestonefox This should be ready to merge again :)

broehl:
I'm trying to do something somewhat similar.

I have a virtual tablet with a world-space canvas on it, and I want to create a virtual stylus that can interact with it by tapping on the virtual screen.

I've given the canvas a VRTK_UICanvas component with "Click on Pointer Collision" set.

The stylus has a collider (with Is Trigger set) and a kinematic rigidbody. The stylus has a parent GameObject that has a VRTK_UIPointer component on it, with a Click Method of "Click on Button Down". The Controller field is set to the RightController (only because I get errors otherwise) and the pointer origin transform is set to the parent of the stylus.

The parent GameObject of the stylus also has a VRTK_Player object, with an Object Type of "Collider".

Nothing happens when the stylus touches the screen.  I've done a bit of digging, and found that the VRTK_VRInputModule's ClickOnDown() method gets called, but returns false because the list of raycast results is empty.

I checked in VRTK_UIGraphicRaycaster, and in the Raycast method it checks against all the UI elements on my virtual tablet. However, the following test:

  if (!RectTransformUtility.RectangleContainsScreenPoint(graphic.rectTransform, pointerPosition, eventCamera))

... never succeeds.

I'm clearly missing something, but I'm really scratching my head on this one.

Note that as an alternative to the stylus, I may eventually use the fingertip of my articulated virtual hand (which is why I'm not using rigid bodies attached to the controller).

I can probably work out a way to do it without using VRTK, but I don't want to reinvent the wheel if I don't need to. I suspect that most of what I need is already implemented in VRTK, and I'd rather simply use it rather than roll my own
.
Any help is much appreciated!

PR #928, opened 2017-02-16
fix(Climbing): Jitter on second hand grab climb state

mattboy64:
When climbing, grabbing the same object with your other hand
would not put the PlayerClimb physics kinematic state in the
correct value. This came from the correct state not being stored
when the player velocity modifies the object. This would cause
heavy flickering of the Oculus touch controllers as physics was
fighting with the system to get the player area in the correct
spot.


mattboy64:
Here's what the system was doing before the fix: 
flickeringmoving hands

PR #953, opened 2017-02-22
feat(Locomotion): abstract touchpad control to object control

thestonefox:
The Touchpad Control script has now been abstracted to Object Control
meaning the touchpad can be just one option of controlling an object
direction.

The Touchpad Control script now inherits the new Object Control script
and works the same way as it did before.

However, now there is a new Button Control script that creates a
pseudo X/Y axis based on button presses. This means it's possible
to move an object round based on pressing a controller button to affect
the specified axis.

The Touchpad Control events have been pushed into the Object Control
abstracted class so they can be used by other inherited classes.

All of the Touchpad Control Actions have now been renamed to
Object Control Actions as they can be used with any implementation
of an Object Control to change the object's direction.


falibu:
i was able to add directional locomotion to my controllers VERY easily using this new form of locomotion. so far everything seems to be working :D

falibu:
Step by Step for my Favorite version of Directional Locomotion:
I opened the 017 example of this build, go down to the VRTK folder in the scene assets pane, it has a left controller game object in there. click on it, and at the top of that game object's inspector panel it has a touch pad walking script on it? delete that. Now in your projects pane, go to the VRTK SDK folder and navigate to the locomotion scripts and look for the button control script. drop that on the left controller game object in the VRTK folder in the scene pane of the hierarchy window. go back to the hierarchy window and click on that left controller in your scene asset pane. it has a child called touchpadcontroloptions. if you expand that, the first thing under it shows like x-axis/y-axis slide. click on that and in the inspector panel, on both of those attached scripts, fill in the "Object Control Script" filed with the left controller. VIOLA! Directional Locomotion! do yourself a favor though and go change the settings to be "Device for Direction" to left controller of headset and initially i set the activation for forward to Grip_press. but that will depend on your game's controls. press play and go test it out! it's pretty freaking awesome! :D

PR #954, opened 2017-02-23
feat(Editor): Add warning messages for missing VRTK_Interact* scripts

dantman:
- For touch, grab, and use on the `VRTK_InteractableObject` inspector
  if a controller object is present but missing the necessary script
  a warning will be displayed.
  interactable object warning
  interactable object left warning
- If the controller alias is missing, an info message reminding to add
  the necessary script to any prefabs will be displayed.
  interactable object prefab controllers
- The `VRTK_ControllerActions` inspector on a controller for each of
  VRTK_Interact{Touch,Grab,Use} will display a warning if the script is
  missing and there are interactable objects present in the scene that require it.
  controller actions warning
- Also fix `VRTK_SDKManager` so that `instance` is not null after a
  script reload where Unity restores a serialized instance and does not
  call Awake.

### DIFF
--- a/Assets/VRTK/Editor/VRTK_ControllerActionsEditor.cs
+++ b/Assets/VRTK/Editor/VRTK_ControllerActionsEditor.cs
@@ -1,0 +1,82 @@
+﻿namespace VRTK
+{
+    using UnityEngine;
+    using UnityEditor;
+    using System;
+    using ControllerHand = SDK_BaseController.ControllerHand;
+    using AllowedController = VRTK_InteractableObject.AllowedController;
+
+    [CustomEditor(typeof(VRTK_ControllerActions), true)]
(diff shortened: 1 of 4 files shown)


bddckr:
The change to the SDK manager will be superseded by #867.

dantman:
For now the change is required to fix the SDKManager bug. (`VRTK_SDK_Bridge` already works in the editor when you load a scene or play, but the bug is that it stops working if the assembly is reloaded due to a script change)

The rest of the change will still work with the SDK discovery code as long as you make sure `VRTK_SDK_Bridge` still works in the editor and doesn't break when Unity reloads the assembly.

dantman:
Rebased for master.

thestonefox:

PR #962, opened 2017-02-25
feat(Interaction): allow sub colliders to be ignored on interactables - fixes #870

### LINKED ISSUE #870: Excluding children from Interactable Object Grab?
I have a item that is set up as a Interactable Object (a Sword) that has a few children attached, the pickup schript works on all the colliders. However i want it setup so that the player can only pickup the sword by it's hilt and not lets say the blade.

Is there currently a way to exclude certain colliders on an object from being interacted with? maybe the policy list? or setting a tag or something.

thestonefox:
A new parameter has been added to the Interactable Object script that
allows an array of colliders to be provided that will be ignored
when the controller touches the Interactable Object.

This is useful if an object has a collection of child colliders but
not all of those should react to the touch events.

Example: A sword will have a collider on the blade and hilt, but the
user will only want to be able to pick the sword up by the hilt so
the blade collider will be added to the ignore list.

If the Ignored Colliders parameter is changed at runtime then the
ResetIgnoredColliders() method also needs to be called as the
ignored controller objects get cached so the colliders are not
continually iterated through on every touch event.

### DIFF
--- a/Assets/VRTK/Scripts/Interactions/VRTK_InteractableObject.cs
+++ b/Assets/VRTK/Scripts/Interactions/VRTK_InteractableObject.cs
@@ -74,6 +74,8 @@ namespace VRTK
         public Color touchHighlightColor = Color.clear;
         [Tooltip("Determines which controller can initiate a touch action.")]
         public AllowedController allowedTouchControllers = AllowedController.Both;
+        [Tooltip("An array of colliders on the object to ignore when being touched.")]
+        public Collider[] ignoredColliders;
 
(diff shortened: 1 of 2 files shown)


davebrownsound:
@thestonefox does this fix also include highlighting? For example, I have a Door that has a DoorKnob as a child. I am trying to get only the DoorKnob object to highlight. I tried dropping the Door's collider into the Ignored Colliders array, and this does make the Door itself non-interactable, but both the Door and DoorKnob highlight upon touching the DoorKnob.

PR #969, opened 2017-02-28
fix(UIPointer): react to SteamVR index change

bddckr:
The UIPointer uses the controller index for its event data. SteamVR will
change the index sometimes, which results in the UI canvases not
reacting correctly anymore after an index change. This fix makes sure to
change the used index for the UIPointer event data whenever SteamVR
changes the indexes.


thestonefox:
This isn't the place to do this fix. I saw it come up in slack last night.

The solution is to listen to the IndexChanged event in the UIPointer and let the UIPointer handle it's own index changes.

Otherwise you have to repeat the logic for every SDK that swaps indexes.

I'm working on the controller separation stuff this week so I'll update the UIPointer to listen for the index changes and manage their own indexes correctly.

bddckr:
Understood. Will leave this PR up so this can be used when implementing it properly.

PR #978, opened 2017-03-02
fix(UICanvas): unusable in Unity 5.6

bddckr:
The UI Canvas creates a child game object to enable drag and drop of UI
elements on the canvas. That child object is a UI object so it gets
added a `RectTransform`. A `RectTransform` is inheriting from
`Transform` which means Unity does some stuff to the child object to
switch to that new transform. In Unity 5.6 this switching changed
somehow so the current implementation results in the parent of the
canvas' game object getting disabled.

The fix is to add the `RectTransform` as early as possible and only
adjust the parent and position of the transform afterwards.


GitCop:

 * Commit: 84c052269861f62aea8c8cd26f72f48cb83f8c61

---

bddckr:
@ciwolsey confirmed it works for him in 5.6b9, too.

PR #981, opened 2017-03-04
fix(SDK): SteamVR defines

bddckr:
The SteamVR defines incorrectly used a SteamVR type directly which
resulted in compiler errors if the SteamVR plugin is not included in the
project. This fix makes sure to not use any SteamVR type directly and
resorts to use reflection instead.


bddckr:
Not ready to be merged...

thestonefox:
is this ready to be merged now?

bddckr:
Yes it is! @mattboy64 confirmed it, too 😄

PR #984, opened 2017-03-05
feat(Simulator): Simulator enhancements

dantman:
* Add key bindings for WASD so movement bindings can be changed for other keyboard layouts.
* Add context sensitive control hints.
* Add an alternative scene view like mouse movement input mode where a button press is required and the mouse cursor can be locked to the game window.

simulator hmd
simulator controller
simulator lock

### DIFF
--- a/Assets/VRTK/Prefabs/VRSimulatorCameraRig.prefab
+++ b/Assets/VRTK/Prefabs/VRSimulatorCameraRig.prefab
@@ -16,11 +16,11 @@ GameObject:
   m_ObjectHideFlags: 1
   m_PrefabParentObject: {fileID: 0}
   m_PrefabInternal: {fileID: 100100000}
-  serializedVersion: 4
+  serializedVersion: 5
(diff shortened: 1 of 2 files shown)


thestonefox:
all of the commit messages need lowercase subject lines

e.g.

`feat(Simulator): allow....` is correct

`feat(Simulator): Allow....` is incorrect

dantman:
Done

dantman:
@virror

lxpk:
The lock cursor doesn't work for me or students when trying to teach VRTK. As written, it only works if you alter sim controller to use the held key for mouse movement, making sim unusable by default. Every attempt to move around or use ALT hand controls clicks out of the scene view.

I think the default should be to have locked cursor whenever simulator runs by default and it should actually work.

thestonefox:
@lxpk you need to raise an issue with steps to reproduce the issue in an example scene so someone can have a look at the issue.

PR #988, opened 2017-03-06
fix(RoomExtender_PlayAreaGizmo): null play area

### LINKED ISSUE #985: Room Extender Editor Gizmo no longer works
Open `028_CameraRig_RoomExtender`

It is supposed to show a gizmo for the play area extended size.

However, it cannot get the playArea object to know how to draw the gizmo.

@bddckr is this due to a change with the SDKManager work?

bddckr:
The play area gizmo of the RoomExtender looks up the play area in
`Awake` and later uses it when drawing the gizmo. The issue is that the
SDK Manager is not set up correctly at that point in time so the
returned play area is null, resulting in errors when trying to draw the
gizmo.

This fix makes sure `VRTK_SDKManager.instance` is always set up, no
matter when the getter is called. Additionally the play area gizmo
drawing method is changed to not throw any errors if the needed objects
weren't found in the `Awake` method. That method logs a warning already,
so the user is notified about the reason the gizmo isn't drawing.

This fix resolves #985.

### DIFF
--- a/Assets/VRTK/Scripts/Internal/VRTK_RoomExtender_PlayAreaGizmo.cs
+++ b/Assets/VRTK/Scripts/Internal/VRTK_RoomExtender_PlayAreaGizmo.cs
@@ -41,8 +41,12 @@
 
         private void DrawWireframe()
         {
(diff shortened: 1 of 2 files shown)


thestonefox:
i think i may have introduced some conflicts :(

bddckr:
Resolved.

PR #1001, opened 2017-03-09
fix(OculusAvatar): define symbol

bddckr:
The scripting define symbol for the Oculus Avatar SDK was never added.
This meant the Oculus SDK didn't work in VRTK if the Avatar SDK was
used, too. The fix is to make sure to add scripting define symbols even
if they don't are the symbols specified in the currently selected
`SDKInfo`s.


bddckr:
This change means we compile all the VRTK SDK classes based on whether a vendor SDK is installed, not based on the selected SDK. I am not sure if this is a good or bad idea.

Thoughts? That new behavior is nice because it means you don't have to let Unity recompile when changing the selected SDK in the SDK Manager, but on the other hand it means you can't easily tell VRTK to stop compiling things that use a specific SDK.
Does this introduce problems when creating builds for Oculus Home for example?

bddckr:
Looks like this change shouldn't introduce problems. @reznovVR thinks so, too.

PR #1004, opened 2017-03-10
fix(OculusSDK): prevent increasing haptics duration on Oculus Touch

robinerd:
The haptics clip used for a tick of vibration was never reset, and would
previously get longer and longer until reaching its maximum length.
Adding a reset clears the clip, without reallocating memory.
Two clips are now used instead (left / right) since clearing the clip
would otherwise prevent both controllers from vibrating simultaneously.


robinerd:
My first pull request, please be extra critical ;)

thestonefox:
❤️

PR #1006, opened 2017-03-11
fix(UICanvas): RectTransform in 5.5.2p2 and 5.6

bddckr:
Unity changed the way additions of components derived from `Transform`
are handled:
- From the 5.5.2p2 changelog:
>Kernel: Prevent scripts from adding Transform-derived components during
their Awake methods.

- From the 5.6.0b11 changelog:
>UI: Adding UI elements via script will cause OnDisable and OnEnable to
be called on all scripts in the hierarchy if a RectTransform needs to be
added.

To fix the first problem the creation of any game object that uses
`RectTransform` is deferred to the next frame. The second problem is
fixed by specifying the `RectTransform` to add in the game object
constructor.

### DIFF
--- a/Assets/VRTK/Scripts/UI/VRTK_UICanvas.cs
+++ b/Assets/VRTK/Scripts/UI/VRTK_UICanvas.cs
@@ -4,6 +4,7 @@ namespace VRTK
     using UnityEngine;
     using UnityEngine.UI;
     using UnityEngine.EventSystems;
+    using System.Collections;
 
     /// <summary>
     /// The UI Canvas is used to denote which World Canvases are interactable by a UI Pointer.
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: 6de58c474b235def69ca7fd6a2fec6e8bb6c626d

---

bddckr:
@ciwolsey confirmed this fixes usage of runtime creation of `VRTK_UICanvas`es.

Assets/VRTK/Scripts/UI/VRTK_UICanvas.cs
```diff
@@ -108,16 +109,17 @@ protected virtual void SetupCanvas()
                 canvasRigidBody.isKinematic = true;
             }
 
-            CreateDraggablePanel(canvas, canvasSize);
+            StartCoroutine(CreateDraggablePanel(canvas, canvasSize));
```
thestonefox: As this `StartCoroutine` is called from an `OnEnable` you'll need to make a reference to the coroutine and call `StopCoroutine` in the `OnDisable` method

e.g.

```
private Coroutine draggablePanelCreation;
...

draggablePanelCreation = StartCoroutine(CreateDraggablePanel(canvas, canvasSize));

....

protected virtual void OnDisable()
{
  StopCoroutine(createDraggablePanel);
}
```

Otherwise, if the panel is enabled and disabled before the next frame happens the coroutine will still attempt to run.

bddckr: Good catch!

PR #1008, opened 2017-03-11
fix(Simulator): render model null

bddckr:
The simulator would throw a `NullReferenceException` in
`SDK_SimController.GetControllerRenderModel` because the alias
controller wasn't parented to the actual one yet. This fix makes sure to
parent the script alias controller to the actual object immediately.


bddckr:
Would be great if someone could test if this still works with the other SDKs (SteamVR, Oculus etc.).

PR #1014, opened 2017-03-12
fix(ObjectFollow): runtime usage

bddckr:
The Object Follow scripts cache some needed components in their
`OnEnable` methods. This meant that using these scripts at runtime
needed a toggle of their enabled state. The fix is to cache the
components even if the needed setup isn't done as soon as `OnEnable` is
called.

### DIFF
--- a/Assets/VRTK/Scripts/Utilities/ObjectFollow/VRTK_RigidbodyFollow.cs
+++ b/Assets/VRTK/Scripts/Utilities/ObjectFollow/VRTK_RigidbodyFollow.cs
@@ -27,23 +27,25 @@ namespace VRTK
         protected Rigidbody rigidbodyToFollow;
         protected Rigidbody rigidbodyToChange;
 
-        protected override void OnEnable()
+        protected virtual void OnEnable()
         {
-            base.OnEnable();
-
-            if (gameObjectToFollow == null)
-            {
-                return;
-            }
-
-            rigidbodyToFollow = gameObjectToFollow.GetComponent<Rigidbody>();
-            rigidbodyToChange = gameObjectToChange.GetComponent<Rigidbody>();
+            CacheRigidbodies();
         }
 
         protected virtual void OnDisable()
         {
             rigidbodyToFollow = null;
             rigidbodyToChange = null;
+        }
+
+        protected override void Update()
+        {
+            base.Update();
+
+            if (rigidbodyToFollow == null || rigidbodyToChange == null)
+            {
+                CacheRigidbodies();
+            }
         }
 
         protected virtual void FixedUpdate()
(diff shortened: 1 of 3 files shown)


Assets/VRTK/Scripts/Utilities/ObjectFollow/VRTK_RigidbodyFollow.cs
```diff
@@ -27,17 +27,9 @@ public enum MovementOption
         protected Rigidbody rigidbodyToFollow;
         protected Rigidbody rigidbodyToChange;
 
-        protected override void OnEnable()
+        protected void OnEnable()
```
thestonefox: needs to be `protected virtual` to allow overriding if required.

Assets/VRTK/Scripts/Utilities/ObjectFollow/VRTK_RigidbodyFollow.cs
```diff
@@ -99,5 +101,16 @@ protected override Vector3 GetScaleToFollow()
         {
             return rigidbodyToFollow.transform.localScale;
         }
+
+        private void CacheRigidbodies()
```
thestonefox: Everything should now be at least `protected virtual`

I should probably update the contrib guidelines with that!

Assets/VRTK/Scripts/Utilities/ObjectFollow/VRTK_TransformFollow.cs
```diff
@@ -22,45 +22,66 @@ public enum FollowMoment
         }
 
         [Tooltip("The moment at which to follow.")]
-        public FollowMoment moment = FollowMoment.OnPreRender;
+        [SerializeField]
+        private FollowMoment _moment = FollowMoment.OnPreRender;
```
thestonefox: This should probably be protected too in case of wanting to inherit and modify the script

Assets/VRTK/Scripts/Utilities/ObjectFollow/VRTK_TransformFollow.cs
```diff
@@ -22,45 +22,66 @@ public enum FollowMoment
...
+                _moment = value;
+            }
+        }
 
         protected Transform transformToFollow;
         protected Transform transformToChange;
-        protected bool isListeningToOnPreRender;
 
-        protected override void OnEnable()
+        protected void OnEnable()
```
thestonefox: `protected virtual`

Assets/VRTK/Scripts/Utilities/ObjectFollow/VRTK_TransformFollow.cs
```diff
@@ -103,5 +124,16 @@ protected override void SetRotationOnGameObject(Quaternion newRotation)
         {
             transformToChange.rotation = newRotation;
         }
+
+        private void CacheTransforms()
```
thestonefox: `protected virtual`

Assets/VRTK/Scripts/Utilities/ObjectFollow/VRTK_TransformFollow.cs
bddckr: I disagree here: I think subclasses should use `moment` instead of `_moment` as the latter is just the backing field. Changing that one without going through `moment`'s setter would result in not correctly subscribing to/unsubscribing from `Camera.onPreRender`.

bddckr:
Updated 😄

PR #1015, opened 2017-03-12
perf(UpdatePrompt): only check for updates every 6 hours

bddckr:
The update prompt was checking for new updates to VRTK every time Unity
was started or recompiled the project. This change makes sure to only
check every 6 hours.


bddckr:
Rebased on latest master and resolved conflicts.

PR #1043, opened 2017-03-20
fix(sdk): add null reference check

gpvigano:
Add a null reference check also for actualRightController in
VRTK_SDKManager::SetupControllers().


bddckr:
Why is this needed? Does the GearVR have no controller at all?

bddckr:
Disregard, just saw this in your PR:
>it does not support motion or pointer based controllers
😅

PR #1067, opened 2017-03-30
fix(SDK): caching SDK Manager objects methods

bddckr:
The SDK bases Boundaries and Headset provided methods to their
subclasses that made it easier to get the SDK Manager's play area and
headset, respectively. The methods themselves and their usage didn't
result in the intended caching of the returned objects. Since the
subclasses do the caching properly already the methods and their usage
is removed in this fix.


bddckr:
After a bit of discussion with @thestonefox I realized that I only did this because I saw the need in the SteamVR Boundaries SDK. There's a totally fine use case for the superclass' methods: Whenever the SDK implementation doesn't need a special script on the game object.

PR #1073, opened 2017-03-30
refactor(UnityEvents): simplify by using new abstract superclass

bddckr:
A new abstract superclass `VRTK_UnityEvents` has been added to simplify
and standardize the various existing UnityEvents scripts. Additionally
all the xml docs have been removed from the UnityEvents scripts because
they add nothing to the documentation (the documentation already refers
to have a look at the respective scripts' documentation to learn more
about the events).


thestonefox:
Previously, the Unity Event Helpers had the RequireComponent to add the component if needed.

From what i can tell it doesn't look like this PR will do that anymore and it will just throw an error? Is that correct?

bddckr:
Correct.

PR #1074, opened 2017-03-30
fix(SDK): find inactive game objects and components

bddckr:
The SDK classes need to search for some game objects or components. This
fix makes sure that the search for active as well as inactive game
objects and components always works.


thestonefox:
Will this cause an issue if someone has lets say the SteamVR CameraRig in their scene but it's disabled?

Previously it wouldn't find those objects because they were disabled, but now even with those disabled we can set things up still. I'm "guessing" you need this for the future SDK PR?

bddckr:
Yeah I need the camera rigs be disabled before starting the game to support runtime switching correctly. SDKs like SteamVR otherwise just disable *some* of their scripts themselves if they are enabled when they can't be set up correctly. 😞

I figured you wouldn't have multiple camera rigs in the scene anyway at which point being able to auto-populate is a feature, not an issue. 😄

PR #1077, opened 2017-03-30
feat(SnapDropZone): add custom material property for zone highlighter

WildStyle69:
There is now a an optional property for adding a custom
material to the snap zone. If present this material will
be passed to the zones highlighter component.


thestonefox:
This is how you achieve a custom highlight material: 

You just add the highlighter manually and use the Custom Material option on the highlighter:

image

PR #1114, opened 2017-04-13
fix(Pointers): do not raise the DestinationMarkerEnter event every frame

DrummerB:
The DestinationMarker**Exit** event is raised _once_ when the pointer detects no more hits or if a different collider is hit.

The DestinationMarker**Enter** event on the other hand is raised every frame while there is a hit detected (see the flooded console in example scene 003). I think it should only be called once for the initial hit of a collider.

This PR fixes this and makes the Pointer raise these events in pairs. Maybe a different DestinationMarkerPoint event could be raised every frame instead.

### DIFF
--- a/Assets/VRTK/Scripts/Pointers/PointerRenderers/VRTK_StraightPointerRenderer.cs
+++ b/Assets/VRTK/Scripts/Pointers/PointerRenderers/VRTK_StraightPointerRenderer.cs
@@ -169,9 +169,9 @@ namespace VRTK
                 if (pointerCollidedWith.collider && pointerCollidedWith.collider != destinationHit.collider)
                 {
                     PointerEnter(pointerCollidedWith);
-                    ChangeColor(validCollisionColor);
                 }
                 destinationHit = pointerCollidedWith;
+                ChangeColor(validCollisionColor);
             }
         }
 
--- a/Assets/VRTK/Scripts/Pointers/PointerRenderers/VRTK_StraightPointerRenderer.cs
+++ b/Assets/VRTK/Scripts/Pointers/PointerRenderers/VRTK_StraightPointerRenderer.cs
@@ -166,10 +166,12 @@ namespace VRTK
         {
             if (rayHit)
             {
-                PointerEnter(pointerCollidedWith);
-
+                if (pointerCollidedWith.collider && pointerCollidedWith.collider != destinationHit.collider)
+                {
+                    PointerEnter(pointerCollidedWith);
+                    ChangeColor(validCollisionColor);
+                }
                 destinationHit = pointerCollidedWith;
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 6ae13fafbde50cadbdc6980fea84b9678047d790

---

thestonefox:
Hmm, Doesn't it give you a different point on the object you're hitting every frame?

So if you want to know the different parts of a collider being hit then this PR would prevent that from happening?

Perhaps it should be optional to only trigger a new event if the collider changes?

DrummerB:
Yes, it does give different hit locations and my PR prevents that. 

I think the semantics of Enter/Exit imply that these events are only raised once and in pairs. That's how Selectable.OnPointerEnter/Exit or OnEnable/OnDisable behave too. 

It's probably not a rare use case to want to do something when you point at an object e.g. spawn some UI next to it or play a sound. With the current implementation you need to maintain an isFirstEnter flag on the objects that you reset in DestinationMarkerExit:

    bool isFirstEnter = true;
    public void OnPointerEnter(VRTK_Pointer pointer, DestinationMarkerEventArgs e)
    {
        if (isFirstEnter)
        {
            // play sound
        }
        isFirstEnter = false;
    }
    public void OnPointerExit(VRTK_Pointer pointer, DestinationMarkerEventArgs e)
    {
        isFirstEnter = true;
    }

Instead of:

    public void OnPointerEnter(VRTK_Pointer pointer, DestinationMarkerEventArgs e)
    {
            // play sound
    }

When you're working with mouse input (not in Unity, but generally) you almost always have mouseEnter, mouseMove and mouseExit methods. Maybe something like this could work for VRTK_Pointer too?

thestonefox:
Ok so maybe you get a Pointer Enter, Pointer Exit and Pointer Hover event?

The Hover event would give you the points on the existing collider?

DrummerB:
Something like that, yes. Would this be considered a major change? It may break existing scripts.

There is a small related issue concerning multiple child colliders on a parent rigidbody. Most of the time you would probably have your scripts on the parent game object, e.g. a simple drawer with a rigidbody and behaviour scripts on the parent and 5 children with box colliders. A script on the parent is probably only concerned about whether you're pointing at *any* part of it and doesn't care if you moved the pointer from one of its colliders to an other.

I'm not sure what the most elegant solution would be or even if this is something VRTK should solve. It's quite simple to solve this by subclassing the pointer renderer or by routing the Enter/Exit events through a manager that filters these events.

thestonefox:
I'd still say fire on collider change and leave it up to the consuming script to care if the main object has changed. maybe provide an additional flag in the event?

thestonefox:
PR to supersede this one: https://github.com/thestonefox/VRTK/pull/1124

PR #1123, opened 2017-04-16
fix(SnapDropZone): allow multiple valid objects to hover at same time - fixes #1122

### LINKED ISSUE #1122: SnapDropZones and multiple objects
### Environment

 * GitHub
 * f2869bed8726079914ccdc1399f624b4d27e4f3f
 * Unity not relevant
 * Any hardware
 * All SDKs

### Steps to reproduce

In any example that uses object snapping grab a snappable object in each hand and hold them both in the snap drop zone. Let go of the object that was moved into the snap drop zone *last*.

### Expected behavior

The object that you let go of first (whichever that is) should snap.

### Current behavior

Only the object that was first placed in the snap drop zone will actually be able to snap. Letting go of the second object will either let it fall to the ground or it will stay in your hand (depending on the object's drop settings).

thestonefox:
The Snap Drop Zone can now have multiple valid snappable objects in
the hover zone and any object that is released in the snap drop zone
will be snapped.

This fixes an issue where previously only the first valid object
that entered the snap drop zone hover space would be able to be
snapped and any other object even if it was valid would not be snapped
and would just fall on drop.

### DIFF
--- a/Assets/VRTK/Prefabs/Resources/Scripts/VRTK_SnapDropZone.cs
+++ b/Assets/VRTK/Prefabs/Resources/Scripts/VRTK_SnapDropZone.cs
@@ -3,6 +3,7 @@ namespace VRTK
 {
     using UnityEngine;
     using System.Collections;
+    using System.Collections.Generic;
     using Highlighters;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


quintesse:
That was fast, you beat me to it :)

thestonefox:
@quintesse if you get chance, take a look at the PR and lemme know if you think it's ok to merge

quintesse:
@thestonefox it looks good to me. I just had one doubt, but I'm pretty sure it's because I don't know how things work exactly. I see that the highlight is turned off when an object leaves the snap zone, Is that because the highlight is done on the specific collider between two objects? So if two objects are in the drop zone each of them has a highlight?

thestonefox:
@quintesse the highlighter is for the snap drop zone, so when one valid object enters it will turn on the snap drop zone highlight, then when another object enters it will turn on the highlight again (which won't do anything as it's already on). Then if you remove an object it will turn the highlight off, but because the other object is still in there hover area, the trigger stay will force the highlight back on so the highlight won't ever go off.

PR #1128, opened 2017-04-19
fix(Climbing): reset body collision and false fall end triggers

mattboy64:
Climbing would mess with the enableBodyColliders parameter in
the body physics object. If a player set that to off, it would
override that value and never set it back. This change fixes
that problem. In addition, falling would sometimes get triggered
as being complete (velocity based) too early based on race cases
between when the FixedUpdate and fall was called. The system now
waits enough time (3 fixed updates worth) before declaring
falling to be complete.

In addition, the body physics was never reset regrabbing (with
the second controller) which put the physics system in a weird
state of having collision enabled and thinking it was in a
falling state, when it wasn't.

### DIFF
--- a/Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
+++ b/Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
@@ -169,6 +169,8 @@ namespace VRTK
         protected string footColliderContainerNameCheck;
         protected const string BODY_COLLIDER_CONTAINER_NAME = "BodyColliderContainer";
         protected const string FOOT_COLLIDER_CONTAINER_NAME = "FootColliderContainer";
+        protected bool enableBodyCollisionsStartingValue;
+        protected float fallMinTime;
 
         // Draws a sphere for current standing position and a sphere for current headset position.
         // Set to `true` to view the debug spheres.
@@ -286,12 +288,21 @@ namespace VRTK
             bodyRigidbody.angularVelocity = Vector3.zero;
         }
 
+        /// <summary>
+        /// The ResetFalling method force stops any falling states and conditions that might be set on this object.
+        /// </summary>
+        public virtual void ResetFalling()
+        {
+            StopFall();
+        }
+
         protected override void OnEnable()
         {
             base.OnEnable();
             SetupPlayArea();
             SetupHeadset();
             footColliderContainerNameCheck = VRTK_SharedMethods.GenerateVRTKObjectName(true, FOOT_COLLIDER_CONTAINER_NAME);
+            enableBodyCollisionsStartingValue = enableBodyCollisions;
             EnableDropToFloor();
             EnableBodyPhysics();
         }
@@ -529,7 +540,7 @@ namespace VRTK
 
         protected virtual void CheckFalling()
         {
-            if (isFalling && VRTK_SharedMethods.RoundFloat(lastPlayAreaPosition.y, fallCheckPrecision) == VRTK_SharedMethods.RoundFloat(playArea.position.y, fallCheckPrecision))
+            if (isFalling && fallMinTime < Time.time && VRTK_SharedMethods.RoundFloat(lastPlayAreaPosition.y, fallCheckPrecision) == VRTK_SharedMethods.RoundFloat(playArea.position.y, fallCheckPrecision))
             {
                 StopFall();
             }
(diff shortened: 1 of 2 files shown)


Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
```diff
@@ -169,6 +169,8 @@ public enum FallingRestrictors
         protected string footColliderContainerNameCheck;
         protected const string BODY_COLLIDER_CONTAINER_NAME = "BodyColliderContainer";
         protected const string FOOT_COLLIDER_CONTAINER_NAME = "FootColliderContainer";
+        private bool enableBodyCollisionsStartingValue;
```
thestonefox: Make these protected. All VRTK stuff should be at least protected now to allow people to extend scripts.

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
```diff
@@ -169,6 +169,8 @@ public enum FallingRestrictors
...
+        private float fallMinTime;
```
thestonefox: Protected here too

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
```diff
@@ -1035,13 +1038,15 @@ protected virtual void StartFall(GameObject targetFloor)
             isMoving = false;
             isLeaning = false;
             onGround = false;
+            fallMinTime = Time.time + (Time.fixedDeltaTime * 3.0f); // Wait at least 3 fixed update frames before declaring falling finished
```
thestonefox: Why the 3 frames?

3 feels a bit like a magic number. I'm guessing this is caused by a timing issue and 3 happens to work? Wonder if there's a better way of solving the timing issue. @bddckr any ideas?

mattboy64: Just want to be sure physics has enough time to settle down or add gravity and such. It's a buffer zone really, so any amount of time would work. What I really need is the ability to know if physics has been put to sleep. The velocity has always been a bit of a hack.

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
mattboy64: Ah, I gathered that from the virtuals, but didn't notice the protecteds

mattboy64:
Code changes are in.

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
bddckr: Maybe `Rigidbody.sleepThreshold/IsSleeping`?

mattboy64: I think whenever you mess with the collider it wakes up. We are constantly changing it's shape and position so I don't think that works. I played with it last year and never came to a good solution.

mattboy64:
Just checked sleeping, it doesn't sleep when the player is moving so I don't think that would be an exact way to test. Since timing isn't precise (even with fixedUpdate timing), you have to give a buffer to even be sure it goes through a physics update. The only other alternative that I can think of is doing a countdown after x FixedUpdates which seems more invasive. Would you rather I go that route @thestonefox?

thestonefox:
are we going to run into a situation where sometimes this wait `3` ticks trips us up because in a certain circumstance it needs to wait 4, or 2 is too many?

I know from bitter experience that numbers that work, end up not working :(

I guess in the short term (if everyone is happy) we can go with this approach and solve the other issue later if it arises.

mattboy64:
Well before this change, it was completely broke anyway probably 30% of the time. It's an improvement if nothing else. Really I just want to let the physics cycle through a time or two to make sure any forces are being applied before setting the fall as complete.

mattboy64:
Added another important state, when grabbing with a second controller, the body physics would go into a falling state (from the first controller ungrab), but never back into the correct grabbing state. ResetFall is now called when grabbing an object to make sure the state goes back to the way it's supposed to be. This might be at least part of the problem with the physics freak out seen when climbing @thestonefox.

thestonefox:
@mattboy64  Shall we merge this? Or do you think there will be additional tweaks as you're going through your stuff?

mattboy64:
I'd say lets go ahead and merge it. I don't anticipate any other tweaks right now.

PR #1129, opened 2017-04-19
feat(ControllerReference): add new way of referencing controllers - resolves #1118

### LINKED ISSUE #1118: Allow all static controller methods to use an index or a given controller game object
At the moment there is a mish-mash of ways of calling static methods for a specific controller

Some methods require the controller index, whilst others can simply use the actual or script alias game object.

It's a good idea for any method that uses index to also have a game object equivalent and any one that uses a game object to have an index equivalent.

Then it's developer preference on how it works.

thestonefox:
Previously, controller calls were referenced in a number of ways from
utilising the controller's id, to passing a game object of either the
actual controller game object or the script alias game object.

This meant that ways to call certain methods was fragmented and lead
to confusing different options between methods.

The new ControllerReference script is used to get a reference to
a controller based on any number of things such as an index, the
actual GameObject, the script alias Object, the model Object or
even the controller hand.

Once a reference is retrieved, it can then access all of the elements
of a controller, such as:

 * index
 * actual
 * scriptAlias
 * model
 * hand

The ControllerReference script also has a collection of static methods
that are preferably used rather than needing to instantiate a new
object as the static methods do some nice checking to make sure
the ControllerReference is valid.

The Controller References are also cached for performance and
different instances of a Controller Reference object can be compared
as a custom comparator has been provided that compares on the index
of the controller which would always be universally unique.

A large collection of methods in the SDK implementations and SDK Bridge
have been re-written to utilise the Controller Reference, whilst a
collection of methods in the SDK Bridge have been deprecated and the
relevant methods in the SDK implementations have been removed.

All of the event payloads that passed a controller index around now
pass a controller reference, with the index being deprecated.


gpvigano:
By mistake I previously rebased my PR #1132 (support for gear vr) on the commit in this PR (yes, maybe I better stop making a mess with Git and go on holiday...), anyway I discovered some problems and I tried to address them, if you want you can check my PR #1142 (or you can just close it if not interested).

thestonefox:
@gpvigano I've updated this PR now to fix a bunch of the things you'd highlighted. I closed your other PR as this PR now supersedes it.

PR #1136, opened 2017-04-25
fix(Locomotion): emit dash teleport events at the correct time

thestonefox:
Previously, the OnTeleporting and OnTeleported event were being emitted
at the same time in the Dash Teleporter, because the events are both
emitted when the `DoTeleport` method is called. However, as the dash
uses a co-routine to set the position, the destination location can
take a certain amount of time be reached so the OnTeleported event
needed to be emitted after that co-routine had finished.

To fix this, a collection of new overridable methods have been added
that break up the functionality of the `DoTeleport` method and allow
the Dash Teleport script to omit the OnTeleported event and then
manually emit it when required in the co-routine.


CraigASanford:
Tested and working correctly! You rock sir.

PR #1151, opened 2017-05-01
fix(DeviceFinder): headset type recognition

### LINKED ISSUE #1113: Controller Tooltip Lines Point to the Wrong Place
### Environment

 * Github build (from 1-2 days ago)
 * Unity 5.6.0f3
 * Vive
 * SteamVR SDK, latest

### Steps to reproduce

Open the Controller Tooltips example. Either use it with your Vive, or force model override for Vive Controller 1.5 with Tooltips.

### Expected behavior

Tooltips should point to the correct locations on the Vive wand.

### Current behavior

Only the trigger seems to be accurate. All other tooltip lines point to the wrong place -- probably the controllers origin. The incorrect tooltip lines may not be finding the components correctly...

Here is a screenshot of the current behavior:

http://i.imgur.com/KOiSG5H.png

bddckr:
The headset type recognition has a fixed list of known headsets.
This list is missing a special Vive Pre model that is reported as
"Vive DVT".
This fix adds that headset and modifies the recognition to also tell
the user to report any new headset type so it can be added to VRTK.
Additionally a slower check is done when a summarization is
requested to support types VRTK doesn't know about yet.

This fixes #1113.

### DIFF
--- a/Assets/VRTK/Scripts/Utilities/VRTK_DeviceFinder.cs
+++ b/Assets/VRTK/Scripts/Utilities/VRTK_DeviceFinder.cs
@@ -36,7 +36,8 @@ namespace VRTK
             OculusRift,
             OculusRiftCV1,
             Vive,
-            ViveMV
+            ViveMV,
+            ViveDVT
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


thestonefox:
seems legit to me

PR #1172, opened 2017-05-09
fix(HipTracking): move to LateUpdate

virror:
Move from Update to LateUpdate to remove stutter.

### DIFF
--- a/Assets/VRTK/Scripts/Locomotion/ObjectControlActions/VRTK_BaseObjectControlAction.cs
+++ b/Assets/VRTK/Scripts/Locomotion/ObjectControlActions/VRTK_BaseObjectControlAction.cs
@@ -152,5 +152,17 @@ namespace VRTK
 
             return axisDirection;
         }
+
+        protected virtual bool CanMove(VRTK_BodyPhysics givenBodyPhysics, Vector3 currentPosition, Vector3 proposedPosition)
+        {
+            if (givenBodyPhysics == null)
+            {
+                return true;
+            }
+
+            Vector3 proposedDirection = (proposedPosition - currentPosition).normalized;
+            float distance = Vector3.Distance(currentPosition, proposedPosition);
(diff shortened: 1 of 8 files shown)


thestonefox:
Why is there stuttering in Update?

May be worth updating the commit message with more of a background as to why this is considered a valid fix?

virror:
This is "just how you should do" tracking, its common knowledge that moving things to follow other moving things in Update usually ends up causing stutter, there are tons of threads about it. When i tried to use this first i immediately noticed quite a lot of stutter and this got rid of all of it.

thestonefox:
Can you link to some of the threads?

virror:
Just try it for your self and you will see that it stutters with the current implementation? : )
Isnt the issue kind of similar to whats described in the Unity docs for LateUpdate: https://docs.unity3d.com/ScriptReference/MonoBehaviour.LateUpdate.html

virror:
Note that this is only tested with the simulator, might not be an issue with the vive, i kind of forgot about that until now : D

thestonefox:
Hah, I mean update the git commit with reasons and links if you have them so people get a better idea when looking through the history :)

virror:
I got that, just wanted your confirmation : D
It sounded a bit like you did not like the pr : p
And now there are stupid conflicts as well, lol

virror:
Updated the commit message, also fixed the conflict but for some reason its still not fixed? No idea whats up with that.

bddckr:
@virror You'll need to squash the commits to only leave the one with the nice commit message.
Maybe @thestonefox can do this since he gets PR access if you have the checkbox on the right side on this page enabled.

virror:
I used the github way of resolving conflicts, thought that was ok since it's only a merge, not a change to the actual product content.

bddckr:
The problem is that that commit shows up as soon it's merged into the master branch.

virror:
Yeah, squash merge should be an option for a pr.

PR #1176, opened 2017-05-10
feat(Utilities): add sdk object alias script for following SDK objects

thestonefox:
The SDK Object Alias script will child the script's GameObject to the
selected SDK Object.

This is useful if a GameObject needs to follow the SDK headset or
play area as these can change during runtime if the SDK is switched.

All of the example scenes that used the Transform Follow have been
updated now to use this new alias script to know which object to
follow as the SDK Object is likely to change when switching SDK.

### DIFF
--- a/Assets/VRTK/Examples/023_Controller_ChildOfControllerOnGrab.unity
+++ b/Assets/VRTK/Examples/023_Controller_ChildOfControllerOnGrab.unity
@@ -37,7 +37,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
   m_CustomReflection: {fileID: 0}
   m_Sun: {fileID: 0}
-  m_IndirectSpecularColor: {r: 0.18053603, g: 0.2260181, b: 0.30718738, a: 1}
+  m_IndirectSpecularColor: {r: 0.18028352, g: 0.22571376, b: 0.30692244, a: 1}
 --- !u!157 &3
 LightmapSettings:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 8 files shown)


Assets/VRTK/Scripts/Utilities/VRTK_PseudoSDKObject.cs
```diff
@@ -0,0 +1,81 @@
...
+    public class VRTK_PseudoSDKObject : MonoBehaviour
+    {
+        /// <summary>
+        /// Valid SDK Objects
+        /// </summary>
+        /// <param name="Boundary">The main camera rig/play area object that defines the player boundary.</param>
+        /// <param name="Headset">The main headset camera defines the player head.</param>
+        /// <param name="LeftController">The main object that defines the left controller.</param>
+        /// <param name="RightController">The main object that defines the right controller.</param>
+        public enum SDKObjects
```
bddckr: This enum should probably be named `SDKObject`, since this is not usable as a bitfield (which would offer using _multiple_ ones).

Assets/VRTK/Scripts/Utilities/VRTK_PseudoSDKObject.cs
```diff
@@ -0,0 +1,81 @@
...
+        protected virtual void ChildToSDKObject()
+        {
+            Vector3 currentPosition = transform.localPosition;
+            Quaternion currentRotation = transform.localRotation;
+            Vector3 currentScale = transform.localScale;
+
+            switch (sdkObject)
+            {
+                case SDKObjects.Boundary:
+                    transform.SetParent(VRTK_DeviceFinder.PlayAreaTransform());
```
bddckr: This may be cleaner by only setting a local `Transform parent` in the `switch` and then doing the `transform.SetParent(parent)` after the `switch`. (Just to save some typing and remove duplicated code.)

PR #1178, opened 2017-05-11
fix(setupWindow): increase window height

virror:
Increase window height so everything fits inside.


bddckr:
The real solution is probably to put that whole window's content into a scroll pane so it's actually scrollable, since we don't know the user's screen resolution!

virror:
User screen resolution should not matter since it's all in pixels. Problem was that someone added content without adjusting the window.

PR #1182, opened 2017-05-11
feat(Interaction): move methods from SharedMethods to specific scripts - resolves #1161

### LINKED ISSUE #1161: Discussion: Structure of controller related functions
First of all this is just my personal thoughts about some of the recent additions and changes, feel free to just ignore it if you think its stupid : )

I think its a bit messy now how you access some of the controller related APIs. Some things are on the controller scripts (use/grab etc), some are on the device finder (velocities) and some in some obscure shared methods class (haptics). I think there should be a common way of getting all APIs related to a controller, to make it easier for the user to figure it out.

Something like:
-Get the controller reference using DeviceFinder or some other means.
-Get the relevant script
-Call the relevant function

And this should be valid for everything that are related to a specific controller, like grab, use, haptics, velocity and whatever other API calls. At least i suggest that haptics stuff gets moved to the haptics script, and also that the velocity and angular velocity gets moved from the DeviceFinder and onto some of the controller scripts.

thestonefox:
The SharedMethods script had a collection of static calls to things
like controller haptics and to object appearances, which relied upon
Instance Method scripts being spun up in the background.

This is a long winded approach and has been replaced now by a specific
script to deal with the static haptic methods and a specific script
to deal with the static object appearance methods.

The respective new scripts will also spin up an instance behind the
scenes to handle any required coroutines rather than having some
overall script manage these instances.

Once the script instance is created, it doesn't need to be created
again.


virror:
Any specific reason for yet another haptics file instead of just using the VRTK_Haptics one? Now there are 3 different files handling haptics.

virror:
Looking good : )

PR #1194, opened 2017-05-13
fix(ScreenFade): ensure alpha does not get locked at 1f - fixes #1186

### LINKED ISSUE #1186: Teleporter stays faded in black on Oculus Utilities
### Environment

 * Source of VRTK: Github
 * Version of VRTK:  d4445f2505fe169e0454cbe2a11d56511efb7c9b
 * Version of Unity3D: 5.5.3f1
 * Hardware used: Oculus
 * SDK used: Oculus Utilities 1.12

### Steps to reproduce

 * Load Example Scene 004
 * Run scene and attempt to teleport

### Expected behavior

Should teleport to new location with fade in/out on headset

### Current behavior

Headset fades to black but never fades back to transparent

thestonefox:
There was a bug where the alpha would not get set fully to 1f so
a fix was introduced to force it to 1f if the alpha was over 0.98f.

However, this created another bug where when the alpha should go from
1f to lower it would keep getting reset to 1f.

This fix checks to see if the alpha is being changed up towards the
1f value before attempting to force the alpha to 1f above 0.98f.

### DIFF
--- a/Assets/VRTK/Scripts/Internal/VRTK_ScreenFade.cs
+++ b/Assets/VRTK/Scripts/Internal/VRTK_ScreenFade.cs
@@ -55,7 +55,7 @@
 
             if (currentColor.a > 0 && fadeMaterial)
             {
-                currentColor.a = (currentColor.a > 0.98f ? 1f : currentColor.a);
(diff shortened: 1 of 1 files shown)


quintesse:
Works for me!

PR #1195, opened 2017-05-13
feat(Tooltips): add option to make text always face the headset

thestonefox:
A new option on the Tooltip has been added that makes the tooltip
text always face the headset by rotating it accordingly.

### DIFF
--- a/Assets/VRTK/Prefabs/Resources/Scripts/VRTK_ObjectTooltip.cs
+++ b/Assets/VRTK/Prefabs/Resources/Scripts/VRTK_ObjectTooltip.cs
@@ -33,13 +33,16 @@ namespace VRTK
         public Color containerColor = Color.black;
         [Tooltip("The colour to use for the line drawn between the tooltip and the destination transform.")]
         public Color lineColor = Color.black;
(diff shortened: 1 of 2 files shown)


Assets/VRTK/Prefabs/Resources/Scripts/VRTK_ObjectTooltip.cs
```diff
@@ -55,40 +58,45 @@ public void ResetTooltip()
...
         {
             displayText = newText;
             ResetTooltip();
         }
 
-        protected virtual void Start()
+        protected virtual void OnEnable()
         {
             ResetTooltip();
+            headset = VRTK_DeviceFinder.HeadsetTransform();
```
bddckr: Since this script now depends on the current SDK Setup's used headset this script needs to be added to the SDK Manager behaviour list to disable & enable on SDK Setup changes:
`AddBehaviourToToggleOnLoadedSetupChange` & `RemoveBehaviourToToggleOnLoadedSetupChange` on `VRTK_SDKManager.instance`. I recommend to add in `Awake` and remove in `OnDestroy`.

PR #1200, opened 2017-05-14
feat(Controls): add `Released` event

Backer-Sultan:
Add `Released` event to VRTK_Button.cs
Add `OnReleased` Button3DEvent to VRTK_Button_UnityEvents.cs

### DIFF
--- a/Assets/VRTK/Scripts/Controls/3D/VRTK_Button.cs
+++ b/Assets/VRTK/Scripts/Controls/3D/VRTK_Button.cs
@@ -67,9 +67,14 @@ namespace VRTK
         public ButtonEvents events;
 
         /// <summary>
-        /// Emitted when the 3D Button has reached it's activation distance.
+        /// Emitted when the 3D Button has reached its activation distance.
         /// </summary>
         public event Button3DEventHandler Pushed;
+
+        /// <summary>
+        /// Emitted when the 3D Button's position has become less than activation distance after being pressed.
+        /// </summary>
+        public event Button3DEventHandler Released;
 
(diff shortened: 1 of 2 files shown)


Assets/VRTK/Scripts/Controls/3D/VRTK_Button.cs
```diff
@@ -464,6 +482,11 @@ protected virtual bool ReachedActivationDistance()
             return (Vector3.Distance(transform.position, restingPosition) >= activationDistance);
         }
 
+        protected virtual bool ReachedDeactivationDistance()
+        {
+            return (Vector3.Distance(transform.position, restingPosition) <= Vector3.kEpsilon);
```
bddckr: You can just use `Vector3.==` here since it [returns true if two vectors are approximately equal.](https://docs.unity3d.com/ScriptReference/Vector3-operator_eq.html).

Backer-Sultan: This line always returns false because `transform.position` is modified by physics when hitting Play.
It's also more realistic to trigger `Released` event when Button's position becomes less than activation distance instead of when reaching resting position.
The code is modified and tested. it's working well.

virror:
You will have to squash those to changes ot @thestonefox will refuse to merge : p

Backer-Sultan:
Sure I will :)

Backer-Sultan:
@bddckr I've squashed the two commits I made. That overrides the old review (and comments related to it). Should I close the pull request and reopen it, or it's ok to have it like this?

bddckr:
@Backer-Sultan You're fine! This is exactly how you should do it 😄

PR #1201, opened 2017-05-14
feat(physicsAudio): add physics based audio playing.

virror:
This feature will add the ability for IOs and other objects to play sounds when hitting and sliding on other objects using physics. This works by assigning "AudioMaterials" to IOs, terrain and other parts of the scene that you want associated with a certain material. The system will also mix sounds if they are from different materials like stone + wood.


GitCop:

 * Commit: 8f7a764982ccd254da56fc379c365ebc3ecd7756

---

virror:
@thestonefox @bddckr 
This is still experimental, but i will put it here so you guys can test it and see what you think. And its also a good way of structuring any discussion around the feature i think.

SpencerMalone:
Oooo, fascinating idea. Along with this, you can get a really good "woosh" sound (like from a pipe, bat, sword, etc.) by taking some brown / pink / white noise and modulating the volume + pitch based on angular velocity and velocity. I'm still playing around with it, but it might be another thing to include alongside this.

AntonShumin:
So its something like this https://www.assetstore.unity3d.com/en/#!/content/37650 plus dynamic wooshes on velocity?

virror:
Yeah, its pretty much like that package, but with one (imo) big difference, and thats surface dependent audio mixing. That package plays only one defined sound depending on material, while this one mixes audioclips on the fly depending on the held object and the surface thats hit.
There are no "wooshes" implemented at this time, but its a cool idea for sure.

thestonefox:

PR #1207, opened 2017-05-17
fix(Simulator): sdd simple neck model.

virror:
Fixed a bug with y rotation when rig was parented that was
caused by mixing of rotation and localRotation. Also add
a simple neck model to the simulator.

### DIFF
--- a/Assets/VRTK/Examples/ExampleResources/Prefabs/SDKSetups.prefab
+++ b/Assets/VRTK/Examples/ExampleResources/Prefabs/SDKSetups.prefab
@@ -1921,6 +1921,21 @@ GameObject:
   m_NavMeshLayer: 0
   m_StaticEditorFlags: 0
   m_IsActive: 1
+--- !u!1 &1832880887312840
+GameObject:
+  m_ObjectHideFlags: 1
+  m_PrefabParentObject: {fileID: 0}
+  m_PrefabInternal: {fileID: 100100000}
+  serializedVersion: 5
+  m_Component:
+  - component: {fileID: 4486131483486730}
+  m_Layer: 0
+  m_Name: Neck
+  m_TagString: Untagged
+  m_Icon: {fileID: 0}
+  m_NavMeshLayer: 0
+  m_StaticEditorFlags: 0
+  m_IsActive: 1
 --- !u!1 &1846200688178366
 GameObject:
   m_ObjectHideFlags: 1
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: 6b9459e089d0c682db203ab29ea753704392e8f2

---

virror:
fixed missing space in title

Assets/VRTK/Examples/ExampleResources/Prefabs/SDKSetups.prefab
```diff
@@ -5144,12 +5173,12 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   autoPopulateObjectReferences: 1
-  actualBoundaries: {fileID: 1000012704122952}
```
bddckr: _All_ changes beginning here in _this_ file (_including_ this line) should be reverted.

Reason is that these got reset because you don't have all the vendor SDKs installed that VRTK supports. But since this is the _prefab_ I suggest to leave these auto populated references in so people don't freak out because the prefab gets modified all the time. It's better to let Unity only change the prefab _instances_ in a scene, than to change the prefab.

virror:
Fixed, sorry about that, thought the rigs where their own prefabs, will be more careful next time : )

Assets/VRTK/Examples/ExampleResources/Prefabs/SDKSetups.prefab
```diff
@@ -5018,7 +5047,7 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   autoPopulateObjectReferences: 1
   actualBoundaries: {fileID: 1033027119906706}
-  actualHeadset: {fileID: 1537013185507012}
```
bddckr: This and the following lines sneaked in again! 😄

virror: wtf : p

virror:
Lets try again : D

thestonefox:
is this good to go @bddckr @virror ?

virror:
Yes. The only think im thinking about is if the neck or camera should count as the headset? Atm its the camera, but its the neck that rotates? What are your opinion on this @thestonefox, want the simulator to be as close as possible to the real thing.

thestonefox:
what's the purpose of the neck addition? what was wrong before with moving the camera around with the mouse?

virror:
This should be pretty obvious, its just a matter of simulating a hmd better. If you even seen a head from the side you know the eyes are not positioned in the center of the head but at the front of it : ) A head/neck model is a classing thing going back to the DK1 days. Just make things behave more like irl.

virror:
@thestonefox anything more needed here?

thestonefox:
Can I be a pain and ask for the scope to be uppercased to `fix(Simulator)`

Also, come to think of it, the title doesn't need to contain the word fix, so something like:

`fix(Simulator): add simple neck model.`

also, again why is `Assets/VRTK/Examples/ExampleResources/Prefabs/SDKSetups.prefab` being changed?

virror:
Fixed!

PR #1220, opened 2017-05-21
feat/version symbol

bddckr:
A new script has been added that defines the current and previous
VRTK versions and adds a scripting define symbol to the player
settings for it. This symbol is useful for third party assets that
use VRTK.

There's an exact version symbol for the current version, as well as
additional `_OR_NEWER` symbols for the current and any previously
known versions of VRTK. They follow this pattern:

- `VRTK_VERSION_3_2_0`
- `VRTK_VERSION_3_2_0_OR_NEWER`

Additionally the Support Info editor window was updated to list this
VRTK version. A note hints that this version may not be correct in
case the GitHub version of VRTK is used.


bddckr:
Note that the version already uses `3.2.0`.

bddckr:
Oh and where do you think this file should live? It's currently at `Assets/VRTK/SDK/VRTK_Defines.cs`. I think both the `SDK` and `Scripts/Utilities` folders are kinda wrong for this 😞

thestonefox:
Ok yeah as we're only introducing this for 3.2.0 (so we won't merge this till before release anyway)

so we probably don't need the `OR_OLDER` but we should probably do `_OR_NEWER` and the `3_2_0` just for completeness.

Not sure where's best to put them, lets chat it through and see what's what (when I get 5 minutes!!!)

thestonefox:
@bddckr shall we look at getting this finished so we can merge it asap?

I think the `VRTK_Defines.cs` should probably go in `Assets/VRTK/Scripts/Internal` as it's an internal script that we don't what the users to use. Also make sure to update the script icon in unity so it's the grey pig with the dark grey line through it (the `Assets/Editor/Icons/vrtk_script_internal` icon)

Also it's probably worth adding the `3_2_0` so this PR will give us

 * `VRTK_VERSION_3_2_0_OR_NEWER`
 * `VRTK_VERSION_3_2_0`

thestonefox:
@bddckr is this ok to merge now? all looks ok to me

bddckr:
@thestonefox I don't like the PR name anymore (because of that second commit), but who cares. We have the commits in git and changelog after merging anyway.
So :shipit:

thestonefox:
Can just rename the PR in the GitHub page anyway if you want.

PR #1227, opened 2017-05-24
fix(Gripping): Grip release on non-grip joint break. - fixes #1269

### LINKED ISSUE #1269: Grip release on non-grip joint break.
### Environment
Github VRTK 94914f07c9a5098ff6a477df03f3e3c8ef8c917c
Vive, SteamVR

### Steps to reproduce

Create two grabbable VRTK_InteractableObjects, A and B.
Give them both rigidbodies.
Add a fixed joint to A, attached to B, with a break force of 100.
Grab A.

### Expected behavior

Object A should be grabbed, the fixed joint should snap, B should fall, and A should remain grabbed.

### Current behavior

When the fixed joint snaps, A is ungrabbed.

tjhowse:
Currently if joint on an object other than the gripping joint breaks, the grip is released. This shouldn't happen.

The grip should only be released if the gripping joint breaks. This change band-aids this issue. It should actually check that the grip joint generated the OnJointBreak, but I see no way of doing this.

### DIFF
--- a/Assets/VRTK/Scripts/Interactions/GrabAttachMechanics/VRTK_BaseJointGrabAttach.cs
+++ b/Assets/VRTK/Scripts/Interactions/GrabAttachMechanics/VRTK_BaseJointGrabAttach.cs
@@ -79,7 +79,10 @@ namespace VRTK.GrabAttachMechanics
 
         protected virtual void OnJointBreak(float force)
         {
-            ForceReleaseGrab();
+			if (controllerAttachJoint == null)
+			{
+				ForceReleaseGrab();
+			}
         }
 
         protected virtual void CreateJoint(GameObject obj)


GitCop:

 * Commit: 25b5125e864b7538947502aaebf8728432614b56

---

thestonefox:
I'm not sure what problem this is attempting to fix.

Could you raise a Github issue and follow the issue template and provide steps to reproduce

tjhowse:
Done: https://github.com/thestonefox/VRTK/issues/1269

thestonefox:
Thanks

thestonefox:
I'm happy to merge this, but can you just change the commit message to:

```
fix(Interaction): prevent secondary joint break causing force release

There was an issue where if a secondary joint was used to attach an
Interactable Object to another rigidbody and the Interactable Object
grab attach mechanic was a joint type, then if the secondary joint
was broken then the OnJointBreak would be called on the Interactable
Object and this would cause the ForceReleaseGrab to be called
meaning the object would be dropped.

It would seem that a quirk of how the joints are ordered that if
the secondary joint is broken then the object will ignore the
controllerAttachJoint so checking to see if this is set will determine
if the breaking joint is that of the grab attach mechanic or the
secondary joint.
```

To amend your commit just do

`git commit --amend`

change the message then force push back up with `git push tjhowse:master -f`

(although you should probably branch off your master)

If this seems too much hassle just let me know and I'll push the fix as it's only like a 1 liner :)

thestonefox:
I'm going to need to revert this commit, it doesn't fix the problem and causes more issues,

PR #1234, opened 2017-05-28
feat(setupWindow): support multi selection

virror:
Support selecting several objects at once when
using the setup window.


virror:
I should have done this ages ago, its really a time saver : D

PR #1237, opened 2017-05-29
feat(Simulator): add sprint and new distance grab mode

virror:
Add sprint modifier key + relevant settings for it.
Adjusted near clip plane from 3dm to 5cm.
Add new distance grab mechanic, when in "mouse look"
mode its possible to press a modifier key (default left shift)
to get a crosshair, when aiming that crosshair on a
object with InteractibleObject on it, that object will
get picked up depending on if left or right hand key
was pressed.


thestonefox:
I've not had chance to try this, but I love the idea of the distance grab mode! VRTK just needs a simulator now to get off life support after the failed kickstarter!

One quick edit just change the scope from `sim` to `Simulator` as I believe that's what previous scopes for the simulator have been (this makes it easier for me when generating the change logs)

so:

`feat(Simulator): add sprint and new distance grab mode`

virror:
Fixed!

thestonefox:
Why is `Assets/VRTK/Examples/ExampleResources/Prefabs/SDKSetups.prefab` being changed?

Also, I don't think the crosshair.png should go in the Materials directory because it's not a material.

I think it should probably go in `Assets/VRTK/Prefabs/Resources/Images` (or `/Textures`)

virror:
Fixed!

thestonefox:
Ok I checked it out, it all works as expected, but I think the following changes are required:

### Commit Description

Contains spelling errors:

> Add sprint modifier key + relevant settings for it.
> Adjusted near **ckip** plane from 3**dm** to 5cm.
> Add new distance grab mechanic, when in "mouse look"
> mode its possible to press a modifier key (default left shift)
> to get a **crosshar**, when aiming that crosshair on a
> object with InteractanöeObject on it, that object will
> get picked up depending on if left or right hand key
> was pressed.

### Cross Hair graphic

I think it needs a white outline around the crosshair to give it some contrast, otherwise over a black object (or dark object) you can't see the cross hair. (if you need me to sort the graphic lemme know and I'll do it and send it to be included in the commit).

### SDK Setup

If you look at the latest github master for the SDK Setups prefab, i've unticked the auto populate button on the Simulator, because there is an issue if you leave it auto populate and switch scenes it doesn't auto populate and the first time you run the scene it crashes (then populates correctly). So I'd suggest updating to the latest github master and rebasing those changes into this one.

Although you may be better creating a new branch from github master and copying your changes over otherwise the prefab is going to be a world of merge pain.

virror:
Fixed!

virror:
Seems like the rebase went without a hitch, but you better make sure : p

PR #1238, opened 2017-05-29
refactor(Structure): replace FindChild with Find

virror:
Transform.FindChild() is no longer part of the documentation,
but still exists as part of the UnityEngine assembly.
It is depricated and is functionally identical to Transform.Find().

### DIFF
--- a/Assets/VRTK/Examples/ExampleResources/Scripts/Archery/ArrowNotch.cs
+++ b/Assets/VRTK/Examples/ExampleResources/Scripts/Archery/ArrowNotch.cs
@@ -9,7 +9,7 @@
 
         private void Start()
         {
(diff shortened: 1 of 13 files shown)


thestonefox:
Code looks ok, but i'd change the git commit message to:

```
refactor(Structure): replace FindChild with Find

Transform.FindChild() is no longer part of the documentation,
but still exists as part of the UnityEngine assembly.
It is depreciated and is functionally identical to Transform.Find().
```

Reason is because it's a refactor as no changes will be noticed (also afaik you get no deprecation warnings even in 5.6), the description text is the official text from Unity on the matter so makes a nice commit description imho.

virror:
Fixed!

PR #1241, opened 2017-05-30
feat(Presence): allow custom body and feet colliders on body physics

virror:
Allow the user to specify custom GameObjects containing
colliders to be used instead of generating its own.

### DIFF
--- a/Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
+++ b/Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
@@ -75,6 +75,10 @@ namespace VRTK
         public int standingHistorySamples = 5;
         [Tooltip("The `y` distance between the headset and the object being leaned over, if object being leaned over is taller than this threshold then the current standing position won't be updated.")]
         public float leanYThreshold = 0.5f;
+        [Tooltip("A GameObject containing a Capsule Collider used as a custom body collider.")]
+        public GameObject customBodyCollider;
+        [Tooltip("A GameObject containing a Capsule Collider used as a custom feet collider.")]
+        public GameObject customFootCollider;
 
         [Header("Step Settings")]
 
@@ -138,7 +142,6 @@ namespace VRTK
         protected Transform playArea;
         protected Transform headset;
         protected Rigidbody bodyRigidbody;
-        protected GameObject bodyColliderContainer;
         protected CapsuleCollider bodyCollider;
         protected CapsuleCollider footCollider;
         protected VRTK_CollisionTracker collisionTracker;
@@ -310,7 +313,7 @@ namespace VRTK
         /// <returns></returns>
         public virtual GameObject GetBodyColliderContainer()
         {
-            return bodyColliderContainer;
+            return customBodyCollider;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


thestonefox:
change scope to `Presence` then mention body physics in the title like:

`feat(Presence): allow custom body and feet colliders on body physics`

This way when I generate the change log, all `presence` related scripts will be grouped.

I'm going to check over the code now too and may leave some inline comments

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
```diff
@@ -75,6 +75,10 @@ public enum FallingRestrictors
         public int standingHistorySamples = 5;
         [Tooltip("The `y` distance between the headset and the object being leaned over, if object being leaned over is taller than this threshold then the current standing position won't be updated.")]
         public float leanYThreshold = 0.5f;
+        [Tooltip("A GameObject containing a Capsule Collider used as a custom body collider.")]
```
thestonefox: I feel these should be under the `Custom Settings` header as I usually put any custom injected game object within the Custom Settings bit.

thestonefox: Also, is it necessary to pass a GameObject as the custom colliders? Would it not be better to have the type as `Collider`? Something seems mismatched between the variable type and the variable name.

I guess either `GameObject customXXColliderContainer` or `Collider customXXCollider`

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
```diff
@@ -869,22 +872,44 @@ protected virtual void CreateCollider()
...
             {
                 generateCollider = true;
-                bodyColliderContainer = CreateColliderContainer(BODY_COLLIDER_CONTAINER_NAME, playArea);
+                customBodyCollider = CreateColliderContainer(BODY_COLLIDER_CONTAINER_NAME, playArea);
 
-                bodyCollider = bodyColliderContainer.AddComponent<CapsuleCollider>();
+                bodyCollider = customBodyCollider.AddComponent<CapsuleCollider>();
                 bodyCollider.radius = bodyRadius;
 
+                customBodyCollider.gameObject.layer = LayerMask.NameToLayer("Ignore Raycast");
```
thestonefox: Are you doing this so your custom injected object can be set to a layer that's not on the ignore raycast layer?

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
```diff
@@ -869,22 +872,44 @@ protected virtual void CreateCollider()
...
+            }
+            else
+            {
+                customBodyCollider.transform.SetParent(playArea);
```
thestonefox: I'm guessing this means that the object has to exist in the scene and you can't use a prefab for the injected game object?

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
```diff
@@ -869,22 +872,44 @@ protected virtual void CreateCollider()
...
+                bodyCollider = customBodyCollider.GetComponent<CapsuleCollider>();
+                VRTK_PlayerObject.SetPlayerObject(customBodyCollider, VRTK_PlayerObject.ObjectTypes.Collider);
```
thestonefox: This isn't being set for the auto generated container anymore and only for the injected one, I guess this line needs to move out of the if/else statement

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
```diff
@@ -869,22 +872,44 @@ protected virtual void CreateCollider()
...
                 {
-                    GameObject footColliderContainer = CreateColliderContainer(FOOT_COLLIDER_CONTAINER_NAME, bodyColliderContainer.transform);
-                    footCollider = footColliderContainer.AddComponent<CapsuleCollider>();
+                    customFootCollider = CreateColliderContainer(FOOT_COLLIDER_CONTAINER_NAME, customBodyCollider.transform);
+                    footCollider = customFootCollider.AddComponent<CapsuleCollider>();
+                }
+            }
+            else
+            {
+                customFootCollider.transform.SetParent(customBodyCollider.transform);
```
thestonefox: This still creates the custom foot collider even if the step up Y offset is `0`, the above code does a check to see if it's above 0f

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
virror: I thought it would be dumb to force the layer since its a custom collider, would kind of defeat the purpose?

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
virror: You mean that its not instantiated?

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
virror: Ahh, will fix : )

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
virror: I was first planning on customXXColliderContainer but it turned out to long for the inspector : p
But maybe thats ok

Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
virror: No, its being set by the CreateColliderContainer function as before, so it should be ok.

thestonefox:
PR to supersede this one: https://github.com/thestonefox/VRTK/pull/1271

PR #1247, opened 2017-06-04
fix(Utilities): allow for trigger colliders to be ignored by casts - fixes #1230

### LINKED ISSUE #1230: VRTK_BodyPhysics Generates Physics Response Upon Colliding with Trigger Colliders
### Environment

 * Source of VRTK (Github)
 * Commit cf63d2b5bd173931900b6dc3b89aba5684f09247
 * Unity 5.6.1p1
 * Vive)
 * SDK used (e.g. OpenVR/SteamVR/Oculus Utilities)

### Steps to reproduce

1. Create a basic default VRTK scene.
2. Add a trigger collider to the scene.
3. Move into the trigger collider.

### Expected behavior

VRTK_BodyPhysics does not generate a physics response when the player object collides with a collider set to trigger.

### Current behavior

VRTK_BodyPhysics generates a physics response.

There is a workaround, and that is to add a VRTK_CustomRaycast script to the PlayArea and to set Trigger Interaction to "Ignore."

However, the default behavior VRTK should not be to respond to triggers like normal colliders. The trigger setting on colliders is a way of turning off physics collisions while still receiving the triggered events upon collision.

thestonefox:
The various casts done by scripts such as Body Physics, teleporters
and the pointers were all colliding with trigger colliders by default.

The CustomRaycast script has now been updated so it can be called
and informed to ignore trigger colliders without providing an
instantiated CustomRaycast object.

Body Physics and the teleporters now ignore trigger colliders by
default, whereas the pointers still interact with trigger colliders
by default because the UI Canvas generates a trigger collider.

If the pointer is required to pass through a trigger collider then
the object should be put on the Ignore Raycasts layer.

### DIFF
--- a/Assets/VRTK/Scripts/Locomotion/VRTK_HeightAdjustTeleport.cs
+++ b/Assets/VRTK/Scripts/Locomotion/VRTK_HeightAdjustTeleport.cs
@@ -62,7 +62,7 @@ namespace VRTK
             Ray ray = new Ray(tipPosition + rayStartPositionOffset, -playArea.up);
             RaycastHit rayCollidedWith;
 #pragma warning disable 0618
-            if (target && VRTK_CustomRaycast.Raycast(customRaycast, ray, out rayCollidedWith, layersToIgnore, Mathf.Infinity))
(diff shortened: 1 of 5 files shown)


thestonefox:
The pointers by default shouldn't ignore triggers otherwise the UI Canvases won't work.

I'd say if you want to have pointers avoid triggers then you set them up manually

PR #1248, opened 2017-06-04
fix(Slide Locomotion): decrease y axis' role in SlideObjectControlAction

SpencerMalone:
Scaling down the Y axis (but not to 0) makes it so that you will experience less "slowdown" while angling the controllers up or down when using the slide locomotion, but I propose avoid just forcing it to zero as that can cause a jerky jolt if you spin the controller vertically.


thestonefox:
I'm not sure what problem this is attempting to fix.

Could you raise a Github issue and follow the issue template and provide steps to reproduce

PR #1252, opened 2017-06-05
fix(SDKManager): honour script enable states - fixes #1216

### LINKED ISSUE #1216: VRTK scripts disabled before running will get enabled anyways when running.
When using scripts like VRTK_BasicTeleport or VRTK_BodyPhysics and you want to disable it for testing something, it will get enabled anyways by the SDK manager. The manager should respect the initial enable state for those scripts. Atm the only way of running without a certain script enabled is to remove it, which will make it loose all the settings.
This issue is also confirmed by @bddckr

thestonefox:
The VRTK scripts that register with the SDK Manager were always
being set to enabled even if the script was set to disabled in the
editor.

This fix stores the state of the registered script and then sets it
back to that original state when the SDK Manager runs through the
behaviours registered with it. This means the enabled state of the
scripts are now remembered and set appropriately.


thestonefox:
I think I'll let @virror test it before merging just to make sure it doesn't break anything I've not understood ;)

virror:
LGTM

PR #1254, opened 2017-06-06
fix(Interactions): pass relevant VRTK_InteractX to interactable object

thestonefox:
The Interactable Object script would previously accept a GameObject
when the Touching/Grabbing/Using methods were called. However, this
was ambiguous as to what the object should represent. The object
that is performing the action should be the one with the relevant
VRTK_InteractX script on like VRTK_InteractTouch, VRTK_InteractGrab
and VRTK_InteractUse.

Because of this, it makes more sense to actually pass the reference
to the script as the GameObject can be obtained from the script
with less overhead than the script being obtained from the GameObject
via the `GetComponent` method call.

The existing methods that accept a GameObject still exist but have
been deprecated in favour of the new methods that take the specific
InteractX script.

### DIFF
--- a/Assets/VRTK/Examples/ExampleResources/Scripts/FireExtinguisher_Base.cs
+++ b/Assets/VRTK/Examples/ExampleResources/Scripts/FireExtinguisher_Base.cs
@@ -10,13 +10,13 @@
 
         private VRTK_ControllerEvents controllerEvents;
 
-        public override void StartUsing(GameObject usingObject)
+        public override void StartUsing(VRTK_InteractUse usingObject)
         {
             base.StartUsing(usingObject);
             controllerEvents = usingObject.GetComponent<VRTK_ControllerEvents>();
         }
 
-        public override void StopUsing(GameObject previousUsingObject)
+        public override void StopUsing(VRTK_InteractUse previousUsingObject)
         {
             base.StopUsing(previousUsingObject);
             controllerEvents = null;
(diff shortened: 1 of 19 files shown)


TheoneandonlyJOD:
I have a bug where a pointer wont trigger startusing functions.
I put the pointerscript on a non controller gameobject, but you can reproduce the bug an easier way:
1. Open Example Scene 019_COntroller_InteractinWithPointer
2. Select Object RightController
3. Attach the Object RightController to the slot Controller in the script VRTK_Pointer

The pointer wont make the whyrligig spin anymore, even though it should.

thestonefox:
Thanks @TheoneandonlyJOD I've found the issue causing that problem and fixed it. If you could test again that would be great

TheoneandonlyJOD:
@thestonefox Sure thing! Where can I find the fixed Version?

thestonefox:
@TheoneandonlyJOD I just force pushed to this Pull Request, so if you get the code from this pull request it should now be the latest

TheoneandonlyJOD:
Now it works as before again, but if I call "Debug.Log(this.usingObject.gameObject);" in StartUsing(...) I still get "RightController" eventhough the StartUsing is being activated by a pointer attached to a GameObject that is not a controller.

thestonefox:
@TheoneandonlyJOD are you able to recreate that in the example scene? The code now checks to see if the attached object has the InteractUse script on it, if it doesn't then it falls back to the controller to see if that has interact use on it.

Does your game object that you're expecting to see have the InteractUse script on it?

TheoneandonlyJOD:
Thank you that was my mistake. I didnt attach the interact use script. I always thought the pointer or interactable object would extend the interact use script. Now it returns the right object.

TheoneandonlyJOD:
But now that I attached the interact use script to my object I cant seem to grab it properly with the controller. Well, I will look at that closer tomorrow.

thestonefox:
superseded by https://github.com/thestonefox/VRTK/pull/1279

PR #1263, opened 2017-06-08
feat(Presence): option to ignore trigger colliders for head collisions

thestonefox:
The Headset Collision script now has a new parameter to allow it to
ignore colliding with Trigger colliders.

This option is also now in the Position Rewind to allow the rewind
to be ignored when colliding with a trigger collider (but the headset
collision could still fade the headset on trigger collision if that was
required).


nadimalam:
this works perfectly now, thank you!

PR #1266, opened 2017-06-09
feat(Presence): provide collision sweep detection for body physics - fixes #1259

### LINKED ISSUE #1259: Player able to walk through doors using trackpad movement
### Environment

 * Source of VRTK (Unity Asset Store or **Github**)
Github
 * Version of VRTK (Unity Asset Store/Github release number) (Github master commit hash)
Commit: 94914f07c9a5098ff6a477df03f3e3c8ef8c917c [94914f0] (Latest Master Release)
 * Version of Unity3D (e.g. Unity 5.4.4f1)
5.5.0f3
 * Hardware used (e.g. Vive/Oculus)
Vive
 * SDK used (e.g. OpenVR/SteamVR/Oculus Utilities)
SteamVR

### Steps to reproduce

 * Open example scene `025_Controls_Overview`
 * Copy one of the `NormalDoor` objects e.g. `ExampleWorldObjects->NormalDoor (inward normal)`
 * Open example scene `017_CameraRig_TouchpadWalking`
 * Paste the `NormalDoor` into the scene.
 * Start the scene and walk at the door (in to the side that shouldn't just swing open)

### Expected behavior

When you walk into door, the players view should be faded to black and be rewinded to his last known good position.

### Current behavior

I am able to walk right through the door, without any fade to black occuring and without any position rewind taking place.

thestonefox:
The Body Physics script now has a new method `SweepCollision` that
can be used to determine if the body physics body collider will
collide with an object if it continues to move in a given direction
for a given distance.

This new method is now used in some of the Locomotion methods that
move the play area via `transform.position` and can now check
if they are provided with a Body Physics script if when they move
forward, will the new position cause collision tunnelling and if
it does then the movement can be denied meaning tunnelling is not
possible.

### DIFF
--- a/Assets/VRTK/Scripts/Locomotion/ObjectControlActions/VRTK_BaseObjectControlAction.cs
+++ b/Assets/VRTK/Scripts/Locomotion/ObjectControlActions/VRTK_BaseObjectControlAction.cs
@@ -152,5 +152,17 @@ namespace VRTK
 
             return axisDirection;
         }
+
+        protected virtual bool CanMove(VRTK_BodyPhysics givenBodyPhysics, Vector3 currentPosition, Vector3 proposedPosition)
+        {
+            if (givenBodyPhysics == null)
+            {
+                return true;
+            }
+
+            Vector3 proposedDirection = (proposedPosition - currentPosition).normalized;
+            float distance = Vector3.Distance(currentPosition, proposedPosition);
+            return !givenBodyPhysics.SweepCollision(proposedDirection, distance);
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


mikekan13:
Does this fix the issue where if you stand close to a wall and then in real life walk forward you end up on the other side?

thestonefox:
@mikekan13 no, that's what the Position Rewind script does.

mikekan13:
Position Rewind doesn't stop that. If I set it up so that in the real world I am standing far away from my VR bounds and then place my virtual body close to the wall and then begin walking in real life it eventually stops rewinding me and I just walk into the wall. After I walk a bit into the wall it will teleport to the other side of the wall.

thestonefox:
Position rewind does stop that. If you physically walk into a wall then it will reset your position back to the last safe known location

mikekan13:
I can walk through walls consistently. Maybe it is something in my project I will test with a clean version.

PR #1271, opened 2017-06-10
feat(Presence): allow custom body/feed colliders for body physics

thestonefox:
It's now possible to provide a custom body collider container and
a custom foot collider container for the Body Physics script.

The containers need to include a capsule collider (no other collider
is allowed). If a capsule collider isn't present on the custom
GameObject then one will be auto generated.


virror:
LGTM

PR #1280, opened 2017-06-13
feat(Utilities): ability to modify transform orientation based on sdk - fixes #1268

### LINKED ISSUE #1268: Ability to provide varying snap handle transform info based on controller type
Currently when applying a snap handle to a Interactable Object, it will only be at the correct orientation based on one controller.

However, the Vive wand in SteamVR is a different orientation to the Oculus Touch on home (and also different to Oculus Touch on SteamVR).

There needs to be a way that you can specify a snap handle but with different transform position/rotations based on the current controller being used.

thestonefox:
The new SDK Transform Modify script can be used to change the
orientation of a GameObject transform determined by the current
loaded SDK.

This is useful if transform orientations differ between SDK setups
or different controllers within the SDK (and cross other SDKs).

The main use for this is to ensure that snap handles on interactable
objects can be positioned accordingly for each SDK controller, for
example the snap handle on the SteamVR Vive wand is at a different
orientation to that of the Oculus Touch controller.

### DIFF
--- a/Assets/VRTK/SDK/Fallback/SDK_FallbackController.cs
+++ b/Assets/VRTK/SDK/Fallback/SDK_FallbackController.cs
@@ -32,6 +32,15 @@ namespace VRTK
         }
 
         /// <summary>
+        /// The GetCurrentControllerType method returns the current used ControllerType based on the SDK and headset being used.
+        /// </summary>
+        /// <returns>The ControllerType based on the SDK and headset being used.</returns>
+        public override ControllerType GetCurrentControllerType()
+        {
+            return ControllerType.Undefined;
+        }
+
+        /// <summary>
         /// The GetControllerDefaultColliderPath returns the path to the prefab that contains the collider objects for the default controller of this SDK.
         /// </summary>
         /// <param name="hand">The controller hand to check for</param>
--- a/Assets/VRTK/Scripts/Utilities/VRTK_SDKTransformModify.cs
+++ b/Assets/VRTK/Scripts/Utilities/VRTK_SDKTransformModify.cs
@@ -1,0 +1,99 @@
+﻿// SDK Transform Modify|Utilities|90064
+namespace VRTK
+{
+    using UnityEngine;
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    [Serializable]
+    public class VRTK_SDKTransformModifiers
+    {
+        [Header("SDK settings")]
+        [Tooltip("An optional SDK Setup to use to determine when to modify the transform.")]
+        public VRTK_SDKSetup loadedSDKSetup;
+        [Tooltip("An optional SDK controller type to use to determine when to modify the transform.")]
+        public SDK_BaseController.ControllerType controllerType;
+
+        [Header("Transform Override Settings")]
+
+        [Tooltip("The new local position to change the transform to.")]
+        public Vector3 position = Vector3.zero;
+        [Tooltip("The new local rotation in eular angles to change the transform to.")]
+        public Vector3 rotation = Vector3.zero;
+        [Tooltip("The new local scale to change the transform to.")]
+        public Vector3 scale = Vector3.one;
+    }
+
+    /// <summary>
+    /// The SDK Transform Modify can be used to change a transform orientation at runtime based on the currently used SDK or SDK controller.
+    /// </summary>
+    public class VRTK_SDKTransformModify : MonoBehaviour
+    {
+        [Tooltip("The target transform to modify on enable. If this is left blank then the transform the script is attached to will be used.")]
+        public Transform target;
+        [Tooltip("A collection of SDK Transform overrides to change the given target transform for each specified SDK.")]
+        public List<VRTK_SDKTransformModifiers> sdkOverrides = new List<VRTK_SDKTransformModifiers>();
+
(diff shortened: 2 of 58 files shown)


Assets/VRTK/SDK/Fallback/SDK_FallbackController.cs
```diff
@@ -32,6 +32,15 @@ public override void ProcessFixedUpdate(VRTK_ControllerReference controllerRefer
         }
 
         /// <summary>
+        /// The GetCurrentControllerType method returns the current used ControllerType based on the SDK and headset being used.
+        /// </summary>
+        /// <returns>The ControllerType based on the SDK and headset being used.</returns>
+        public override ControllerType GetCurrentControllerType()
+        {
+            return ControllerType.Custom;
```
bddckr: Shouldn't this use `Undefined`?

Assets/VRTK/Scripts/Utilities/VRTK_SDKTransformModify.cs
```diff
@@ -0,0 +1,75 @@
...
+    using UnityEngine;
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    [Serializable]
+    public class VRTK_SDKTransformModifiers
+    {
+        [Tooltip("An optional SDK Setup to use to determine when to modify the transform.")]
+        public VRTK_SDKSetup sdkType;
```
bddckr: Should this variable be named `sdkSetup`?

Assets/VRTK/Scripts/Utilities/VRTK_SDKTransformModify.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            //If no sdk set up is found or it is null then try and find by the SDK controller
+            if (selectedModifier == null)
+            {
+                SDK_BaseController.ControllerType currentController = VRTK_DeviceFinder.GetCurrentControllerType();
+                selectedModifier = overrides.FirstOrDefault(item => item.controllerType == currentController);
+            }
+
+            //If a modifier is found then change the transform
+            if (selectedModifier != null && target != null)
```
bddckr: Remove the check `target != null` here and use it right at the start of this method to bail out early `if (target == null)`. This will save some unnecessary computation since we don't use the resulting modifier anyway in case of no target.

Assets/VRTK/Scripts/Utilities/VRTK_SDKTransformModify.cs
```diff
@@ -0,0 +1,75 @@
...
+    using System.Linq;
+
+    [Serializable]
+    public class VRTK_SDKTransformModifiers
+    {
+        [Tooltip("An optional SDK Setup to use to determine when to modify the transform.")]
+        public VRTK_SDKSetup sdkType;
+        [Tooltip("An optional SDK controller type to use to determine when to modify the transform.")]
+        public SDK_BaseController.ControllerType controllerType;
+        [Tooltip("The new position to change the transform to.")]
```
bddckr: This should probably mention that it's in `target`'s local space.

Assets/VRTK/Scripts/Utilities/VRTK_SDKTransformModify.cs
```diff
@@ -0,0 +1,75 @@
...
+        public Vector3 position = Vector3.zero;
+        [Tooltip("The new rotation to change the transform to.")]
```
bddckr: This should probably mention that it's in `target`'s local space.

Assets/VRTK/Scripts/Utilities/VRTK_SDKTransformModify.cs
```diff
@@ -0,0 +1,75 @@
...
+        public Vector3 rotation = Vector3.zero;
+        [Tooltip("The new scale to change the transform to.")]
```
bddckr: This should probably mention that it's in `target`'s local space.

Assets/VRTK/SDK/Fallback/SDK_FallbackController.cs
thestonefox: yeah, it should (undefined was a last minute addition :p)

Assets/VRTK/Scripts/Utilities/VRTK_SDKTransformModify.cs
thestonefox: maybe `sdkSetupType`? It's the type of SDK to check for

bddckr: I think this is misleading, since this PR is only doing an equals check with the currently loaded SDK Setup. If we want real "type" checking, we'd probably check the used SDK settings inside the Setup? Though I'm not sure how that should work... And checking for SDK Setup equality is probably fine, since the user can provide their own strange and custom ones.

We just need to make sure users understand how this is used when checking, I'm fine with whatever name, it's just a suggestion from me. 😄

thestonefox: what about `loadedSDKSetup`?

bddckr: 👍

PR #1281, opened 2017-06-14
feat(Pointer): set PointerDirectionIndicator's color

Backer-Sultan:
Make PointerDirectionIndicator's color to match the pointer-cursor's
color so it looks the same when used with BezierPointerRenderer.

### DIFF
--- a/Assets/VRTK/Prefabs/Resources/Scripts/VRTK_PointerDirectionIndicator.cs
+++ b/Assets/VRTK/Prefabs/Resources/Scripts/VRTK_PointerDirectionIndicator.cs
@@ -63,6 +63,18 @@ namespace VRTK
         }
 
         /// <summary>
+        /// The SetColor method is used to set the color for the direction indicator to match the custom pointer cursor.
+        /// </summary>
+        /// <param name="color">The color to assign.</param>
+        public virtual void SetColor(Color color)
+        {
+            foreach(Renderer rend in GetComponentsInChildren<Renderer>())
+            {
+                rend.material.color = color;
+            }
+        }
+
+        /// <summary>
         /// The GetRotation method returns the current reported rotation of the direction indicator.
         /// </summary>
         /// <returns>The reported rotation of the direction indicator.</returns>
--- a/Assets/VRTK/Scripts/Pointers/PointerRenderers/VRTK_BasePointerRenderer.cs
+++ b/Assets/VRTK/Scripts/Pointers/PointerRenderers/VRTK_BasePointerRenderer.cs
@@ -439,6 +439,7 @@ namespace VRTK
                 currentColor = givenColor;
                 ChangeMaterial(givenColor);
             }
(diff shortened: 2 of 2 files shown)


Assets/VRTK/Prefabs/Resources/Scripts/VRTK_PointerDirectionIndicator.cs
```diff
@@ -63,6 +64,21 @@ public virtual void SetPosition(bool active, Vector3 position)
...
 
         /// <summary>
+        /// The SetColor method is used to set the clor for the direction indicator to match the custom pointer cursor.
+        /// </summary>
+        /// <param name="color">The color to assign.</param>
+        public virtual void SetColor(Color color)
+        {
+            List<Renderer> renderers = new List<Renderer>();
+            renderers.Add(GetComponent<Renderer>());
+            GetComponentsInChildren<Renderer>(renderers);
```
bddckr: I may be wrong here, but AFAIK `GetComponentsInChildren` already includes the parent/self game object's component, so there's no need to do it this way since you're setting the color twice for the current game object.

Backer-Sultan: @bddckr well if this is true then I don't need to use List<T> as well. 
I just need to store them in an array instead.

bddckr: Can even do `foreach(Renderer renderer in GetComponentsInChildren<Renderer>)` then 😄

Backer-Sultan: That's so clean! I like it! 👍

Assets/VRTK/Prefabs/Resources/Scripts/VRTK_PointerDirectionIndicator.cs
```diff
@@ -63,6 +63,18 @@ public virtual void SetPosition(bool active, Vector3 position)
         }
 
         /// <summary>
+        /// The SetColor method is used to set the clor for the direction indicator to match the custom pointer cursor.
```
bddckr: Typo: `clor` 😄

Backer-Sultan: I'll fix this 😅

thestonefox:
@Backer-Sultan here's my PR on this:

https://github.com/thestonefox/VRTK/pull/1287

It's a bit more in depth :) but covers a lot more features. Including new events for valid/invalid cursor states, changing the colour of the direction indicator, allowing for a valid/invalid direction indicator object, the moving of the direction indicator to be a dependency of the pointer renderer, and the ability to hide the direction indicator when hovering over a destination point.

Backer-Sultan:
Great! I close this PR then.

PR #1292, opened 2017-06-16
fix(UIPointerEvents_ListenerExample): Ensure pointers not toggled if active

ciwolsey:
Calling VRTK_Pointer.Toggle(true) when the pointer is already active seems to cause it to toggle false, then true under some situations causing the pointer to flicker on and off rapidly.


GitCop:

 * Commit: 7b403625ae7db101b23335e3c2fe953d7a2077a6

---

thestonefox:
Should we not fix this within the actual Pointer code? So if you call `Toggle(true)` but it's already on then the `Toggle` method should just exit early?

thestonefox:
This is my PR to solve it in the UIPointer

https://github.com/thestonefox/VRTK/pull/1294

What do you think?

ciwolsey:
Should be good, I can test this when I wake if you like.

PR #1293, opened 2017-06-16
fix(SDK): SteamVR defines

MayBeTall:
Type.GetType requires the use of an assembly name. This fix adds an improved version.

### DIFF
--- a/Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
+++ b/Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
@@ -180,6 +180,24 @@ namespace VRTK
         }
 
         /// <summary>
+        /// The GetTypeUnknownAssembly method is used to find a Type without knowing the exact assembly it is in.
+        /// </summary>
+        /// <param name="typeName">The name of the type to get.</param>
+        /// <returns>The Type, or null if none is found.</returns>
+        public static Type GetTypeUnknownAssembly(string typeName)
+        {
+            var type = Type.GetType(typeName);
+            if (type != null) return type;
+            foreach (var a in AppDomain.CurrentDomain.GetAssemblies())
+            {
+                type = a.GetType(typeName);
+                if (type != null)
+                    return type;
+            }
+            return null;
+        }
+
+        /// <summary>
         /// The IsEditTime method determines if the state of Unity is in the Unity Editor and the scene is not in play mode.
         /// </summary>
         /// <returns>Returns true if Unity is in the Unity Editor and not in play mode.</returns>
--- a/Assets/VRTK/SDK/SteamVR/SDK_SteamVRDefines.cs
+++ b/Assets/VRTK/SDK/SteamVR/SDK_SteamVRDefines.cs
@@ -38,7 +38,7 @@ namespace VRTK
                 return false;
             }
 
-            return systemMethodParameters[0].ParameterType == Type.GetType("Valve.VR.EVREventType");
+            return systemMethodParameters[0].ParameterType == VRTK_SharedMethods.GetTypeUnknownAssembly("Valve.VR.EVREventType");
         }
 
         [SDK_ScriptingDefineSymbolPredicate(ScriptingDefineSymbol, BuildTargetGroupName)]


GitCop:

 * Commit: fca21618a4e3400aa4e5d605c106336d577eb6a6

---

bddckr:
Interesting. So you compile the SteamVR plugin into its own .dll?

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -180,6 +180,24 @@ public static float RoundFloat(float givenFloat, int decimalPlaces, bool rawFide
...
 
         /// <summary>
+        /// The GetTypeUnknownAssembly method is used to find a Type without knowing the exact assembly it is in.
+        /// </summary>
+        /// <param name="typeName">The name of the type to get.</param>
+        /// <returns>The Type, or null if none is found.</returns>
+        public static Type GetTypeUnknownAssembly(string typeName)
+        {
+            var type = Type.GetType(typeName);
+            if (type != null) return type;
```
bddckr: Must use braces, see [here](https://github.com/thestonefox/VRTK/blob/master/CONTRIBUTING.md#coding-conventions).

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -180,6 +180,24 @@ public static float RoundFloat(float givenFloat, int decimalPlaces, bool rawFide
...
+            foreach (var a in AppDomain.CurrentDomain.GetAssemblies())
+            {
+                type = a.GetType(typeName);
+                if (type != null)
```
bddckr: Must use braces, see [here](https://github.com/thestonefox/VRTK/blob/master/CONTRIBUTING.md#coding-conventions).

bddckr:
Ah I see. You're putting the SteamVR plugin into the `Plugins` folder probably. Note that you shouldn't do this, as it's a special folder.

Assembly-CSharp-firstpass is `Phase 1` [here](https://docs.unity3d.com/Manual/ScriptCompileOrderFolders.html).

MayBeTall:
Ah, SteamVR used to put openvr_api.cs into the Plugins folder. When upgrading that file needs to be deleted or it will do this.

bddckr:
That doesn't matter, as that file is just the OpenVR API. SteamVR is never put into the Plugins folder by default. You can see that it's put into the Assets/SteamVR folder [here](https://github.com/ValveSoftware/steamvr_unity_plugin/blob/59cdcce1e69e17946168f1c11b56258225958315/Assets/SteamVR/Editor/SteamVR_Editor.cs#L112-L123), which means you as user probably moved it into Plugins later instead.

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -180,6 +180,24 @@ public static float RoundFloat(float givenFloat, int decimalPlaces, bool rawFide
         }
 
         /// <summary>
+        /// The GetTypeUnknownAssembly method is used to find a Type without knowing the exact assembly it is in.
+        /// </summary>
+        /// <param name="typeName">The name of the type to get.</param>
+        /// <returns>The Type, or null if none is found.</returns>
+        public static Type GetTypeUnknownAssembly(string typeName)
+        {
+            var type = Type.GetType(typeName);
```
thestonefox: Don't use `var` as you can't determine the variable type from the right of the equals

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
thestonefox:
```
if (type != null)
{
  return type;
}
```

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -180,6 +180,24 @@ public static float RoundFloat(float givenFloat, int decimalPlaces, bool rawFide
...
+            if (type != null) return type;
+            foreach (var a in AppDomain.CurrentDomain.GetAssemblies())
```
thestonefox: Should we try and stick to `for` rather than `foreach` for those poor folks using unity 5.5?

bddckr: This is only run once.

thestonefox: I guess if it's just once then the `foreach` isn't a big deal :)

thestonefox:
@bddckr just to confirm, is this PR a valid fix?

bddckr:
@thestonefox It is! But we probably should update all the other `SDK_*Defines` classes, too.

thestonefox:
@bddckr I'm going to look at doing a PR today re-implementing this work but for the other define classes too (as you suggested). Otherwise we may not get this ready in time for 3.2.0 as it's been 4 days since original commit and no sign from the author :(

thestonefox:
@bddckr only SteamVRDefines uses `Type.GetType`

The other defines all seem to use `Assembly.GetType`

thestonefox:
So does that mean for instance instead of doing

`Type eventClass = typeof(SDK_SteamVRDefines).Assembly.GetType("SteamVR_Events");`

we should do

`Type eventClass = VRTK_SharedMethods.GetTypeUnknownAssembly("SteamVR_Events");`

thestonefox:
PR to supersede this PR https://github.com/thestonefox/VRTK/pull/1309

PR #1294, opened 2017-06-16
fix(UIPointer): ensure toggle does not execute if on existing state

thestonefox:
The UIPointer `Toggle` method now exits early if the toggle state
that is passed to the method is already in that state.

e.g. If `Toggle(true)` is called but the pointer is already active
then the Toggle method will exit early.


ciwolsey:
Will test when I wake.

ciwolsey:
Tested, can merge, thanks :)

PR #1298, opened 2017-06-17
fix(Presence): reduce GetComponent calls in body physics update

thestonefox:
The BodyPhysics script was doing a `GetComponent` call every
`FixedUpdate` when checking to see if the snap to floor was valid
by checking to see if the object being held was underneath the
headset.

The fix attempts to cache the check the first time around to reduce
any further unnecessary calls to `GetComponent` if the held object
has not changed.


WildStyle69:
Confirming this has helped from my side of things. Sorry for delay checking, had to recover my 65GB git repo last few days, so been a little busy.

WildStyle69:
Will test further over the weekend... thanks for the update!

PR #1305, opened 2017-06-20
fix(Utilities): new vive keyword for headset discovery in device finder

thestonefox:
The DeviceFinder script has a method for determining which headset is
being used. However, there seemed to be an issue where newer HTC Vives
would report their name as `Vive MV.` with a dot at the end rather than
the standard `Vive MV` which was reported by the original HTC Vive
consumer edition.

This fix just checks both values and assumes the headset is `ViveMV`
to prevent users with newer Vive headsets receiving a warning.


davelloyd:
I've just had a customer contact with a headset reporting itself as "Vive. MV"
Presumably a similar fix will work.

bddckr:
@davelloyd Already handled by the latest VRTK master:
https://github.com/thestonefox/VRTK/blob/3c38c7cc4e34bd74601d8a1be57ce4b632aa26d8/Assets/VRTK/Scripts/Utilities/VRTK_DeviceFinder.cs#L391-L392

PR #1310, opened 2017-06-22
Fix/pre release fixes

thestonefox:


virror:
Sim fixes looks good, i have not tested it but i assume you have : p

PR #1312, opened 2017-06-22
feat(Locomotion): rename force teleport to teleport and add new method

thestonefox:
The `ForceTeleport` method has now been renamed to `Teleport` and a
new `ForceTeleport` method has been added that simply takes a
position and a rotation and sets the player location to those given
parameters.


virror:
Looking great!

PR #1315, opened 2017-06-24
feat(DesktopCamera): render desktop-only camera while in VR

bddckr:
There's a need to show something different on the desktop window
than what's shown in the VR headset. The potential use cases range
from offering nicer spectator views to showing completely different
views for asymmetrical games.
In combination with VRTK's existing ObjectFollow scripts this allows
for trailer recordings in which the VR headset's movement is
smoothed.

Having a separated desktop-only camera allows for:
- additional overlays or effects
- render other layers in addition/only
- changing the FOV
- ...

This change adds the new `VRTK_DesktopCamera` script and a
corresponding prefab that shows how to use the optional support for
rendering the headset's view in addition to the desktop-only view.

### DIFF
--- a/Assets/VRTK/Prefabs/DesktopCamera.prefab
+++ b/Assets/VRTK/Prefabs/DesktopCamera.prefab
@@ -1,0 +1,404 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!1001 &100100000
+Prefab:
+  m_ObjectHideFlags: 1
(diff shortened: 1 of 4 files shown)


bddckr:
I'm looking for a better name for this: What do you guys think about `SecondaryCamera`?

ciwolsey:
Good idea! Would be really handy.

thestonefox:
Are you happy to keep this called `DesktopCamera` ? if you are i'll merge it

PR #1317, opened 2017-06-24
fix(SDKManager): deprecate persist on load setting - fixes #1316

### LINKED ISSUE #1316: VRTK_SDKManager.persistOnLoad considered harmful
# What?!

`VRTK_SDKManager.persistOnLoad` is a nice shortcut to `DontDestroyOnLoad` the game object the SDK Manager is on.

When Unity introduced [Multi Scene Editing](https://docs.unity3d.com/Manual/MultiSceneEditing.html) usage of [`Object.DontDestroyOnLoad`](https://docs.unity3d.com/ScriptReference/Object.DontDestroyOnLoad.html) has kinda been deprecated, as we can see in the first link:

> It is recommended to avoid using DontDestroyOnLoad to persist manager GameObjects that you want to survive across scene loads. Instead, create a manager scene that has all your managers and use [`SceneManager.LoadScene(<path>, LoadSceneMode.Additive)`](https://docs.unity3d.com/ScriptReference/SceneManagement.SceneManager.LoadScene.html) and [`SceneManager.UnloadScene`](https://docs.unity3d.com/ScriptReference/SceneManagement.SceneManager.UnloadScene.html) to manage your game progress.

With `VRTK_SDKManager.persistOnLoad` we now hit one of the reasons for that recommendation:
When the `DontDestroyOnLoad` call is done Unity moves the game object (and its children) to a new scene called `DontDestroyOnLoad`:

> In Playmode, whith [_sic_] multiple scenes in the Hierarchy, an additional scene will show up called DontDestroyOnLoad.
> 
> Prior to Unity 5.3, any objects you would instantiate in Playmode marked as “DontDestroyOnLoad” would still show up in the hierarchy. These objects are not considered part of any scene but for Unity to still show the objects, and for you to inspect them, these objects are now shown as part of the special DontDestroyOnLoad scene.
> 
> You do not have access to the DontDestroyOnLoad scene and it is not available at runtime.

The last sentence is important here, because we found out even a simple call like 
```C#
gameObjectThatIsInDontDestroyOnLoadScene.GetComponent<ComponentThatIsOnThatGameObject>()
```
results in `null`, despite that component actually being on the game object!

# So how do we fix this?

We follow Unity's suggestion, of course! That is: Using additive scenes instead.

**Because we know additive scenes are probably something that is still new to tons of our users we want to emphasize: The easiest solution for now is to just use the same SDK Manager & SDK Setups in _every_ scene. This means duplicated game objects of course, but it's easier than using additive scenes.**

Let's see how we can use additive scenes then:
- Move SDK Manager with SDK Setups into a new (empty) scene.
- Load that scene additively (using `LoadSceneMode.Additive`).
- Load any other scene additively, too.
- If there is a need to unload a scene do so via `SceneManager.UnloadScene`.

bddckr:
When using the persist on load setting on the SDK Manager it is
moved to the `DontDestroyOnLoad` scene at runtime by Unity. That
scene is inaccessible which means there's no way to find objects in
that scene. Because of the issues related to looking up objects in
the inaccessible scene the persist on load setting on the SDK
Manager is now deprecated.

It's recommended to use additive scene loading instead. To reduce
the likelihood of these issues breaking the project for users
already using the setting this fix also makes sure to only populate
the SDK Setup object references when this setting is *not* checked.

### DIFF
--- a/Assets/VRTK/Scripts/Utilities/SDK/VRTK_SDKManager.cs
+++ b/Assets/VRTK/Scripts/Utilities/SDK/VRTK_SDKManager.cs
@@ -152,7 +152,8 @@ namespace VRTK
         }
         private static VRTK_SDKManager _instance;
 
-        [Tooltip("If this is true then the instance of the SDK Manager won't be destroyed on every scene load.")]
+        [Tooltip("**OBSOLETE. STOP USING THIS ASAP!** If this is true then the instance of the SDK Manager won't be destroyed on every scene load.")]
+        [Obsolete("`VRTK_SDKManager.persistOnLoad` has been deprecated and will be removed in a future version of VRTK. See https://github.com/thestonefox/VRTK/issues/1316 for details.")]
         public bool persistOnLoad;
 
         [Tooltip("Determines whether the scripting define symbols required by the installed SDKs are automatically added to and removed from the player settings.")]
@@ -610,7 +611,9 @@ namespace VRTK
 
         private void OnDisable()
         {
+#pragma warning disable 618
             if (_instance == this && !persistOnLoad)
+#pragma warning restore 618
             {
                 UnloadSDKSetup();
             }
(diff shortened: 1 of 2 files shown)


thestonefox:
I think we should push this onto master as it's a fix and isn't causing any new feature

PR #1319, opened 2017-06-25
feat(SDKObjectAlias): add component copying

bddckr:
The SDK Object Alias script now can copy a list of components over
to the SDK Object in addition to the existing childing.

### DIFF
--- a/Assets/VRTK/Scripts/Utilities/VRTK_SDKObjectAlias.cs
+++ b/Assets/VRTK/Scripts/Utilities/VRTK_SDKObjectAlias.cs
@@ -2,6 +2,9 @@
 namespace VRTK
 {
     using UnityEngine;
+    using System;
+    using System.Collections.Generic;
+    using System.Reflection;
 
     /// <summary>
     /// The GameObject that the SDK Object Alias script is applied to will become a child of the selected SDK Object.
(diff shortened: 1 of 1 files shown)


bddckr:
I'm kinda torn on this.

I tested by adding a Bloom effect to the game object the SDK Object Alias script is on. That results in a camera being added which we probably _don't_ want to copy over 😛
That's pretty much the reason I scrapped my initial implementation that just copied over every component of the game object (except the Transform and SDKObjectAlias of course).

Another issue are things like ParticleSystem: That one will always throw an error because we set all the fields and properties to be the same after copying the component. But ParticleSystem doesn't like us changing the `randomSeed` after the ParticleSystem is running already.
I guess this can be worked around by turning off the ParticleSystem and turning it on later, after the component is copied over?

Almost feels like one always has to write a helper script (turning off/on ParticleSystems, adjusting a setting, i.e. making sure something is copied over correctly)... One could just do the initial component add and configuration in a script then and no one would need this PR.

Thoughts?

thestonefox:
i wonder if for things that people need to apply directly to the camera in the editor they just have to apply those scripts to every camera rig they're supporting.

bddckr:
Yes that works of course, it means duplicated stuff everywhere, though. And you can't prefab-ify this for the same reasons we need this copying mechanic in the first place.

thestonefox:
I just think the more we try and consolidate the SDK camera rigs, the more hassle we're getting in to. It's like they've been designed to not work together! :p

bddckr:
Something something take over ownership of the VR camera something. Not possible for every SDK I'd guess.

thestonefox:
One day all headsets will be supported through Direct Reality/OpenXR/Unity/Whatever

and we never have to worry about multiple SDK set ups ever again! 😄

bddckr:
You just named 4 different ones. These will be the new things we have to support instead of the current SDKs 😝

thestonefox:
What are your thoughts on this one still? Is it still worth while having the component copy? or is it cludgy?

bddckr:
It works for the original thing we wanted it for: For copying over camera image effect stuff.

PR #1327, opened 2017-06-29
feat(sdk): implement support for Ximmerse Flip SDK Cardboard

tazzkiller:
Ximmerse has release a SDK for the Flip Controller ( https://github.com/Ximmerse/SDK_Flip ) but is not compatible with VRTK. You think poqqible to add comaptibilty with VRTK and Cardboard for unity ?


tazzkiller:
sorry bad manipulation, i move it to issues for request

PR #1332, opened 2017-06-30
Fix/attach references - fixes #1330

### LINKED ISSUE #1330: VRTK_SDK Manager/SDKSetupSwitcher - Simulator loaded first - Causes issues with SteamVR and grabbed objects
### Environment

 * Github
 * 3.2.0 and 7490bea
* SteamVR version 1.2.1 and 1.2.2
 * Unity 5.6.1f1
 * Simulator and Vive
 * Simulator and then switched to SteamVR using SDKSetupSwitcher

### Steps to reproduce

1) Load up example scene 008 (or any scene that has a grabbable object and the SDKSetupSwitcher)
2) In the SDK Manager, move the Simulator rig to the top position so it loads first
3) After starting the scene, switch to SteamVR
4) Try to grab a grabbable object.  Sometimes this glitch happens only with one controller (and it could be left or right), other times it happens with both

### Expected behavior

The grabbed object will attach to the controller and behave like it would when SteamVR is loaded first

### Current behavior

The grabbed object teleports to a random position and does not follow any controller movements when "grabbed"

Here are a couple of videos explaining the issue: 

1) Scene 008 - SteamVR 1.2.1 - https://streamable.com/q02l0
2) Empty scene with just a grabbable cube, plane and the SDK setups - SteamVR 1.2.2 - https://streamable.com/8pj3a

The videos were made in separate projects

thestonefox:


thestonefox:
Another potential option would be to listen to the is render model available stuff in steamvr.

But how do we abstract that into a generic SDK event? at the moment the SDK classes are static methods and they don't re-emit any events.

Assets/VRTK/Scripts/Interactions/VRTK_InteractGrab.cs
```diff
@@ -316,17 +336,32 @@ protected virtual void SetControllerAttachPoint()
...
-                    }
+        protected virtual IEnumerator SetCurrentControllerAttachPoint(string searchPath, int attempts, float delay)
+        {
+            Transform defaultAttachPoint = controllerReference.model.transform.Find(searchPath);
+
+            while (defaultAttachPoint == null && attempts > 0)
+            {
+                defaultAttachPoint = controllerReference.model.transform.Find(searchPath);
+                attempts--;
+                yield return new WaitForSeconds(delay);
```
bddckr: Move this _outside the while loop_, to reduce the number of allocations of `WaitForSeconds`.

thestonefox: what was your suggestion for this again about sticking something something in variables or something?

bddckr: Well you said that moving the `new WaitForSeconds(delay)` to be outside the inner loop doesn't work, which I still find strange, but didn't test.
So I then recommended moving that creation outside the inner loop, but saving it into a private class variable, to workaround the issue you seem to have there.

thestonefox: Oh I thought you were talking about the `defaultAttachPoint` assignment.

If the yield return is outside the while loop then it won't pause for the delay before trying the while loop again.

I don't want it to do x number of attempts then wait and try again.

I want it to do 1 attempt, wait a bit, the try again until all of the max attempts have been used up

bddckr: I only mean to move the actual C# object instantiation to be outside the loop. Create it once with `new` into a local variable, then inside the loop do `yield return theVariable;`. :smile:

thestonefox: write the code how you see it working, i still have no idea what you're suggesting :p

PR #1334, opened 2017-07-01
fix(Presence): calculate collider height based on global positions - fixes #1325

### LINKED ISSUE #1325: Body collider hight for Simulator calculated wrongly
After the addition of the Neck for the Simulator the body collider height now gets calculated incorrectly.
This can be seen very easy by just playing using the Simulator and bodyPhysics.

The fix is to change one line in VRTK_BodyPhysics.cs in the UpdateCollider method:
Old:
float newpresenceColliderYSize = (headset ? headset.transform.localPosition.y - (headsetYOffset + CalculateStepUpYOffset()) : 0f);
New:
float newpresenceColliderYSize = (headset ? (headset.position.y - playArea.position.y) - (headsetYOffset + CalculateStepUpYOffset()) : 0f);

This will calculate the height based on the difference of the headset and play area positions instead of just using the localPosition. This should also be a safer method overall since it does not rely on the assumption that the headset Object is always a direct child of the playArea.

thestonefox:
There was an issue with the Body Physics body collider being sized
incorrectly in the simulator because the calculation was relying
on local positions and the simulator has a more complex set up
with the new neck object.

This fix uses global positions and is more generic to provide a
better future proof solution.

Thanks to @virror for the fix.

### DIFF
--- a/Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
+++ b/Assets/VRTK/Scripts/Presence/VRTK_BodyPhysics.cs
@@ -1095,7 +1095,7 @@ namespace VRTK
         {
             if (bodyCollider != null && headset != null)
             {
-                float newpresenceColliderYSize = (headset ? headset.transform.localPosition.y - (headsetYOffset + CalculateStepUpYOffset()) : 0f);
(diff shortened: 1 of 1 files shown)


virror:
Looks like you have double checks for headset != null?

virror:
lgtm now : )

PR #1339, opened 2017-07-04
fix(UpdatePrompt): long changelog error

bddckr:
Long changelogs resulted in Unity logging an error. The fix is to
add pagination to the changelog so less text is rendered at one
point.

Additionally the changelog is updated to follow the markdown look a
bit more.


bddckr:
unity_2017-07-04_21-43-57

PR #1347, opened 2017-07-07
fix(Interaction): ensure interactable object rigidbody is set - fixes #1340

### LINKED ISSUE #1340: Snap Drop Zone "Default Snapped Object" loses kinematic state on first grab
### Environment

 * Source of VRTK: Github master
 * Version of VRTK:  9d09301
 * Version of Unity3D: Unity 5.5.3
 * Hardware used: Vive
 * SDK used: SteamVR

### Steps to reproduce

 * Open Example 041
 * Set the rigidbody to kinematic on `FlyingSaucer_HightlightOnTouch`
 * Set the `Default Snapped Object` on `Saucer_SnapDropZone` to `FlyingSaucer_HightlightOnTouch`
 * Start the scene
 * The saucer should be snapped automatically to the drop zone
 * Grab the saucer from the drop zone and drop the object

### Expected behavior

The saucer should float in the air as it is kinematic

### Current behavior

The saucer drops to the ground as kinematic has been toggled off. This may be due to the `SaveCurrentState` isn't being called because no grab action happens, therefore the load previous state is resetting the kinematic state to false.

thestonefox:
There was an issue with the Interactable Object attempting to set
the rigidbody variable. As it was set in the `Awake` method and
only attempted to be set once, there could be times when the rigidbody
was checked for existence and it didn't exist.

This was present in the Snap Drop Zone if an Interactable Object was
snapped to a drop zone on scene start and the Interactable Object's
rigidbody was set to kinematic. The part of the Interactable Object
that saves the state of the rigidbody wasn't being run because
the rigidbody hadn't been set at that time, even though it was called
in `Awake`.

The fix now implements a new method that checks to see if the rigidbody
has been set and if not attempts to set it before returning if it
is valid or not.

### DIFF
--- a/Assets/VRTK/Scripts/Interactions/VRTK_InteractableObject.cs
+++ b/Assets/VRTK/Scripts/Interactions/VRTK_InteractableObject.cs
@@ -113,6 +113,7 @@ namespace VRTK
         [Tooltip("Determines which controller can initiate a use action.")]
         public AllowedController allowedUseControllers = AllowedController.Both;
 
+        #region Interaction Events
         /// <summary>
         /// Emitted when another object touches the current object.
         /// </summary>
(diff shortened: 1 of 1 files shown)


thestonefox:
I don't like this solution

PR #1350, opened 2017-07-09
fix(Examples): remove unused variable to resolve warning

bddckr:
There was an unused variable introduced in the previous capsense
axis commit. The fix is to remove said variable.


bddckr:
Note: This targets the release branch because the issue is only apparent in that one. The master branch doesn't have the issue.

PR #1356, opened 2017-07-11
feat(SDK): implement internal Unity VR support

thestonefox:
Unity supports basic VR functionality without the need for any
third party SDK. A new SDK bridge has been added that hooks into
this internal Unity VR support so VRTK can be used for any device
that Unity supports.

The InputManager has been update to have the default axis information
required for the axes on the controllers to work correctly.


thestonefox:
There are still issues with this PR.

The velocity and angular velocity of the controllers isn't working correctly.

There seems to be a lot of lag when attempting to check the grip button press.

PR #1359, opened 2017-07-12
feat(SDKManager): simplify single SDK Setup configurations

bddckr:
Supporting multiple SDKs means VRTK had to disable all the SDK 
Setups that are used automatically to ensure they're usable for
runtime switching. This change allows to completely bypass the
runtime loading and that requirement in case only one SDK Setup is
used. This means users of VRTK don't have to bother with handling
when an SDK Setup is loaded and usable, nor do they have to
re-enable their SDK Setups game objects again to work on them in the
editor while they're visible.

### DIFF
--- a/Assets/VRTK/Editor/VRTK_SDKSetupEditor.cs
+++ b/Assets/VRTK/Editor/VRTK_SDKSetupEditor.cs
@@ -432,37 +432,39 @@
 
             private static void FixOpenAndUnsavedScenes()
             {
-                for (int index = 0; index < EditorSceneManager.loadedSceneCount; index++)
-                {
-                    Scene scene = SceneManager.GetSceneAt(index);
-                    if (scene.isLoaded && scene.isDirty)
-                    {
-                        VRTK_SDKSetup[] setups = FindObjectsOfType<VRTK_SDKSetup>();
-                        foreach (VRTK_SDKSetup setup in setups)
-                        {
-                            setup.gameObject.SetActive(false);
-                        }
 
-                        if (setups.Length != 0)
-                        {
-                            string infoMessage = string.Format("The following game objects have been set inactive to allow for SDK loading and switching using the SDK Setups on them:\n{0}", string.Join(", ", setups.Select(setup => setup.name).ToArray()));
-                            if (EditorApplication.isPlayingOrWillChangePlaymode)
-                            {
+                List<VRTK_SDKSetup> setups = Enumerable.Range(0, EditorSceneManager.loadedSceneCount)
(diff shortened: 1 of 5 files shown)


thestonefox:
Could this be merged into master?

bddckr:
After tons of testing, yup. _changes target branch_

thestonefox:
Hehe because you branched from release, your commit is now full of commits.

You probably need to branch from master, then cherry pick the commit into that branch (or just copy and paste the changes over!)

bddckr:
Yeah I was just too lazy to stash my current work just yet.. On it!

thestonefox:
Still getting an error when using the simulator.

Steps to reproduce:

 * Open Example scene 3
 * Delete all SDK setups except the Simulator
 * Update the SDK Manager to only include the Simulator
 * Run the scene
 * Press `Space` to move to next scene
 * Press `Backspace` to move back to previous scene
 * Get the error: `NullReferenceException: Object reference not set to an instance of an object VRTK.SDK_SimController.IsButtonPressed (UInt32 index, ButtonPressTypes type, KeyCode button) (at Assets/VRTK/SDK/Simulator/SDK_SimController.cs:550)`

Looks like it's being caused by an issue in `SDK_SimController` on line `543` and `550` where it's trying to access data on objects that don't exist.

This seems to be a valid fix:

```
            if (index == 1)
            {
                if (rightController == null || !rightController.Selected)
                {
                    return false;
                }
            }
            else if (index == 2)
            {
                if (leftController == null || !leftController.Selected)
                {
                    return false;
                }
            }
            else
            {
                return false;
            }
```

thestonefox:
Another issue (which i'm less bothered about)

 * Open up example scene (lets say 007)
 * Delete all SDKs except SteamVR
 * Update the SDK Manager to only include the SteamVR
 * Make sure the SteamVR SDK object is an active game object in the scene
 * Run the scene
 * Switch to next scene pressing `Space`
 * On next scene (008) switch the SDK to Simulator
 * Now press `Backspace` to go back to previous scene (007)
 * It will attempt to remember you were on the Simulator but the simulator doesn't exist so it switches to SteamVR as it's the only one available in the scene.
 * You'll get a bunch of `Assertion failed on expression: 'IsMatrixValid(matrix)'` errors spammed in the console

bddckr:
Should be fixed now.

thestonefox:
@bddckr all seems to be working ok for me now, I did another merge test and it merged into release ok as well.

Happy for me to merge this PR?

bddckr:
@thestonefox Yes, looks good to me now, too. :shipit:

PR #1370, opened 2017-07-15
fix(Locomotion): set position offset with rotation on dash teleport

thestonefox:
The DashTeleport script was putting the play area into the incorrect
position if dash also had a rotation applied because the rotation
would affect the headset compensation.

This fix recalculates the position based on the headset offset and
the target rotation.


thestonefox:
Superseded by https://github.com/thestonefox/VRTK/pull/1405

PR #1371, opened 2017-07-16
feat(Interaction): emit event when controller type is available

thestonefox:
The Controller Type is now emitted via a new event on the
TrackedController script (which is also re-thrown by the
ControllerEvents script). This event is emitted when the type
of connected controller is known.

Previously, a number of scripts would duplicate the same coroutine
for determining when the controller type was known. This new event
removes the need for this duplication as these scripts can just
listen for this event.

The `GetCurrentControllerType` method now accepts a ControllerReference
to determine specific connected controller type. If `null` is passed
then it will attempt to get the right controller first then the left
if the right is not available.


thestonefox:
superseded by https://github.com/thestonefox/VRTK/pull/1417

PR #1373, opened 2017-07-16
feat(SDK): support GearVR input

thestonefox:
The GearVR HMD input and controller input are now supported.
Along with basic support for the Oculus remote.

### DIFF
--- a/Assets/VRTK/SDK/Oculus/SDK_OculusController.cs
+++ b/Assets/VRTK/SDK/Oculus/SDK_OculusController.cs
@@ -22,8 +22,8 @@ namespace VRTK
         protected SDK_OculusBoundaries cachedBoundariesSDK;
         protected VRTK_TrackedController cachedLeftController;
         protected VRTK_TrackedController cachedRightController;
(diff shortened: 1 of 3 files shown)


thestonefox:
[] Test with GearVR HMD and Controller
[] Test with Oculus Rift and Oculus Touch
[] Test with Oculus Rift and Oculus Remote

Assets/VRTK/SDK/Oculus/SDK_OculusController.cs
```diff
@@ -728,21 +740,52 @@ protected virtual bool IsButtonPressed(uint index, ButtonPressTypes type, OVRInp
...
+            switch (GetCurrentControllerType())
+            {
+                case ControllerType.Oculus_OculusTouch:
+                case ControllerType.Oculus_OculusGamepad:
+                    return thumbsticks[index];
+            }
+            return touchpads[index];
+        }
+
+        protected virtual OVRInput.Button GetTouchpadButtonMask()
```
thestonefox: Button mask needs to take the touch type, at the moment it only returns Button press and does not return Touch state

PR #1390, opened 2017-07-28
feat(Locomotion): jumping mechanic using controller slingshot

mattboy64:
Allows the player to precisely jump by pulling the trigger on
bother controllers, pulling back the desired amount, and releasing.
The system combines the vectors and directions of both controllers
to sling the player in the defined direction.

### DIFF
--- a/Assets/VRTK/Scripts/Locomotion/VRTK_SlingshotJump.cs
+++ b/Assets/VRTK/Scripts/Locomotion/VRTK_SlingshotJump.cs
@@ -1,0 +1,329 @@
+﻿// Slingshot Jump|Locomotion|20120
+namespace VRTK
+{
+    using System;
+    using GrabAttachMechanics;
+    using UnityEngine;
+
+    /// <summary>
+    /// Event Payload
+    /// </summary>
+    /// <param name="target">The GameObject of the interactable object that is being interacted with by the controller.</param>
+    public struct SlingshotJumpEventArgs
+    {
+        public GameObject target;
+    }
+
+    /// <summary>
+    /// Event Payload
+    /// </summary>
+    /// <param name="sender">this object</param>
+    /// <param name="e"><see cref="PlayerClimbEventArgs"/></param>
+    public delegate void SlingshotJumpEventHandler(object sender, SlingshotJumpEventArgs e);
+
+    /// <summary>
+    /// Slingshot Jump allows player jumping based on the direction and amount pulled back of each controller. This slingshots the player in defined direction and speed.
+    /// </summary>
+    /// <example>
+    /// `VRTK/Examples/037_CameraRig_ClimbingFalling` shows how to set up a scene with slingshot jumping. This script just needs to be added to the PlayArea object and the requested forces and buttons set.
+    /// </example>
+    [AddComponentMenu("VRTK/Scripts/Locomotion/VRTK_SlingshotJump")]
+    public class VRTK_SlingshotJump : MonoBehaviour
+    {
+        [Header("SlingshotJump Settings")]
+
+        [Tooltip("How close together the trigger releases have to be to initiate a jump.")]
+        public float releaseWindowTime = 0.5f;
+        [Tooltip("Multiplier that increases the jump strength.")]
+        public float velocityMultiplier = 5.0f;
+        [Tooltip("The maximum velocity a jump can be.")]
+        public float velocityMax = 8.0f;
+
+        [Tooltip("This button will cancel an already tensioned sling shot.")]
+        [SerializeField]
+        protected VRTK_ControllerEvents.ButtonAlias cancelButton = VRTK_ControllerEvents.ButtonAlias.Undefined;
+        [Tooltip("This button will start the sling shot move.")]
+        [SerializeField]
+        protected VRTK_ControllerEvents.ButtonAlias activationButton =  VRTK_ControllerEvents.ButtonAlias.TriggerPress;
+        [Tooltip("The VRTK_BodyPhysics object used on the player. If the script is being applied onto an object that already has a VRTK_BodyPhysics component, this parameter can be left blank as it will be auto populated by the script at runtime.")]
+        [SerializeField]
+        protected VRTK_BodyPhysics bodyPhysics;
+
+        /// <summary>
+        /// Emitted when a slingshot jump occurs
+        /// </summary>
+        public event SlingshotJumpEventHandler SlingshotJumped;
+
+        protected Transform playArea;
+        protected VRTK_PlayerClimb playerClimb;
+
+        protected Vector3 leftStartAimPosition;
+        protected Vector3 leftReleasePosition;
+        protected bool leftIsAiming;
+
+        protected Vector3 rightStartAimPosition;
+        protected Vector3 rightReleasePosition;
+        protected bool rightIsAiming;
+
+        protected VRTK_ControllerEvents leftControllerEvents;
+        protected VRTK_ControllerEvents rightControllerEvents;
+
+        protected VRTK_InteractGrab leftControllerGrab;
+        protected VRTK_InteractGrab rightControllerGrab;
+
+        protected bool leftButtonReleased;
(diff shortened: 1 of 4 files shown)


mattboy64:
This is basically climbey style jumping.

climbey

paulstraw:
Confirmed working with Rift+Touch in Unity 2017.1.0f3. This feels really good. 👏

mikepangel:
Also confirmed working with Vive in Unity 2017.1.0f3. Feels great, very smooth and allows you to grip the movement that feels right.

However, the code is constantly checking in Update() for controller events instead of registering to respond to them, so this could use a bit more work to be lighter on runtime.

mattboy64:
@mikepangel thank you for the suggestion. The system has been re-written to only use events now. @thestonefox VRTK_SlingshotJump.cs has been considerably changed and could use another once over.

Assets/VRTK/Scripts/Locomotion/VRTK_SlingshotJump.cs
```diff
@@ -0,0 +1,283 @@
...
+    public delegate void SlingshotJumpEventHandler(object sender, SlingshotJumpEventArgs e);
+
+    /// <summary>
+    /// Slingshot Jump allows player jumping based on the direction and amount pulled back of each controller. This slingshots the player in defined direction and speed.
+    /// </summary>
+    /// <example>
+    /// `VRTK/Examples/037_CameraRig_ClimbingFalling` shows how to set up a scene with slingshot jumping. This script just needs to be added to the PlayArea object and the requested forces and buttons set.
+    /// </example>
+    [AddComponentMenu("VRTK/Scripts/Locomotion/VRTK_SlingshotJump")]
+    [RequireComponent(typeof(VRTK_BodyPhysics))]
```
thestonefox: My one concern here is that this will force the script to be on the same GameObject as the BodyPhysics script.

Other scripts have now moved away from using `RequireComponent` and have a public parameter (usually under a header of `Custom Settings`) that allows you to inject the required script into the component.

E.g. a param called `bodyPhysics` and if you leave it blank then it will attempt to look for the BodyPhysics script on Component or Parent. If it can't find a body physics script after all that then it's fine to throw an error (and make sure all references to the body physics script are null checked).

Assets/VRTK/Scripts/Locomotion/VRTK_SlingshotJump.cs
```diff
@@ -0,0 +1,283 @@
...
+    public class VRTK_SlingshotJump : MonoBehaviour
+    {
+        [Tooltip("This button will cancel an already tensioned sling shot.")]
+        public VRTK_ControllerEvents.ButtonAlias cancelButton = VRTK_ControllerEvents.ButtonAlias.Undefined;
+        [Tooltip("This button will start the sling shot move.")]
+        public VRTK_ControllerEvents.ButtonAlias triggerButton =  VRTK_ControllerEvents.ButtonAlias.TriggerPress;
```
thestonefox: would this be better renamed to like `activationButton` as `triggerButton` may be confusing with actually being the trigger button even though it can be set to anything else?

Assets/VRTK/Scripts/Locomotion/VRTK_SlingshotJump.cs
```diff
@@ -0,0 +1,283 @@
...
+        {
+            bodyPhysics = GetComponent<VRTK_BodyPhysics>();
+            playerClimb = GetComponent<VRTK_PlayerClimb>();
+
+            VRTK_SDKManager.instance.AddBehaviourToToggleOnLoadedSetupChange(this);
+        }
+
+        protected virtual void OnEnable()
+        {
+            InitListeners(true);
```
thestonefox: One thing to note here, is if the button settings are changed at runtime, the only way to get the listeners to update programatically would be to disable the script, change the button settings, then re-enable the script.

Alternatively you could cache the previous button settings, then in a LateUpdate check if they've been changed and unregister the old buttons and re-register the new buttons?

Assets/VRTK/Scripts/Locomotion/VRTK_SlingshotJump.cs
bddckr: ([`VRTK_Pointer` names it `activationButton`, too.](https://github.com/thestonefox/VRTK/blob/master/Assets/VRTK/Scripts/Pointers/VRTK_Pointer.cs#L27-L28))

Assets/VRTK/Scripts/Locomotion/VRTK_SlingshotJump.cs
bddckr: (Example [here](https://github.com/thestonefox/VRTK/blob/master/Assets/VRTK/Scripts/Interactions/VRTK_InteractGrab.cs#L185) and usage example [here](https://github.com/thestonefox/VRTK/blob/master/Assets/VRTK/Scripts/Interactions/VRTK_InteractGrab.cs#L94).)

mattboy64: This has been changed to an optional parameter.

Assets/VRTK/Scripts/Locomotion/VRTK_SlingshotJump.cs
mattboy64: Changed to activationButton.

Assets/VRTK/Scripts/Locomotion/VRTK_SlingshotJump.cs
mattboy64: I added a SetActivationButton() and SetCancelButton() function that can be called from code that will reset the button listeners.

thestonefox:
Just noticed your Assets are set to binary serialisation so the example scene is a binary file. It needs to be text serialised

Assets/VRTK/Scripts/Locomotion/VRTK_SlingshotJump.cs
```diff
@@ -0,0 +1,311 @@
...
+
+        protected bool leftButtonReleased;
+        protected bool rightButtonReleased;
+        protected float countDownEndTime;
+
+        /// <summary>
+        /// The SetActivationBUtton method sets the button used to activate a slingshot jump.
+        /// </summary>
+        /// <param name="button">The controller button to use to activate the jump.</param>
+        public virtual void SetActivationBUtton(VRTK_ControllerEvents.ButtonAlias button)
```
thestonefox: `Button` not `BUtton` ;)

Assets/VRTK/Scripts/Locomotion/VRTK_SlingshotJump.cs
```diff
@@ -0,0 +1,311 @@
...
+        [Tooltip("How close together the trigger releases have to be to initiate a jump.")]
+        public float releaseWindowTime = 0.5f;
+        [Tooltip("Multiplier that increases the jump strength.")]
+        public float velocityMultiplier = 5.0f;
+        [Tooltip("The maximum velocity a jump can be.")]
+        public float velocityMax = 8.0f;
+
+        [Tooltip("This button will cancel an already tensioned sling shot.")]
+        [SerializeField]
+        protected VRTK_ControllerEvents.ButtonAlias cancelButton = VRTK_ControllerEvents.ButtonAlias.Undefined;
```
thestonefox: initially i liked this concept of having these fields private then having the setter.

But, what if you wanted to query the script to find out what button was being used? afaik a serialised protected field won't be exposed to other classes at run time?

mattboy64: Hmmm, good point. I was thinking of using properties, but those don't show up in the inspector. I can add some Get/Set functions instead. I just want to avoid people setting these directly and wondering why it's not working.

mattboy64:
The example scene was changed to text mode and the casing has been fixed. Let me know if there are any other issues.

PR #1403, opened 2017-08-07
fix(pointers): Preserve last known angle for direction indicator.

insominx:
Made the direction indicator only update when there is valid input rather than snapping back to zero when letting go of the analog stick.  Preserving the last known angle seems like the expected behavior.

### DIFF
--- a/Assets/VRTK/Prefabs/Resources/Scripts/VRTK_PointerDirectionIndicator.cs
+++ b/Assets/VRTK/Prefabs/Resources/Scripts/VRTK_PointerDirectionIndicator.cs
@@ -112,7 +112,7 @@ namespace VRTK
 
         protected virtual void Update()
         {
-            if (controllerEvents != null)
+            if (controllerEvents != null && controllerEvents.touchpadTouched && controllerEvents.touchpadAxisChanged)
             {
                 float touchpadAngle = controllerEvents.GetTouchpadAxisAngle();
                 float angle = ((touchpadAngle > 180) ? touchpadAngle -= 360 : touchpadAngle) + headset.eulerAngles.y;


GitCop:

 * Commit: df509de18705d93ff8b38767240064dee808c30e

---

bddckr:
You have to update your commit, the PR title doesn't count.

Amend the latest commit to change the commit message, then force push your PR branch. That will update this PR here on GitHub.

Assets/VRTK/Prefabs/Resources/Scripts/VRTK_PointerDirectionIndicator.cs
```diff
@@ -112,7 +112,7 @@ protected virtual void Awake()
 
         protected virtual void Update()
         {
-            if (controllerEvents != null)
+            if (controllerEvents != null && controllerEvents.touchpadTouched && controllerEvents.touchpadAxisChanged)
```
thestonefox: doing the ` && controllerEvents.touchpadAxisChanged` causes the pointer direction indicator to not update it's position when you release and then re-touch the touchpad which is incorrect behaviour

thestonefox:
Why do you need to know the last known angle?

thestonefox:
You only really need to do the `touchpadTouched` check like in this PR:

https://github.com/thestonefox/VRTK/pull/1407

insominx:
@thestonefox I did initially try what you have you #1407.  However the results weren't really stable enough.  It seemed as though there were cases where letting go of the analog stick would rebound in the opposite direction.  Notice the twitchiness in this video:

[https://www.youtube.com/watch?v=-5uMCcGH6-o&feature=youtu.be](url)

Assets/VRTK/Prefabs/Resources/Scripts/VRTK_PointerDirectionIndicator.cs
insominx: wouldn't retouching it cause an axis change too?

GitCop:

 * Commit: df509de18705d93ff8b38767240064dee808c30e

 * Commit: df2053c9f675e37b51c6cbf3845152872d62b4e5

---

GitCop:

 * Commit: df509de18705d93ff8b38767240064dee808c30e

 * Commit: df2053c9f675e37b51c6cbf3845152872d62b4e5

---

thestonefox:
@insominx your fix of checking the touchpadAxisChange doesn't work on the vive.

I will attempt to recreate the issue you're having from the video you posted on the rift and look for a better solution.

thestonefox:
@insominx Can you try the fix on this PR

https://github.com/thestonefox/VRTK/pull/1407/commits/2a027d53c4899e6b45937c80d0943206ccf34422

and see if it works for you

insominx:
@thestonefox This is an improvement but it still suffers from the case when you let go of the analog stick and then you subsequently touch it without moving the stick right away which causes it to reset to 0.  This makes it look jumpy because it goes back to 0 angle before returning to the angle you are beginning to press toward.  Maybe a solution would be to have a "dead zone" where if the distance from center isn't great enough, then you don't register a new angle.

thestonefox:
@insominx I'm not sure a dead zone would work for the vive wand, because you can circle your finger around in the middle of the pad and still get a valid angle.

I'll play with it some more and get it working for rift too.

thestonefox:
@insominx ok I've had another go

https://github.com/thestonefox/VRTK/pull/1407/commits/42c09f9e99aaa6524b85847f054749a1bf526494

I've tested it on the oculus touch and it doesn't seem to snap back when you release because the axis will always be 0,0 on the touch when its released.

However, I think an optional deadzone setting for the pointer will also help, but I'll do that as a separate commit as it will need to go into the `release/3.3.0-alpha` branch.

insominx:
👍

thestonefox:
@insominx just confirming the new PR works for you now?

thestonefox:
superseded by https://github.com/thestonefox/VRTK/pull/1407

insominx:
@thestonefox Just tried it out and it looks good to me.

PR #1435, opened 2017-08-17
chore(SDK): provide more meaningful SDK pretty names

thestonefox:
The pretty names for each SDK platform and build target has been
updated to include the library and build target to make it easier
to see which SDK option is for which output device.


thestonefox:
@bddckr what do you think about these for names?

bddckr:
Oh nice, these offers all the info!

I just noticed that the name "standalone" is a stupid one from Unity, though I don't know whether I like "desktop" more...

thestonefox:
They call it `PC, Mac & Linux Standalone`. Desktop is a hard one because people may be like "yo brah it's a laptop tho"

bddckr:
Yeah go with `Standalone` then. Should be known by Unity using devs.

PR #1439, opened 2017-08-18
chore(Avatar): update avatar hand models

thestonefox:
The Avatar hand models have been updated with a lower poly count
version to make the performance better. The armature and animation
structure has also been improved.


thestonefox:
Link to the blend file for completeness https://www.dropbox.com/s/lqfvc47479fzpmr/VRTK_BasicHand_1.3.blend?dl=0

PR #1443, opened 2017-08-19
feat(ObjectLerp): Add object lerp script for interactable object grab - fixes #1433

### LINKED ISSUE #1433: New Feature: VRTK_ObjectLerp (lerp an VRTK_InteractableObject on grab).
**Overview**

General idea is to add new functionality that smoothly animates an object into position when you grab it. This will be an optional / extra script that you can attach to an  `VRTK_InteractableObject`, providing a slightly different mechanic and visual from the existing immediate grab.

**Code In-progress**

- Created a new script class called `VRTK_ObjectLerp` that smoothly handles the lerp position and rotation of an object over time.
- You can attach this script to an `VRTK_InteractableObject`, which then will set itself up to use the `VRTK_ObjectLerp`.
- When an object is grabbed by `VRTK_InteractGrab` this class checks on the grabbing `VRTK_InteractableObject` to see if there is a valid / configured `VRTK_ObjectLerp` available, and if the object is set to `lerpTransformOnGrab`.
- If the `VRTK_InteractableObject` is setup to lerp on grab, then rather than complete the grab as normal, the `VRTK_InteractGrab` method calls back to a new method on the `VRTK_InteractableObject` that will start the lerp to position, i.e. `StartLerpTransformOnGrab`.
- The `VRTK_InteractableObject` will then call the `VRTK_ObjectLerp` to start a coroutine for the lerp to position,  i.e. `DoObjectTranslation`.

**Not yet done, to be confirmed:**

- Once the `DoObjectTranslation` method completes, callback / notify the `VRTK_InteractableObject` that it's done.
- The `VRTK_InteractableObject` then calls the `VRTK_InteractGrab` script to grab the object again. Only this time it will tell the `VRTK_InteractGrab` somehow that we don't need to lerp, just do a normal / immediate grab.

**Other things to consider, as per our chat on Slack:**

- What to do if the object is ungrabbed mid-way, before completion:
  : Move the object back to the start position.
  : Or if it's pass the half way point complete the lerp.
  : Perhaps a combination, could have settings to specificy what you want to happen.
  
- From @bddckr: "Should work with every grab mechanic, but I think currently some of these things take at least 1 frame, so would be visible. Maybe instantiate a temporary copy and turn of visibility and collisions etc.":
  : Not sure I understand what this is referencing yet.. perhaps I'll see this issue once the base code I've explained above is working.

WildStyle69:
The Object Lerp script is usually applied to an Interactable
Object and provides a mechanism for moving the GameObject
in a lerped translation when grabbed.

### DIFF
--- a/Assets/VRTK/Examples/005_Controller_BasicObjectGrabbing.unity
+++ b/Assets/VRTK/Examples/005_Controller_BasicObjectGrabbing.unity
@@ -13,7 +13,7 @@ OcclusionCullingSettings:
 --- !u!104 &2
 RenderSettings:
   m_ObjectHideFlags: 0
-  serializedVersion: 7
+  serializedVersion: 8
   m_Fog: 0
   m_FogColor: {r: 0.5, g: 0.5, b: 0.5, a: 1}
   m_FogMode: 3
@@ -25,6 +25,7 @@ RenderSettings:
   m_AmbientGroundColor: {r: 0.047, g: 0.043, b: 0.035, a: 1}
   m_AmbientIntensity: 1
   m_AmbientMode: 0
+  m_SubtractiveShadowColor: {r: 0.42, g: 0.478, b: 0.627, a: 1}
   m_SkyboxMaterial: {fileID: 10304, guid: 0000000000000000f000000000000000, type: 0}
   m_HaloStrength: 0.5
   m_FlareStrength: 1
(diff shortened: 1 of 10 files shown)


WildStyle69:
@thestonefox -- this feature is working but not 100% complete. There are a number of points we should discuss and figure out the best way to handle. I've created a pull request now, so you can look at my direction and progress... to see if it's going to work out and send any feedback. Thanks!

WildStyle69:
Updated to handle `disableWhenIdle ` on `VRTK_InteractableObject`

thestonefox:

PR #1482, opened 2017-09-06
chore(Structure): support new XR naming convention in 2017.2

thestonefox:
Unity 2017.2 has changed all of the VR packages and VR class names
and renamed them to XR. To ensure future proofing, all of the
references to VR classes have been updated XR and then script
define symbols for any Unity version older than 2017.2 create
class aliases from the old VR convention to the new XR convention.

The VRSettings.renderScale method has been completely renamed so
it's not possible to alias it. In this case a new helper method
has been created in SharedMethods that allows the retrieval of
the renderScale and another method which sets it.

### DIFF
--- a/Assets/VRTK/Source/SDK/Unity/SDK_UnityControllerTracker.cs
+++ b/Assets/VRTK/Source/SDK/Unity/SDK_UnityControllerTracker.cs
@@ -2,7 +2,12 @@
 namespace VRTK
 {
     using UnityEngine;
+#if UNITY_2017_2_OR_NEWER
+    using UnityEngine.XR;
+#else
     using UnityEngine.VR;
+    using XRNode = UnityEngine.VR.VRNode;
+#endif
 
     /// <summary>
     /// The Controller Tracker enables the GameObject to track it's position/rotation to the available connected VR Controller via the `UnityEngine.VR` library.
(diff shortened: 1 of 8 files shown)


thestonefox:
@bddckr if you get chance, do you mind taking a look over this to see if I've missed anything or if anything could be done in a better way?

Assets/VRTK/Source/SDK/Unity/SDK_UnityControllerTracker.cs
```diff
@@ -2,7 +2,12 @@
 namespace VRTK
 {
     using UnityEngine;
+#if UNITY_2017_2_OR_NEWER
+    using UnityEngine.XR;
+#else
     using UnityEngine.VR;
```
bddckr: Sorry I have no time to test this by pulling down, but this looks suspicious: Either it's not needed and can be removed or it should be `using XYZ = UnityEngine.VR.XYZ;` like done everywhere else.

thestonefox: The only change that requires the alias is the XRNode.

The other VR class is TrackingSpace, but that hasn't changed between VR and XR

https://github.com/thestonefox/VRTK/blob/73abc8cb15f33a2a187f506a4f5505dbc3a489d0/Assets/VRTK/Source/SDK/Unity/SDK_UnityControllerTracker.cs#L60-L61

So just including the relevant package is enough for that one.

bddckr: I see. Why do they rename everything but don't rename that one. :/

thestonefox: Better still why is `UnityEditorInternal.VR;` still that in 2017.2?!

PR #1483, opened 2017-09-06
feat(SDK): ability to synchronise the play area transform on sdk switch

thestonefox:
A new checkbox on the SDK Setup Switcher that when checked will
synchronise the play area transform to the previous play area transform
when the SDK is switched.

This is useful when wanting to move the play area around with the
Simulator and then switch to an SDK of an actual HMD and be in the
position of where the Simulator was moved to.

### DIFF
--- a/Assets/VRTK/Source/Scripts/Utilities/SDK/VRTK_SDKSetupSwitcher.cs
+++ b/Assets/VRTK/Source/Scripts/Utilities/SDK/VRTK_SDKSetupSwitcher.cs
@@ -38,6 +38,8 @@ namespace VRTK
         private Button cancelButton;
         [SerializeField]
         private Button chooseButton;
+        [SerializeField]
+        private bool playareaSync = true;
 
         protected enum ViewingState
         {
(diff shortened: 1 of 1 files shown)


Assets/VRTK/Source/Scripts/Utilities/SDK/VRTK_SDKSetupSwitcher.cs
```diff
@@ -79,6 +82,14 @@ protected virtual void OnDisable()
         protected virtual void OnLoadedSetupChanged(VRTK_SDKManager sender, VRTK_SDKManager.LoadedSetupChangeEventArgs e)
         {
             Show(ViewingState.Status);
+            if (playareaSync && currentPlayarea != null)
+            {
+                Transform newPlayarea = VRTK_DeviceFinder.PlayAreaTransform();
+                newPlayarea.transform.position = currentPlayarea.transform.position;
+                newPlayarea.transform.rotation = currentPlayarea.transform.rotation;
+                newPlayarea.transform.localScale = currentPlayarea.transform.localScale;
```
bddckr: Maybe this should use our helper method to set the global scale instead? Depending on the exact hierarchy the play area game objects are in `localScale` would probably fail to work in some cases.

Issue #1484, opened 2017-09-07
Extract highlighter from Interactable Object


thestonefox:
The highlighter on the interactable object is currently hard coded to only work when a touch is initiated.

It would be better to extract the highlight code into it's own class that can work on any of the interaction types.

Then it would be possible to do highlight on grab or on use.

PR #1485, opened 2017-09-07
feat(Interaction): add near touch interactions

thestonefox:
The Interact Near Touch script generates a collider around the
controller that should be larger than the controller collider and
then when the controller is near something it will emit a near
touch event.

This can be useful for knowing what the controllers are near.


thestonefox:
I purposely didn't add Highlighting to near touch as i think this issue is a better way of adding highlighting

https://github.com/thestonefox/VRTK/issues/1484

Issue #1486, opened 2017-09-09
feat(RadialMenu): Add dead zone to middle of radial menu


Fulby:
Add an optional dead zone to the radial menu (float, 0-1 range) so that when the user touches the touchpad/thumbstick in the middle no button is immediately hovered over.

The rationale for this is that I want to use OnHoverEnter to trigger actions instead of OnClick as clicking the Oculus Touch thumbstick while it is deflected is uncomfortable (especially when deflected forwards).

Fulby:
I've implemented this in https://github.com/Fulby/VRTK/tree/release/3.3.0-alpha, just trying to

thestonefox:
PR https://github.com/thestonefox/VRTK/pull/1510

PR #1488, opened 2017-09-09
feat(RadialMenu): add dead zone to radial menu

Fulby:
feat(RadialMenu): add dead zone to radial menu

Allow the player to touch the centre of the touchpad without
immediately triggering OnHoverEnter event of a button.

### DIFF
--- a/Assets/VRTK/Prefabs/Internal/Scripts/VRTK_RadialMenu.cs
+++ b/Assets/VRTK/Prefabs/Internal/Scripts/VRTK_RadialMenu.cs
@@ -8,19 +8,19 @@ namespace VRTK
     using UnityEngine.UI;
     using UnityEngine.EventSystems;
 
-    public delegate void HapticPulseEventHandler(float strength);
-
     public struct TouchAngleDeflection
     {
-        public TouchAngleDeflection(float a, float d)
+        public TouchAngleDeflection(float angle, float deflection)
         {
-            angle = a;
-            deflection = d;
+            this.angle = angle;
+            this.deflection = deflection;
         }
 
         public float angle;
         public float deflection;
     }
+
+    public delegate void HapticPulseEventHandler(float strength);
 
     /// <summary>
     /// Provides a UI element into the world space that can be dropped into a Controller GameObject and used to create and use Radial Menus from the touchpad.
@@ -109,10 +109,10 @@ namespace VRTK
         /// <summary>
         /// The HoverButton method is used to set the button hover at a given angle and deflection.
         /// </summary>
-        /// <param name="tad">The angle and deflection on the radial menu.</param>
-        public virtual void HoverButton(TouchAngleDeflection tad)
+        /// <param name="givenTouchAngleDeflection">The angle and deflection on the radial menu.</param>
+        public virtual void HoverButton(TouchAngleDeflection givenTouchAngleDeflection)
         {
-            InteractButton(tad, ButtonEvent.hoverOn);
+            InteractButton(givenTouchAngleDeflection, ButtonEvent.hoverOn);
         }
 
         /// <summary>
@@ -128,10 +128,10 @@ namespace VRTK
         /// <summary>
         /// The ClickButton method is used to set the button click at a given angle and deflection.
         /// </summary>
-        /// <param name="tad">The angle and deflection on the radial menu.</param>
-        public virtual void ClickButton(TouchAngleDeflection tad)
+        /// <param name="givenTouchAngleDeflection">The angle and deflection on the radial menu.</param>
+        public virtual void ClickButton(TouchAngleDeflection givenTouchAngleDeflection)
         {
-            InteractButton(tad, ButtonEvent.click);
+            InteractButton(givenTouchAngleDeflection, ButtonEvent.click);
         }
 
         /// <summary>
@@ -147,10 +147,10 @@ namespace VRTK
         /// <summary>
         /// The UnClickButton method is used to set the button unclick at a given angle and deflection.
         /// </summary>
-        /// <param name="tad">The angle and deflection on the radial menu.</param>
-        public virtual void UnClickButton(TouchAngleDeflection tad)
+        /// <param name="givenTouchAngleDeflection">The angle and deflection on the radial menu.</param>
+        public virtual void UnClickButton(TouchAngleDeflection givenTouchAngleDeflection)
         {
-            InteractButton(tad, ButtonEvent.unclick);
+            InteractButton(givenTouchAngleDeflection, ButtonEvent.unclick);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Fulby:
Implements #1486

Assets/VRTK/Prefabs/Internal/Scripts/VRTK_RadialMenu.cs
```diff
@@ -85,27 +100,27 @@ public enum ButtonEvent
         /// The HoverButton method is used to set the button hover at a given angle.
         /// </summary>
         /// <param name="angle">The angle on the radial menu.</param>
-        public virtual void HoverButton(float angle)
+        public virtual void HoverButton(TouchAngleDeflection tad)
```
thestonefox: This (and in the other public accessors) makes changes to an existing public method that will cause breaking changes for anyone upgrading.

The correct solution would be to provide a new method and deprecate this method and then in this deprecated method just build up the appropriate call to the new method.

This way people will get the deprecation warning and know they need to upgrade their code, but things won't immediately break.

Fulby: I've [made this change](https://github.com/Fulby/VRTK/commit/4836289b8c51cbcd4cc38f060e26a32c358633eb) but any code which overrode HoverButton(float) will still break as the radial controller no longer calls that method. I'm not sure how to fix that?

thestonefox: If people have override it, then that's up to them to fix :)

I'd suggest changing the deprecation message to match the format of existing deprecation messages:

"`ClassName.methodName` has been replaced with `ClassName.methodName`. This method will be removed in a future version of VRTK."

Fulby: OK cool. I've fixed the deprecation messages.

bddckr:
Only suggestions I have:
- Change the `tad` naming to something more verbose as acronyms are generally not used anywhere else.
- The `TouchAngleDeflection` constructor should go **after** the fields.
- The `TouchAngleDeflection` constructor arguments shouldn't be acronyms. You can name them the same as the fields, you just have to prefix your field usage in the constructor with `this.` then. 😃

Fulby:
What should I change 'tad' to?

thestonefox:
if tad is an acronym of `touchAngleDeflection` then that's fine. Unless you're using that variable name internally elsewhere, then you can just use something like `givenTouchAngleDeflection` as parameters in methods,

Fulby:
Yeah it's just an acronym of touchAngleDeflection. I don't use 'tad' except as a function argument name. I do use initialTad and currentTad in the radial menu controller - I'm not sure if those are acceptable?

I understand @bddckr's third point and will make that change tonight but I'm not sure on the second point (where to move the definition). The only fields in that file are inside the VRTK_RadialMenu class, but the TouchAndDeflection is defined outside of that class. Should it be moved and if so where to?

Fulby:
I've renamed 'tad' as I'm not sure where it would be acceptable.

I've moved the struct definition to between the using statement and the delegate definition. I saw other helper structs in this position in other files.

I haven't added any help comments to the struct - some other structs have them and some don't. If you'd like me to add one please let me know.

bddckr:
>but I'm not sure on the second point (where to move the definition)

I'm talking about the `TouchAngleDeflection` struct. In that struct you currently have the constructor first, fields second. The existing codebase does it the other way round AFAIK.

It's all just nitpicking from me, so don't let that discourage you or something: It's a nicely done PR and a welcome addition. 😄

Fulby:
Ah I get what you mean now thanks, yep I'll fix that.

thestonefox:
For some reason this PR now has 5 commits in it. I think you need to git squash it into 1 commit

GitCop:

 * Commit: 6d9b67bbf0ccb4b8ed94aaacfe4376c50dabfd7f

 * Commit: 641a031d333d6bef535bb397409a1cf2d8e5b032

 * Commit: a43e04718c226bb254d2e7304b2b44c2040b47b7

 * Commit: 599d3722a18e224f6bc3d0251597d19bc1dd6fe1

---

Fulby:
Well I think I've broken the branch, I was trying to squash the commits and got something wrong which has added a load of other changes to this pull request. I may need to start again, if so I'll create a separate pull request as I think this one has had it.

Assets/VRTK/Prefabs/Internal/Scripts/VRTK_RadialMenu.cs
```diff
@@ -8,8 +8,37 @@ namespace VRTK
     using UnityEngine.UI;
     using UnityEngine.EventSystems;
 
+    /// <summary>
+    /// Angle and Deflection of the user's touch on the touchpad
+    /// </summary>
+    /// <param name="angle">The angle of the touch on the radial menu.</param>
```
bddckr: This line is not the valid place for this XML docs comment 😃

It has to go to be in front of the field itself. If you're using Visual Studio you should be able to just place the cursor on the line before the field and type in `///`. It should automatically fill in the XML docs comment template for fields.

Assets/VRTK/Prefabs/Internal/Scripts/VRTK_RadialMenu.cs
```diff
@@ -8,8 +8,37 @@ namespace VRTK
...
+    /// <param name="deflection">Deflection of the touch, where 0 is the centre and 1 is the edge.</param>
```
bddckr: Same as above: Move this to be in front of the field instead.

Assets/VRTK/Prefabs/Internal/Scripts/VRTK_RadialMenu.cs
Fulby: I copied the style from UIPointerEventArgs, BodyPhysicsEventArgs, etc which have param comments above the struct definition, I can add a comment to the constructor as well.

GitCop:

 * Commit: 6d9b67bbf0ccb4b8ed94aaacfe4376c50dabfd7f

 * Commit: 641a031d333d6bef535bb397409a1cf2d8e5b032

 * Commit: a43e04718c226bb254d2e7304b2b44c2040b47b7

 * Commit: 599d3722a18e224f6bc3d0251597d19bc1dd6fe1

---

Fulby:
I've created a new pull request which has a single commit: https://github.com/thestonefox/VRTK/pull/1509

@thestonefox, what would have been the correct way to squash the commits without having to introduce a new branch (and so a new pull request)? As far as I can tell, you can only squash while merging.

bddckr:
You can always squash commits using interactive rebase.

Assets/VRTK/Prefabs/Internal/Scripts/VRTK_RadialMenu.cs
bddckr: @Fulby You found a docs bug 😄

@thestonefox The struct docs are wrong, same as before for the enums ([enum PR](https://github.com/thestonefox/VRTK/pull/1465)).

Fulby:
Thanks @bddckr. I thought git philosophy was that commits are sacrosanct (even editing the commit message is difficult), but I'm guessing that's only once the pull request is accepted so the code is in multiple repos?

Hopefully this is now a clean pull request: https://github.com/thestonefox/VRTK/pull/1510

PR #1507, opened 2017-09-15
fix(pointers): Custom tracers not cleaned up on destroy.

insominx:
When disabling the BezierPointerRenderer, it doesn't destroy GameObjects created using custom tracers.


thestonefox:
This doesn't actually work as if you destroy the line renderer then it won't be recreated next time the pointer is activated.

Plus it doesn't go far enough to cover all edge cases.

This PR should fix the issue and any additional edge cases: 
https://github.com/thestonefox/VRTK/pull/1508

PR #1510, opened 2017-09-16
feat(RadialMenu): add dead zone to radial menu - fixes #1486

### LINKED ISSUE #1486: feat(RadialMenu): Add dead zone to middle of radial menu

Fulby:
Allow the player to touch the centre of the touchpad without
immediately triggering the OnHoverEnter event of a button.

### DIFF
--- a/Assets/VRTK/Prefabs/Internal/Scripts/VRTK_IndependentRadialMenuController.cs
+++ b/Assets/VRTK/Prefabs/Internal/Scripts/VRTK_IndependentRadialMenuController.cs
@@ -290,7 +290,7 @@ namespace VRTK
             }
         }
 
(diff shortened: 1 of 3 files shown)


thestonefox:
Everything looks ok to me now, I'll pull down and test the PR later today before merging.

Thanks for sticking with it and enduring the contribution process :)

Fulby:
Cool, thanks to you and @bddckr for enduring my git/github fumblings.

PR #1513, opened 2017-09-20
WIP - Feature - Interactive / Continuous haptics

kristianpd:
WIP PR, will clean up commits/etc when ready to actually commit. Right now I'm opening this for feedback and discussion.

### What's this?

This PR provides a solution for interactive / continuous haptics when interacting with `InteractableObject`s. The main issue I have with the current `InteractHaptics` implementation, is that it's  either a basic `OnX` event, or in the alpha release, the ability to provide a static pulse/interval until the interaction ends. 

This PR introduces the idea I'm currently calling "Haptic Curves". Haptic curves allow the developer to specify an interaction curve that will be applied to an `InteractableObject`. A basic example of this would be an old wooden Door that opens with a "tighter" feel and slowly loosens up as you pull it out from the frame.

There are two components to an interactive haptic. `VRTK_InteractiveHaptics` which specifies the strength and interval of the feedback over a fixed range and `VRTK_InteractiveHapticsInput` which gives a normalized value of `0..1` representing that range.

## VRTK_InteractiveHaptics
image

`InteractionType` - `WhileTouching`, `WhileGrabbing`, `WhileUsing`. Specifies the start/stop conditions for haptic feedback.
`Strength curve` - Given a value from 0 to 1, evaluate a strength along the curve. Here, we use a constant strength.
`Interval curve` - Given a value from 0 to 1, evaluate an interval along the curve. Here, we want it to feel "tighter" at the start, and slowly open up into less feedback.
`Interval Min/Max` - Specify the interval range the curve applies to. `Y` value of 1 on the curve, corresponds to the `Min` interval (the fastest one), 0, the `Max` interval.
`Min Value Delta` - Currently a solution I'm using to prevent constant feedback since it's impossible to hold something completely still, this corresponds to 0.5% in this example.

## VRTK_InteractiveHapticsInput

An abstract class that provides a normalized value from `0..1` that corresponds to your range. A few examples are:

image a transform one that can watch position, rotation, scale

image a hinge joint one that works better for hinge joint angle rotations than the transform one above

image a simple velocity magnitude input example

I'm still very open to refactoring this interaction. I played with configurable unity events, but I think this kind of "`Input`" approach works a bit better. I'm also open to combining these into a more generalized input class (think `ConfigurableJoint`) that would give us a bigger suite of behaviours that we could pick from drop downs. Very much open to discussion.

### Example scene
`045_InteractiveHaptics`, please try it out and let me know what you think!

### Video demo

Turn down your sound, it's a little annoying as I play a blip on feedback.
[InteractiveHaptics demo](http://www.youtube.com/watch?v=IgkSwqmknsw)

### Changes to work on master

Swap out `OnEnable` and `OnDisable` functions with this implementation to try this out on master: 
https://github.com/thestonefox/VRTK/pull/1513/files#diff-8b5330bc56b1097360fa12a80280bf19R58

```

		protected virtual void OnEnable()
		{
			objectToAffect = (objectToAffect != null ? objectToAffect : GetComponentInParent<VRTK_InteractableObject>());

			if (objectToAffect != null)
			{
				if (interactionType == InteractionType.WhileTouching)
				{
					objectToAffect.InteractableObjectUntouched += StopHaptics;
					objectToAffect.InteractableObjectTouched += StartHaptics;
				}
				else if (interactionType == InteractionType.WhileGrabbing)
				{
					objectToAffect.InteractableObjectUngrabbed += StopHaptics;
					objectToAffect.InteractableObjectGrabbed += StartHaptics;
				}
				else
				{
					objectToAffect.InteractableObjectUnused += StopHaptics;
					objectToAffect.InteractableObjectUsed += StartHaptics;
				}
			}
			else
			{
				VRTK_Logger.Error(VRTK_Logger.GetCommonMessage(VRTK_Logger.CommonMessageKeys.REQUIRED_COMPONENT_MISSING_FROM_GAMEOBJECT, "VRTK_InteractiveHaptics", "VRTK_InteractableObject", "the same or parent"));
			}

			interactiveHapticsInput = (interactiveHapticsInput != null ? interactiveHapticsInput : GetComponent<VRTK_InteractiveHapticsInput>());

			if (interactiveHapticsInput != null)
			{
				interactiveHapticsInput.InputProvided += Interact;
			}
			else
			{
				VRTK_Logger.Error(VRTK_Logger.GetCommonMessage(VRTK_Logger.CommonMessageKeys.REQUIRED_COMPONENT_MISSING_FROM_GAMEOBJECT, "VRTK_InteractiveHaptics", "VRTK_InteractiveHapticsInput", "the same"));
			}
		}

		protected virtual void OnDisable()
		{
			if (objectToAffect != null)
			{
				if (interactionType == InteractionType.WhileTouching)
				{
					objectToAffect.InteractableObjectUntouched -= StopHaptics;
					objectToAffect.InteractableObjectTouched -= StartHaptics;
				}
				else if (interactionType == InteractionType.WhileGrabbing)
				{
					objectToAffect.InteractableObjectUngrabbed -= StopHaptics;
					objectToAffect.InteractableObjectGrabbed -= StartHaptics;
				}
				else
				{
					objectToAffect.InteractableObjectUnused -= StopHaptics;
					objectToAffect.InteractableObjectUsed -= StartHaptics;
				}
			}

			if (interactiveHapticsInput != null)
			{
				interactiveHapticsInput.InputProvided -= Interact;
			}
		}
```


GitCop:

 * Commit: 1500f6c012e9d2b7357293b52b3977d786550167

 * Commit: d93b8c3a1ec7e53c052b4ad33e83b7ed8b4c30da

 * Commit: 01e8dea4d91aeedfcd962e7d221a8147a40926be

---

kristianpd:
Any thoughts on this @thestonefox @bddckr ? If this is something you'd like to see in VRTK core i can clean up the PR and make it ready for merge.

thestonefox:
I can't check it till I'm back in the united kingdoms of great britain

bddckr:
I don't have time for it, sorry.

thestonefox:
@kristianpd it's good to try and get other community members to help review and test pull requests. it's easy to forget that @bddckr and I both have day jobs and can't always attend to reviewing stuff in a timely manner.

kristianpd:
Still looking for any volunteers to test this out... You weren't kidding.

This is straightforward to set up if anyone watching the repo wants to give it a spin.

andybak:
One approach taken in other open source projects is "Keep features separate unless they require major changes to the core - or are almost universally useful to users".

As a project matures, backwards compatibility becomes more important and people will tend to stick on particular release versions. This means that new features are unable to evolve quickly and it becomes almost impossible for alternative (and potentially better) implementations to become widely adopted.

Is there any reason this feature couldn't be an add-on to VRTK for now? It would be much easier for people to try it out. And maybe VRTK could add a list of "blessed" add-ons to the documentation - which helps discoverability without overburdening the core.

I haven't looked very closely at the code but it doesn't seem to require any changes to existing files so seems an ideal candidate to be a plugin/add-on/whatever.

kristianpd:
@andybak I definitely agree with the concern of bloating vrtk core. Since opening this pr we've had a lot of discussion in #vrtk-core-development on Slack about separating out vrtk-core from vrtk-use-cases, reducing the number of examples, and moving lots of the helpers into the use case / examples repo. 

You're right in the fact that this PR is pretty lightweight and is easily added on top of VRTK as a sort of plugin, though I did feel VRTK Core was lacking a bit in terms of haptic options. That said, maybe my use case is less common than I think (I can't find anyone to test it after all 😆 .

Coming from the web dev world, I do like the idea of a package manager for VRTK. It definitely seems a bit overkill for the current community and size of VRTK, but it would be nice.  I see a few options for us:

1. Keep it in VRTK core as an enhancement above the "on/off" haptic solution currently available.
2. Move it to the proposed but not yet defined "use cases" separation out of VRTK (lots of other VRTK examples would be candidates as well)
3. Consider separating the "use cases" into many different repos instead of just one and track with a file or ghetto package installer. I think this goes beyond the average VRTK user's comfort level with installing plugins and my concern here is that without a test suite or large contributor base, these extra repos would all fall behind and break compatibility

If the repo existed for # 2 i'd probably pick that one as I was able to tie this in with existing (basic) VRTK haptics rather quickly and it does not need any VRTK core changes to work.

andybak:
A few off the cuff thoughts:

>  That said, maybe my use case is less common than I think.

The thing that made me look at this PR closely was a conversation with someone about how lacking SteamVR haptics were compared to Oculus. This could help mitigate that difference.

> Coming from the web dev world, I do like the idea of a package manager for VRTK.

Don't we already have one built into Unity? ;-)

>  I think this goes beyond the average VRTK user's comfort level with installing plugin

If someone's comfort level doesn't extend to downloading a zip from Github then I'm not sure they are VRTK's intended audience!

I imagined something as simple as a handful of agreed conventions in terms of project name prefix and folder structure and a division between "blessed" (popular, well-maintained and listed in the VRTK docs) add-ons and "community" (everything else). People are pretty good at judging the quality of an add-on from various signals (number of open issues, recent commits, quality of docs etc).

This kind of thing worked well for jQuery for a long time - and javascript developers weren't exactly at the cutting edge of computer science at that time. (of course since the advent of Angular et al. they have ascended to a higher plane that is beyond my mortal understanding...)

bkensey:
After having tried this out, I think it works wonderfully and should be part of VRTK core but needs better labeling. The hardest thing to wrap my head around was what the x and y axis represented for each interaction type. When you say "along the curve" in your property descriptions, I think you mean something more like "as a function of the haptic objects's change in state (location, rotation, scale, acceleration) relative to its origin."

`Interval Min/Max` (If I'm understanding things correctly) should be something like "Min/Max Intensity."

`Min Value Delta` might be better described as "Sensitivity Threshold."

kristianpd:
Still alive, hoping to get back to this in a week or so. Sorry!

PR #1518, opened 2017-09-23
fix(SnapDropZone): remove all joints before removing rest - fixes #1517

### LINKED ISSUE #1517: SnapDropZone highlighter falls to floor if object with joints are used as highlighter object
### Environment
VRTK 3.3.0 alpha
Unity 5.6.3

### Steps to reproduce
 * Create an interactible object that has some kind of joint setup
 * Assign that as the highlighter object for a snap drop zone.
 * Since it has joints, it will fail at removing the RigidBody because the Joint depends on it.
 * This will result in a Highlighter object with a RigidBody on it that will fall to the floor when playing the game.

One solution would be to first make sure to get rid of all the joints before getting rid of the Components.

virror:
This stops removing RigidBodys from failing cause Joints depends on them.

### DIFF
--- a/Assets/VRTK/Prefabs/Internal/Scripts/VRTK_SnapDropZone.cs
+++ b/Assets/VRTK/Prefabs/Internal/Scripts/VRTK_SnapDropZone.cs
@@ -998,7 +998,14 @@ namespace VRTK
             //determine components that shouldn't be deleted from highlight object
             string[] validComponents = new string[] { "Transform", "MeshFilter", "MeshRenderer", "SkinnedMeshRenderer", "VRTK_GameObjectLinker" };
 
-            //go through all of the components on the highlighted object and delete any components that aren't in the valid component list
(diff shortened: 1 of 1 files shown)


thestonefox:
All looks ok, can you do me 1 favour and rebase the commit message so it doesn't have a full stop at the end of the commit title, like so:

`fix(SnapDropZone): remove all joints before removing rest`

Actually it may read better if it reads:

```
fix(SnapDropZone): prioritise joint removal on highlight object

If the interactable object that is being used as the highlight object
template has a joint on it, then it would cause issues due to the
rigidbody attempting to be removed first which would then fail because
a Joint was present. This would cause the snap drop zone highlight
object to keep it's rigidbody and fall will the physics gravity when
the scene started.
```

virror:
Fixed!

kristianpd:
:shipit:

PR #1529, opened 2017-09-29
fix(UI): have UIPointer look for ControllerEvents in parents

Fulby:
Allow the UIPointer to be a child of the object with the controller,
instead of having to be on the same object. Now matches VRTK_Pointer
and several other scripts.


thestonefox:
You need to update, this has already been done ;)

https://github.com/thestonefox/VRTK/blob/release/3.3.0-alpha/Assets/VRTK/Source/Scripts/UI/VRTK_UIPointer.cs#L426

Fulby:
Sorry my bad, I didn't check the 3.3.0 branch.

PR #1530, opened 2017-10-01
fix(Simulator): ensure mouse lock behaviour works correctly

reznovVR:
This fixes the simulator mouse cursor lock to behave as expected with
the "Always" selection of the "Mouse movement input" mode and actually
lock the cursor in the game view.


thestonefox:
Because this is adding a new option I'd say it's actually not a `fix` but a `feat`

then it will also need to go onto the `release/3.3.0-alpha` branch

thestonefox:
Oh wait, it's not adding a new option, its just adding a new header, right?

thestonefox:
In that case the only thing i'd suggest is changing the commit message, it doesn't need to have the word `fix` in the title as it starts with `fix` so something like:

`fix(Simulator): ensure mouse lock behaviour works correctly`

reznovVR:
Yeah changed it :)

PR #1534, opened 2017-10-02
feat(Locomotion): add tunnel overlay camera effect for comfort options

reznovVR:
The Tunnel Overlay script creates a texture overlay that can increase
in it's radius towards the centre of the screen to reduce the perceived
field of view.

This can be triggered based on the movement and rotation of the play
area to reduce nausea with artificial locomotion. The tunnel effect
can also be applied when stationary to simulate effects such as damage
being taken.

Thanks to @reznovVR for the original port of the excellent
UnityVRTunnelling repo by @sixways:

https://github.com/SixWays/UnityVrTunnelling

### DIFF
--- a/Assets/VRTK/Documentation/API.md
+++ b/Assets/VRTK/Documentation/API.md
@@ -78,8 +78,8 @@ Use the mouse and keyboard to move around both play area and hands and interacti
  * **Show Control Hints:** Show control information in the upper left corner of the screen.
  * **Hide Hands At Switch:** Hide hands when disabling them.
  * **Reset Hands At Switch:** Reset hand position and rotation when enabling them.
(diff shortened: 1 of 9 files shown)


thestonefox:
I've taken pretty much most of your work and added some additions in this PR:

https://github.com/thestonefox/VRTK/pull/1538

I'm happy if you want to swipe my code (if you think it's worth doing) and rebasing it into your PR so you get the glory, but I'm equally happy to merge my own PR if you can't be bothered :) :p


My changes are:

 * Don't load the shader and have a dedicated material, then people don't need to include the shader in builds
 * Allow for the tunnel colour to be changed
 * Allow for an initial tunnel size to be set.

The last two points would be good if you wanted to use the tunnel effect for something else like showing damage being taken for instance.

PR #1552, opened 2017-10-09
feat(Locomotion): add disable in simulator to the tunnel overlay effect

reznovVR:
This change adds an option to disable the effect in the simulator by
default as it rarely makes sense to have the effect active for other than
temporary settings tweaking and testing.


GitCop:

 * Commit: c124efddf9c7f82d59f5da0ad229f871d0d92b2d

---

thestonefox:
I'd prefer if this was more generic like a list of SDKs to ignore then you can add the simulator to that list.

reznovVR:
Does such list have any meaning as with the headsets it would be always on/off per game settings?

thestonefox:
You may want it off for gear vr maybe but on for vive. Alternatively you could put the tunnel script on its own game object and use the sdk object state script to turn on and off.

PR #1562, opened 2017-10-17
fix(TrackObjectGrabAttach): dropping objects early - fixes #1563

### LINKED ISSUE #1563: TrackObjectGrabAttach dropping objects early
### Environment

 * Source of VRTK (master-branch, through `git clone`)
 * Version of VRTK (revision `e16de5a`)
 * Version of Unity3D (e.g. Unity 2017.2.0f3)
 * Hardware used (HTC Vive)
 * SDK used (SteamVR, also tested with NewtonVR's SteamVR camera rig)

### Steps to reproduce

1. Open example project `021_Controller_GrabbingObjectsWithJoints`
2. Select `ExampleWorldObjects > TrackGrab_Cube` (the green cube of the three)
3. Scroll the inspector to find Track Object Grab Attach options and set `Velocity Limit` to anything other than Infinity (I used 1, but this happens with any value). This affects `Angular Velocity Limit` aswell but not as demonstratibly.
4. Enter VR and attempt to move the green `TrackGrab_Cube`.

### Expected behavior

The cube's movement speed should be limited and if you set the `Velocity Limit` to something like 1 or less, it should be visibly slowly following the controller. Otherwise the cube should follow the controller as previously to the changes to the example, except it's velocity should be limited.

### Current behavior

The cube is difficult to pick up as it will stop following the controller if the controller moves even slightly too fast for the cube, but it won't even detach from the controller properly.

Example gif:
preview

Teascade:
maxDistanceDelta being a constant caused the magnitudes of the
calculated positional and angular velocities to not be limited properly.
This resulted in the held objects getting dropped prematurely without
properly detaching them: The objects were attached in theory but not
following the controller as intended.

The solution is to change the constant to the predefined velocity limit,
causing the magnitude to limit properly to fit the later if-statements.

### DIFF
--- a/Assets/VRTK/Scripts/Interactions/GrabAttachMechanics/VRTK_TrackObjectGrabAttach.cs
+++ b/Assets/VRTK/Scripts/Interactions/GrabAttachMechanics/VRTK_TrackObjectGrabAttach.cs
@@ -89,7 +89,7 @@ namespace VRTK.GrabAttachMechanics
                 return;
             }
 
-            float maxDistanceDelta = 10f;
+            float maxDistanceDelta = velocityLimit;
             Vector3 positionDelta = trackPoint.position - (grabbedSnapHandle != null ? grabbedSnapHandle.position : grabbedObject.transform.position);
             Quaternion rotationDelta = trackPoint.rotation * Quaternion.Inverse((grabbedSnapHandle != null ? grabbedSnapHandle.rotation : grabbedObject.transform.rotation));
 


thestonefox:
I don't think this is a viable solution.

It only actually makes sense if the velocity limit is lower than 10f

If the velocity limit is 20f for instance and your max distance delta then also becomes 20f, then you get unexpected results.

I would say a better fix would be having additional parameters for allowing custom settings of angular velocity distance delta and velocity distance delta.

thestonefox:
Superseded by https://github.com/thestonefox/VRTK/pull/1600

PR #1575, opened 2017-10-26
feat(Presence): add body physics to simulator

ctorkut:
feat(Simulator): SDK_InputSimulator update methods marked as protected virtual.
feat(Simulator): add a custom setting to SDK_InputSimulator to enable Body Physics.
feat(Presence): add CanMove method to VRTK_BodyPhysics to mimic movement behaviour on
ObjectControlAction.
feat(Presence): add a seperate CustomRaycast to VRTK_BodyPhysics to enable
differentiation between floor snap and movement collisions.

### DIFF
--- a/Assets/VRTK/Source/SDK/Simulator/SDK_InputSimulator.cs
+++ b/Assets/VRTK/Source/SDK/Simulator/SDK_InputSimulator.cs
@@ -111,6 +111,11 @@
         [Tooltip("Key used to switch between hair touch mode.")]
         public KeyCode hairTouchModifier = KeyCode.H;
 
+        [Header("Custom Settings")]
+
+        [Tooltip("An optional Body Physics script to check for potential collisions in the moving direction. If any potential collision is found then the move will not take place.")]
+        public VRTK_BodyPhysics bodyPhysics;
+
         #endregion
         #region Private fields
 
@@ -327,7 +332,7 @@
             }
         }
 
-        private void UpdateHands()
+        protected virtual void UpdateHands()
         {
             Vector3 mouseDiff = GetMouseDelta();
 
@@ -369,7 +374,7 @@
             }
         }
 
-        private void UpdateRotation()
+        protected virtual void UpdateRotation()
         {
             Vector3 mouseDiff = GetMouseDelta();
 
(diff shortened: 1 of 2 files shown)


thestonefox:
There seems to be multiple pull requests that are handling separate features that have all been squashed into one commit.

Each commit should be around 1 particular feature or fix, if a pull request is proposing an overall feature or fix that requires smaller commits to ensure the commits are self contained then they should.

ctorkut:
Yes, the last feature in the description is not connected to the overall functionality. How do you want to handle this. TBH, I didn't think that feature should occupy a whole pull request since it is a 3 line change on one file, but I understand your concerns. 

How do you want to handle this? Should I close this pull request and open a completely new one, or should I make an additional commit removing the last feature?

virror:
"feat(Simulator): add a custom setting to SDK_InputSimulator to enable Body Physics."
Why not just add the body physics script instead? there is no need for this option as the bodyPhysics already works as expected for the simulator?

virror:
Ahh, then I get what you mean. Makes sense!

thestonefox:
You can't walk through the walls on example 17 with the simulator, the body physics stops you.

ctorkut:
You can't walk through them, but can walk inside them and see what's behind them with a constant jitter with headset fade popping on and off. However, when using a headset, if you set body physics within TouchpadControlOptions, you won't be able to walk inside the walls using touchpad as the collision sweep will stop you, making headset fade and position rewind only necessary when user moves inside the play area. I implemented the same functionality for the simulator.

thestonefox:
But moving the simulator around isn't the same as implementing the touchpad control mechanic.

Moving the simulator is just a quick way of getting around a level, the WASD keys are not a replication of the touchpad control mechanic.

I don't think it's a good idea to have a body physics script injected into the simulator, seems counter intuitive to me.

ctorkut:
TBH, if one would simulate Touchpad Walking (which is the go-to locomotion method after teleportation), WASD keys on simulator is the nearest available option. I think they should behave similarly.

On the other hand, I can see why you want to decouple simulator from other internal components of VRTK even though a lot of other components have soft dependencies on Body Physics.

Maybe a better solution would be to implement the body physics to handle its own movement, ensuring these kinds of inconsistencies won't appear between different input mechanisms.

thestonefox:
The better solution would be to simulate the touchpad within the simulator so you can use the touchpad locomotion script.

The WASD keys in the simulator is basically a SLEW mode and isn't representative of how it would actually work on a real VR setup.

virror:
I agree that implementing touch pad support would be the best option, not sure how that would be done though. Either just arrow keys or maybe numpad can be used to get 8 directions instead of 4?

thestonefox:
I wonder if some sort of on screen pad could be used, you know like how they do it in mobile games?

virror:
Could be an option, would be nice to have at least some form of keyboard input as well, or it can be very frustrating to use together with handling the hands?

ctorkut:
You can simulate touchpad on Oculus Touch or Vive controller with an on-screen solution; but there are other input controllers like Oculus Remote and X-box gamepad with analog buttons. Movement physics should be independent from movement controller.

thestonefox:
Oculus remote and xbox controller aren't supported by vrtk (also the oculus remote has been discontinued). But there may be future controllers that have this limitation too

thestonefox:
Lets close this for now, with the mindset that we'll make a touchpad/thumbstick simulator which would replicate the solution in the correct fashion.

PR #1576, opened 2017-10-27
fix(Utilities): fixed error for Unity versions prior to 5.6

gpvigano:
In Unity 5.5.4 you get this error:
`Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs(359,20):
 error CS0103: The name 'VRStats' does not exist in the current context`

The error disappears changing `VRStats` to `XRStats`.

### DIFF
--- a/Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
+++ b/Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
@@ -356,7 +356,7 @@ namespace VRTK
             float gpuTimeLastFrame;
             return (XRStats.TryGetGPUTimeLastFrame(out gpuTimeLastFrame) ? gpuTimeLastFrame : 0f);
 #else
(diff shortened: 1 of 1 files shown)


thestonefox:
Are you sure in 5.5 it was called XR?

the XR namespace only came in 2017.2

thestonefox:
Also this fix will also break for anyone still using the VRStats namespace so it's not really valid.

gpvigano:
Maybe this happened because I have both Unity 5.5.4 and 2017.2 installed, does it make sense?

thestonefox:
Ah right i see the issue now, it's because line 359 wasn't updated which is what your PR is doing

thestonefox:
one thing you need to do is rebase the commit subject line, the subject shouldn't be in past tense and also doesn't need to repeat the word fix

e.g.

`fix(Utilities): update VR namespace to be XR`

gpvigano:
Sorry...

PR #1582, opened 2017-10-31
feat(Locomotion): apply play area parent position to teleporter - fixes #1559

### LINKED ISSUE #1559: Samsung GearVR - Dash Teleport - OVRCameraRig drops to floor height after initial move
### Environment
Using:
VRTK 3.3.0 from Git Hub (via Slack)
Unity 2017.1

Samsung GearVR
Samsung S7 Edge

Latest Steam and Oculus (1.18.1)

### Steps to reproduce
Following You Tube tutorial, Getting Started With GearVR, load Dash Teleport example sample.

Ensure you select the OculusVR child object and set its height to 1.5 in the Y axis. 

### Expected behavior

Deploy to device.
Wait for a moment, assess if the camera rig is at the correct height when the demo starts.
Use the Samsung controller to teleport.  Your viewpoint should remain at the same height it initialises at.  Your height should be around eye level for a human.

### Current behavior

You will find your eyes viewpoint at ground level suddenly, instead of where you expect it to be - eye height for a human.  Now each time you teleport, moving does work, however your eyes remain at floor level.

thestonefox:
The Height Adjust Teleport script now has a new option that will also
apply the play area parent position as an offset when teleporting.

This fixes issues with headsets that aren't room scale aware and
therefore cannot determine where the floor is so will always send the
HMD view to the point of teleport. This would cause the user to
be teleported into the floor.

This new feature allows for the play area camera rig to be placed
in a parent GameObject and that parent GameObject to have a Y position
set on it to be used as the default standing height of the HMD view.

### DIFF
--- a/Assets/VRTK/Documentation/API.md
+++ b/Assets/VRTK/Documentation/API.md
@@ -2050,6 +2050,7 @@ Updates the `x/y/z` position of the SDK Camera Rig with an optional screen fade.
 ### Inspector Parameters
 
  * **Snap To Nearest Floor:** If this is checked, then the teleported Y position will snap to the nearest available below floor. If it is unchecked, then the teleported Y position will be where ever the destination Y position is.
+ * **Apply Playarea Parent Offset:** If this is checked then the teleported Y position will also be offset by the play area parent Transform Y position (if the play area has a parent).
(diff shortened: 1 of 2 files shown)


IPete2:
Thank you! 

P.

PR #1585, opened 2017-11-03
fix(PointerRenderer): too large collider of the ObjectInteractor of the tip of the PointerRenderer

feranti:
Way too large. It touches InteractableObjects which have no contact with the ray at all.
This is due to the fact that the default Sphere collider has a 0.5 radius. The correct value seems to be 0.025.


GitCop:

 * Commit: 521f24686df28e816f8e36f50be9a9704ca8497a

---

thestonefox:
Don't see the point in this, the ObjectInteractor is scaled at runtime to match the cursor. Just arbitrary changing a value makes it fine for one use case but not another.

feranti:
I can understand your point, and I would have said the same thing in your place. But I use the basic default BasePointerRenderer from VRTK and have the issue.
I will dig into this deeper to see if I didn't miss anything there. I wil try in a clean scene.

thestonefox:
If you're using the base pointer renderer and inheriting into your own concrete class then all you need to do is call the scale object interactor code.

PR #1613, opened 2017-11-14
feat(SDK): add option to exclude target build platforms

thestonefox:
There was an issue with Unity 2017.2 where it would display a warning
when trying to manage the VR settings due to the Facebook and Switch
platform not being supported.

The new Target Platform Group Exclusions will allow for these two
platforms to be automatically excluded but also the array can be
updated with additional platforms if required.


thestonefox:
@bddckr if you get chance, take a look over this and see if it makes sense.

basically it's just preventing the stupid Facebook/Switch warnings in 2017.2 by having a managed list of the build targets to ignore.

bddckr:
I guess this is fine as the warning was only logged when trying to set the VR settings for the platforms. 👍

thestonefox:
well it removes the warnings at least :) and means we don't have to keep hard coding bad platforms as people can just update the list. I guess it works for now at least :)

PR #1622, opened 2017-11-17
feat(SDK): add WindowsMR support for SteamVR

thestonefox:
The SteamVR SDK also supports Windows Mixed Reality devices but again
has some slight differences such as controller collider shapes and
controller orientations.

The Windows Mixed Reality controller also has an additional thumbstick
as well as a touchpad. The thumbstick is now referred to as
`TouchpadTwo` but only has axis change events as the button press is
reserved as a system event in SteamVR and there is no touch sensing
available.


thestonefox:
Only supports Acer headset at the moment as I don't know the names of any of the other headsets :p

PR #1634, opened 2017-11-23
refactor(Structure): provide helper methods for lists and dictionaries - fixes #1631

### LINKED ISSUE #1631: Replace all dictionary contains checks with TryGetValue
A number of dictionary datatypes are used in VRTK, however many of the checks first do a `Contains` check before getting the value.

This does a double dictionary lookup which is inefficient.

The `TryGetValue` method on a dictionary does the same thing with a single lookup.

The code is a bit more verbose but can be easily wrapped into a simple `DictionaryContains` method on the Shared Methods class.

thestonefox:
There are a number of places where the use of a List or Dictionary has
unnecessary repetition in checking if a value or key is present before
either adding or removing from the list or dictionary.

There are also times where a value is attempted to be retrieved from
a dictionary but it doesn't exist which would then need to be added
so multiple dictionary lookups are done, which is inefficient. The
proposed changes ensure a single look up is done via `TryGetValue`
which will reduce dictionary usage overhead.

### DIFF
--- a/Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
+++ b/Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
@@ -155,19 +155,16 @@ namespace VRTK
         /// <returns>An array of Colliders that are found in the given GameObject array.</returns>
         public static Collider[] GetCollidersInGameObjects(GameObject[] gameObjects, bool searchChildren, bool includeInactive)
         {
-            List<Collider> foundColliders = new List<Collider>();
+            HashSet<Collider> foundColliders = new HashSet<Collider>();
             for (int i = 0; i < gameObjects.Length; i++)
             {
                 Collider[] gameObjectColliders = (searchChildren ? gameObjects[i].GetComponentsInChildren<Collider>(includeInactive) : gameObjects[i].GetComponents<Collider>());
                 for (int j = 0; j < gameObjectColliders.Length; j++)
                 {
-                    if (!foundColliders.Contains(gameObjectColliders[j]))
-                    {
-                        foundColliders.Add(gameObjectColliders[j]);
-                    }
+                    foundColliders.Add(gameObjectColliders[j]);
                 }
             }
-            return foundColliders.ToArray<Collider>();
+            return foundColliders.ToArray();
         }
 
         /// <summary>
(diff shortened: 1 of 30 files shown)


Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -472,6 +469,109 @@ public static Vector3 AxisDirection(int axisIndex, Transform givenTransform = nu
...
+        }
+
+        /// <summary>
+        /// The RemoveListValue method attempts to remove an entry from the given list based on the given value.
+        /// </summary>
+        /// <typeparam name="TValue">The datatype for the list value.</typeparam>
+        /// <param name="list">The list to remove the value from.</param>
+        /// <param name="value">The value to remove.</param>
+        /// <returns>Returns `true` if the value was found and removed. Returns `false` if the value was not found and therefore not removed.</returns>
+        public static bool RemoveListValue<TValue>(ref List<TValue> list, TValue value)
```
bddckr: Why is `list` a `ref`? If you change this do a null check like in the dictionary method.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -472,6 +469,109 @@ public static Vector3 AxisDirection(int axisIndex, Transform givenTransform = nu
...
 
+        /// <summary>
+        /// The AddListValue method attempts to add a value in the given list if the value does not already exist. `forceAdd = true` will always add a new value to the list even if it already exists.
+        /// </summary>
+        /// <typeparam name="TValue">The datatype for the list value.</typeparam>
+        /// <param name="list">The list to retrieve the value from.</param>
+        /// <param name="value">The value to attempt to add to the list.</param>
+        /// <param name="forceAdd">If this is `true` then the value provided will always be appended to the list. If this is `false` the value provided will only be added to the list if it doesn't already exist.</param>
+        /// <returns>Returns `true` if the given value was successfully added to the list. Returns `false` if the given value already existed in the list and `forceAdd = false`.</returns>
+        public static bool AddListValue<TValue>(ref List<TValue> list, TValue value, bool forceAdd = false)
```
bddckr: Why is `list` a `ref`? If you change this do a null check like in the dictionary method.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -472,6 +469,109 @@ public static Vector3 AxisDirection(int axisIndex, Transform givenTransform = nu
...
+        {
+            if (forceAdd || !list.Contains(value))
```
bddckr: Use `HashSet<TValue>` if you need to ensure there are no duplicated items, don't use a `List<TValue>`.

`HashSet<TValue>` has way better performance.

If Unity doesn't offer a default inspector for `HashSet<TValue>`s we can provide one.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -472,6 +469,109 @@ public static Vector3 AxisDirection(int axisIndex, Transform givenTransform = nu
...
+        /// <summary>
+        /// The RemoveListValue method attempts to remove an entry from the given list based on the given value.
+        /// </summary>
+        /// <typeparam name="TValue">The datatype for the list value.</typeparam>
+        /// <param name="list">The list to remove the value from.</param>
+        /// <param name="value">The value to remove.</param>
+        /// <returns>Returns `true` if the value was found and removed. Returns `false` if the value was not found and therefore not removed.</returns>
+        public static bool RemoveListValue<TValue>(ref List<TValue> list, TValue value)
+        {
+            if (list.Contains(value))
```
bddckr: Remove this and change the whole method to
```c#
return list.Remove(value);
```
instead. Or better: Remove this helper method altogether since it's now adding nothing to the existing method on `List<TValue>`.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -472,6 +469,109 @@ public static Vector3 AxisDirection(int axisIndex, Transform givenTransform = nu
...
+        /// <param name="dictionary">The dictionary to retrieve the value from.</param>
+        /// <param name="key">The key to retrieve the value for.</param>
+        /// <param name="setMissingKey">If this is `true` and the given key is not present, then the dictionary value for the given key will be set to the `setValue` parameter. If this is `false` and the given key is not present then the `setValue` parameter will be returned as the value.</param>
+        /// <param name="setValue">The value to utilise when either setting the missing key (if `setMissingKey` is `true`) or the default value to return when no key is found (if `setMissingKey` is `false`).</param>
+        /// <returns>The found value for the given key in the given dictionary, or the default value if no key is found.</returns>
+        public static TValue GetDictionaryValue<TKey, TValue>(Dictionary<TKey, TValue> dictionary, TKey key, bool setMissingKey = false, TValue setValue = default(TValue))
+        {
+            if (dictionary == null)
+            {
+                return default(TValue);
```
bddckr: This should perhaps return a `defaultValue` which is supplied as a parameter instead.

Since this means the method has a lot parameters with confusing `setValue` vs. `defaultValue` I suppose splitting this method into two is a good idea, though I'm not sure what the use cases are exactly. E.g. do we need a way to set a different `defaultValue` compared to `setValue` at all?

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -472,6 +469,109 @@ public static Vector3 AxisDirection(int axisIndex, Transform givenTransform = nu
...
+        /// The RemoveDictionaryKey method attempts to remove an entry from the given dictionary based on the given key.
+        /// </summary>
+        /// <typeparam name="TKey">The datatype for the dictionary key.</typeparam>
+        /// <typeparam name="TValue">The datatype for the dictionary value.</typeparam>
+        /// <param name="dictionary">The dictionary to remove the entry from.</param>
+        /// <param name="key">The key to remove the entry for.</param>
+        /// <returns>Returns `true` if the entry for the given key was removed successfully, returns `false` if the entry for the given key was not found in the dictionary.</returns>
+        public static bool RemoveDictionaryKey<TKey, TValue>(ref Dictionary<TKey, TValue> dictionary, TKey key)
+        {
+            if (dictionary.ContainsKey(key))
```
bddckr: Remove this and change the whole method to
```c#
return dictionary.Remove(key);
```
instead. Or better: Remove this helper method altogether since it's now adding nothing to the existing method on `Dictionary<TValue>`.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -472,6 +469,109 @@ public static Vector3 AxisDirection(int axisIndex, Transform givenTransform = nu
...
+
+        /// <summary>
+        /// The RemoveDictionaryKey method attempts to remove an entry from the given dictionary based on the given key.
+        /// </summary>
+        /// <typeparam name="TKey">The datatype for the dictionary key.</typeparam>
+        /// <typeparam name="TValue">The datatype for the dictionary value.</typeparam>
+        /// <param name="dictionary">The dictionary to remove the entry from.</param>
+        /// <param name="key">The key to remove the entry for.</param>
+        /// <returns>Returns `true` if the entry for the given key was removed successfully, returns `false` if the entry for the given key was not found in the dictionary.</returns>
+        public static bool RemoveDictionaryKey<TKey, TValue>(ref Dictionary<TKey, TValue> dictionary, TKey key)
```
bddckr: Why is `dictionary` a `ref`? If you change this do a null check like in the other dictionary method.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
thestonefox: Because you're passing a list that needs changing outside of the scope of the method.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
thestonefox: iirc hash sets are only more performant at higher record counts, and lower records a list should outperform a hashset

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
bddckr: You are changing the list, but you aren't changing the reference AFAICS. I see no need to use `ref` here.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
thestonefox: the `setValue` param could be used here as it is already defaulted to the `default` value anyway.

i dunno if it's accurate to say it has lots of confusing params, two ambiguous at best ;)

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
thestonefox: if the key doesn't exist then won't that throw an exception?

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
thestonefox: If the value doesn't exist then won't that throw an exception?

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
bddckr: That may be true but the difference is either negligible or the added readability and transparency on the handling of duplicates is worth it in my opinion.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
bddckr: Nope, it will instead return true/false like you do:
https://msdn.microsoft.com/en-us/library/cd666k3e(v=vs.90).aspx

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
thestonefox: But then you have to build a custom property drawer for Unity as it doesn't support HashSets.

It's a wibbly wobbly world of it's own! :)

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
bddckr: I agree. Maybe just use that parameter then and add it to its documentation. (I can't think of a better name than `defaultValue` and `setDefaultValue` as a replacement for the existing parameter names.)

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
bddckr: Nope, it will instead return true/false like you do:
https://msdn.microsoft.com/en-us/library/kabs04ac(v=vs.90).aspx

thestonefox: Ok cool, so this method can be removed then in place of `dictionary.Remove(key)`

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
thestonefox: same as the dictionary then, remove in place of `list.Remove(value)`

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
thestonefox: i suppose list is already being passed as a reference because c# passes objects as references by default right?

bddckr: Exactly! You only need `ref` if you want to change the reference the caller has to a completely different list.

thestonefox:
@bddckr I've updated and removed the `Remove` methods.

I've also made the `Add` methods more similar by having them both return `true` if the method actually did add something to the collection.

I've swapped the params around on the `GetDictionaryValue` method and renamed the `setValue` to `defaultValue`. It means you don't have to provide all of the values all of the time.

thestonefox:
@bddckr I've had another bit of a refactor

renamed `forceAdd` on `AddListValue` to `preventDuplicates` and it's `false` by default, meaning the list will act as a normal `List<>` by adding duplicates, unless you pass `true` which will then prevent duplicates in the list.

I've also renamed `forceAdd` on `AddDictionaryValue` to `overwriteExisting` as it wasn't forcibly adding another element to the dictionary, it's actually just forcing it to overwrite the existing value if the key already exists. `false` there won't overwrite the existing value if the key is already present.

thoughts?

bddckr:
@thestonefox Looks good to me. 👍

thestonefox:
Bonanza!

thestonefox:
@bddckr I can't help but keep tinkering with this one!!

So, I went back over the code and looked at what "actually" needed to be a list that didn't accept duplicates. And there are some places that need to be a list (either because we use it as a public variable and Unity doesn't have an inspector for a HashSet, or the order is actually important but duplicates are not). But there are also a bunch of places where duplicates shouldn't exist but order doesn't really matter because we're just doing a `contains` check on the set.

So.... I've gone through again, and replaced a bunch of `List` uses with the `HashSet` anywhere i felt it is more appropriate.

Thoughts again?! :)

PR #1639, opened 2017-11-29
fix(SDK): update Google VR SDK vrDeviceName

tnterdan:
The latest version of the Google VR SDK (v1.110.0 released 2017-11-08) uses the string constant "daydream" (as opposed to "Daydream") as the `VRSettings.loadedDeviceName`. See `GoogleVR/Scripts/GvrSettings.cs` in the Google VR SDK.

Let me know if this fix should go into master instead. I was concerned about breaking the current build for those using an older version of Google VR SDK, but I am not sure if that matters too much since Daydream support is currently experimental anyway.

### DIFF
--- a/Assets/VRTK/Source/SDK/Daydream/SDK_DaydreamSystem.cs
+++ b/Assets/VRTK/Source/SDK/Daydream/SDK_DaydreamSystem.cs
@@ -4,7 +4,7 @@ namespace VRTK
     /// <summary>
     /// The Daydream System SDK script provides dummy functions for system functions.
     /// </summary>
-    [SDK_Description("Google Daydream (Android:Daydream)", SDK_DaydreamDefines.ScriptingDefineSymbol, "Daydream", "Android")]
+    [SDK_Description("Google Daydream (Android:Daydream)", SDK_DaydreamDefines.ScriptingDefineSymbol, "daydream", "Android")]
     public class SDK_DaydreamSystem
 #if VRTK_DEFINE_SDK_DAYDREAM
         : SDK_BaseSystem
--- a/Assets/VRTK/Source/SDK/Unity/SDK_UnitySystem.cs
+++ b/Assets/VRTK/Source/SDK/Unity/SDK_UnitySystem.cs
@@ -6,8 +6,8 @@ namespace VRTK
     /// </summary>
     [SDK_Description("Unity (Standalone:Oculus)", null, "Oculus", "Standalone")]
     [SDK_Description("Unity (Standalone:OpenVR)", null, "OpenVR", "Standalone", 1)]
-    [SDK_Description("Unity (Android:Cardboard)", null, "Cardboard", "Android", 2)]
-    [SDK_Description("Unity (Android:Daydream)", null, "Daydream", "Android", 3)]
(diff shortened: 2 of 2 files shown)


thestonefox:
I wonder if there is still any need for the Daydream SDK plugin for VRTK. You can just use the built in Unity SDK in 3.3.0

What are the benefits of keeping the core Daydream SDK support?

tnterdan:
@thestonefox Hmm, do you mean that essentially the Google VR SDK may be bundled with Unity? If so, I think the name of the built-in SDK still refers to "daydream" rather than "Daydream" if I'm reading [this](https://docs.unity3d.com/Manual/VRDevices-GoogleVR.html) right:

> * Use the device names daydream and cardboard to load a specific device when you want to enable XR for that device. To do this, call VRSettings.LoadDeviceByName and pass in the string name of the device.

I can do some more investigation tomorrow.

tnterdan:
@thestonefox Oops you're right, looks like I need to change it in both the Unity SDK and Google VR SDK. I somehow missed that.

Updated and tested by making a new project with only VRTK and using the built in Unity SDK for Daydream.

I am not sure about Ximmerse; I don't know how to test that one. I didn't want to change that one unless someone can verify it works.

tnterdan:
@thestonefox Is there anything else I need to do before this is ready to be merged? I believe I have addressed your comment.

thestonefox:
I'm a bit concerned that the change will break any backward compatibility. If older versions still require the uppercase D then anyone still using the older version of the google SDK will get breaking changes.

Also, what about google Cardboard, does that need changing to a lowercase `c` ?

Finally the commit message needs changing.

the scope should be SDK and the message should start lowercase like this:

`fix(SDK): update Google VR SDK vrDeviceName`

tnterdan:
@thestonefox Yep, Cardboard also needed updating! Tested and updated my commit, as well as the commit message. Thanks for the comments.

I agree that the breaking change can be a concern; at the very least it makes sense to not push this change to master and only target the alpha branch. While I could download an older version of the Google SDK for Daydream without this change, I think I would have to downgrade Unity from 2017.2 to use an older Cardboard SDK which is not very feasible. I am currently using this branch just to get my game to work on Daydream rather than downgrade to an older Google SDK.

If you want to play it safe and have backwards compatibility for users who are having trouble upgrading, then I am all ears too. I'm having a hard time finding a list of the VR SDK versions bundled with Unity, so I wouldn't really know which version of Unity should target which SDK, so if we want to do this, it may have to be through trial and error. But then this fix could go into master.

thestonefox:
@tnterdan what version of Google SDK introduces the lowercase breaking change?

@bddckr is there any way we could wrap this up with a script define symbol? (I am loathed to use script define symbols in the Unity plug in though)

bddckr:
@thestonefox Yes, either we use one if the Google SDK had one for that other/new version, or we define our own one using the same stuff we use in the `*Defines` classes: Reflection to check for existence of any symbol (type/method/whatever) to identify the old/new version and automatically add a scripting define symbol based on that.

thestonefox:
This was also missing the Ximmerse stuff, I've raised a new PR and used the new co-author system on github to give credit

https://github.com/thestonefox/VRTK/pull/1760

PR #1645, opened 2017-11-30
fix(SharedMethods): change FindEvenInactive to search in loaded scenes

### LINKED ISSUE #1580: FindEvenInactiveComponent(s) only searches in active scene
### Environment

 * Source of VRTK: Asset Store
 * Version of VRTK: 3.2.1
 * Version of Unity3D: 2017.1.0p5
 * Hardware used: None
 * SDK used: Simulator SDK

### Steps to reproduce

Use VRTK_SharedMethods.FindEvenInactiveComponent, FindEvenInactiveComponents or FindEvenInactiveGameObject to find a GameObject or Component which is not in the active scene (with mutliple scenes loaded)

### Expected behavior

The GameObject or Component should be found.

### Current behavior

The methods dont find anything because they check for "gameObject.scene == activeScene".

All 3 methods can only find objects in the currently active scene which leads to errors e.g. if the SimulatorRig is not located in the active scene because it will not find itself.

### Possible solution

I would recommend using [this solution](http://answers.unity3d.com/answers/1272001/view.html) to find (even inactive) Components in all loaded Scenes.

Peaj:
Placing the SDKManger in any other then the active scene led to
multiple errors because FindEvenInactiveComponent,
FindEvenInactiveComponents and FindEvenInactiveGameObject
only searched in the currently active scene.

Now it will search in all currently loaded scenes - fixes #1580

### DIFF
--- a/Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
+++ b/Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
@@ -309,7 +309,7 @@ namespace VRTK
         /// Finds the first <see cref="GameObject"/> with a given name and an ancestor that has a specific component.
         /// </summary>
         /// <remarks>
-        /// This method returns active as well as inactive <see cref="GameObject"/>s in the scene. It doesn't return assets.
+        /// This method returns active as well as inactive <see cref="GameObject"/>s in all loaded scenes. It doesn't return assets.
         /// For performance reasons it is recommended to not use this function every frame. Cache the result in a member variable at startup instead.
         /// </remarks>
         /// <typeparam name="T">The component type that needs to be on an ancestor of the wanted <see cref="GameObject"/>. Must be a subclass of <see cref="Component"/>.</typeparam>
@@ -323,14 +323,7 @@ namespace VRTK
                 return foundComponent == null ? null : foundComponent.gameObject;
             }
 
-            Scene activeScene = SceneManager.GetActiveScene();
-            IEnumerable<GameObject> gameObjects = Resources.FindObjectsOfTypeAll<T>()
-                                                           .Select(component => component.gameObject)
-                                                           .Where(gameObject => gameObject.scene == activeScene);
-
-#if UNITY_EDITOR
-            gameObjects = gameObjects.Where(gameObject => !AssetDatabase.Contains(gameObject));
-#endif
+            IEnumerable<GameObject> gameObjects = FindEvenInactiveComponents<T>().Select(component => component.gameObject);
 
             return gameObjects.Select(gameObject =>
                               {
@@ -344,41 +337,59 @@ namespace VRTK
         /// Finds all components of a given type.
         /// </summary>
         /// <remarks>
-        /// This method returns components from active as well as inactive <see cref="GameObject"/>s in the scene. It doesn't return assets.
+        /// This method returns components from active as well as inactive <see cref="GameObject"/>s in all loaded scenes. It doesn't return assets.
         /// For performance reasons it is recommended to not use this function every frame. Cache the result in a member variable at startup instead.
         /// </remarks>
         /// <typeparam name="T">The component type to search for. Must be a subclass of <see cref="Component"/>.</typeparam>
         /// <returns>All the found components. If no component is found an empty array is returned.</returns>
         public static T[] FindEvenInactiveComponents<T>() where T : Component
         {
-            Scene activeScene = SceneManager.GetActiveScene();
-            return Resources.FindObjectsOfTypeAll<T>()
-                            .Where(@object => @object.gameObject.scene == activeScene)
-#if UNITY_EDITOR
-                            .Where(@object => !AssetDatabase.Contains(@object))
-#endif
-                            .ToArray();
+            List<T> results = new List<T>();
+            for (int sceneIndex = 0; sceneIndex < SceneManager.sceneCount; sceneIndex++)
+            {
+                Scene scene = SceneManager.GetSceneAt(sceneIndex);
+                if (scene.isLoaded)
+                {
+                    GameObject[] rootObjects = scene.GetRootGameObjects();
+                    for (int rootObjectIndex = 0; rootObjectIndex < rootObjects.Length; rootObjectIndex++)
+                    {
+                        GameObject rootObject = rootObjects[rootObjectIndex];
+                        results.AddRange(rootObject.GetComponentsInChildren<T>(true));
+                    }
+                }
+            }
+            return results.ToArray();
         }
 
         /// <summary>
         /// Finds the first component of a given type.
         /// </summary>
         /// <remarks>
-        /// This method returns components from active as well as inactive <see cref="GameObject"/>s in the scene. It doesn't return assets.
+        /// This method returns components from active as well as inactive <see cref="GameObject"/>s in all loaded scenes. It doesn't return assets.
         /// For performance reasons it is recommended to not use this function every frame. Cache the result in a member variable at startup instead.
         /// </remarks>
         /// <typeparam name="T">The component type to search for. Must be a subclass of <see cref="Component"/>.</typeparam>
         /// <returns>The found component. If no component is found <see langword="null"/> is returned.</returns>
         public static T FindEvenInactiveComponent<T>() where T : Component
         {
-            Scene activeScene = SceneManager.GetActiveScene();
-            return Resources.FindObjectsOfTypeAll<T>()
-                            .Where(@object => @object.gameObject.scene == activeScene)
-#if UNITY_EDITOR
-                            .FirstOrDefault(@object => !AssetDatabase.Contains(@object));
-#else
-                            .FirstOrDefault();
-#endif
+            for (int sceneIndex = 0; sceneIndex < SceneManager.sceneCount; sceneIndex++)
+            {
+                Scene scene = SceneManager.GetSceneAt(sceneIndex);
+                if (scene.isLoaded)
+                {
+                    GameObject[] rootObjects = scene.GetRootGameObjects();
+                    for (int rootObjectIndex = 0; rootObjectIndex < rootObjects.Length; rootObjectIndex++)
+                    {
+                        GameObject rootObject = rootObjects[rootObjectIndex];
+                        T component = rootObject.GetComponentInChildren<T>(true);
+                        if(component != null)
+                        {
+                            return component;
+                        }
+                    }
+                }
+            }
+            return null;
         }
 
         /// <summary>


Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -344,41 +337,59 @@ public static float Mod(float a, float b)
...
+            {
+                var scene = SceneManager.GetSceneAt(i);
+                if (scene.isLoaded)
+                {
+                    var rootObjects = scene.GetRootGameObjects();
+                    for (int j = 0; j < rootObjects.Length; j++)
+                    {
+                        var go = rootObjects[j];
+                        var component = go.GetComponentInChildren<T>(true);
+                        if(component)
```
bddckr: [This check isn't doing what you expect.](https://blogs.unity3d.com/2014/05/16/custom-operator-should-we-keep-it/) Use the explicit `if (component != null)` instead.

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -323,14 +323,7 @@ public static float Mod(float a, float b)
...
 
-            Scene activeScene = SceneManager.GetActiveScene();
-            IEnumerable<GameObject> gameObjects = Resources.FindObjectsOfTypeAll<T>()
-                                                           .Select(component => component.gameObject)
-                                                           .Where(gameObject => gameObject.scene == activeScene);
-
-#if UNITY_EDITOR
-            gameObjects = gameObjects.Where(gameObject => !AssetDatabase.Contains(gameObject));
-#endif
+            var gameObjects = FindEvenInactiveComponents<T>().Select(component => component.gameObject);
```
thestonefox: Don't use `var` keyword, be explicit about the type

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -344,41 +337,59 @@ public static float Mod(float a, float b)
...
-            return Resources.FindObjectsOfTypeAll<T>()
-                            .Where(@object => @object.gameObject.scene == activeScene)
-#if UNITY_EDITOR
-                            .Where(@object => !AssetDatabase.Contains(@object))
-#endif
-                            .ToArray();
+            List<T> results = new List<T>();
+            for (int i = 0; i < SceneManager.sceneCount; i++)
+            {
+                var scene = SceneManager.GetSceneAt(i);
```
thestonefox: Don't use `var` not clear as to what the variable type is.

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -344,41 +337,59 @@ public static float Mod(float a, float b)
...
+                if (scene.isLoaded)
+                {
+                    var rootObjects = scene.GetRootGameObjects();
```
thestonefox: Don't use var not clear as to what the variable type is.

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -344,41 +337,59 @@ public static float Mod(float a, float b)
...
+                    for (int j = 0; j < rootObjects.Length; j++)
+                    {
+                        var go = rootObjects[j];
```
thestonefox: Don't use var not clear as to what the variable type is.

Also `go` is not a readable variable name

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -344,41 +337,59 @@ public static float Mod(float a, float b)
...
-            return Resources.FindObjectsOfTypeAll<T>()
-                            .Where(@object => @object.gameObject.scene == activeScene)
-#if UNITY_EDITOR
-                            .FirstOrDefault(@object => !AssetDatabase.Contains(@object));
-#else
-                            .FirstOrDefault();
-#endif
+            for (int i = 0; i < SceneManager.sceneCount; i++)
+            {
+                var scene = SceneManager.GetSceneAt(i);
```
thestonefox: Don't use var not clear as to what the variable type is.

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -344,41 +337,59 @@ public static float Mod(float a, float b)
...
+                if (scene.isLoaded)
+                {
+                    var rootObjects = scene.GetRootGameObjects();
```
thestonefox: Don't use var not clear as to what the variable type is.

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -344,41 +337,59 @@ public static float Mod(float a, float b)
...
+                    for (int j = 0; j < rootObjects.Length; j++)
+                    {
+                        var go = rootObjects[j];
```
thestonefox: Don't use var not clear as to what the variable type is.

Also `go` is not a readable variable name

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -344,41 +337,59 @@ public static float Mod(float a, float b)
...
+                        var component = go.GetComponentInChildren<T>(true);
```
thestonefox: Don't use var not clear as to what the variable type is.

thestonefox:
You need to squash your commit, you don't want a commit that is fixing things in the review. You want to either edit your previous commit or squash those two commits so it's in one commit

Peaj:
Ok I squashed those 2 commits into one. I hope it is correct now. Is this a standard github workflow, because I did not find it in either your or githubs Fork guide?

thestonefox:
@Peaj yes it's standard practice, the contributing guidelines for vrtk state

`Make commits of logical units.`

Which would mean each commit is it's own specific piece of work focusing on a specific result.

Having two commits, 1 that does some work, then another that fixes some stuff on the first commit before anything has been merged isn't a logical unit of work.

The logical step here is to have 1 commit that implements the fix outlined (therefore squashing the commits is the correct course of action).

Hope that helps.

thestonefox:
All the code looks fine, but what if you actually do only want to search for a GameObject in the current loaded scene? What about in the current example scenes if you switch scene then won't this look up the first GameObject it finds from the previous scene?

I wonder if it's better to either have a flag on the method or two methods?

Peaj:
I guess there is a slight misconception about what "**loaded** scenes" are.

The example scenes and scene switching works just like before as they do not load additively and
there is only one **loaded** scene at any time.
Scenes may stay in RAM as long as they arent explicitly disposed, but thats not what Unitys **loaded** flag refers to. The previous scenes are not **loaded** anymore.

The new implementation does more closely reflect the behaviour of Unitys FindObjectOfType<> method (which sadly does not have a flag to search for inactive objects)

If you load multiple scenes at once with additive loading you get multiple **loaded** scenes,
but only one of those will be the **active** scene (displayed with a bold scene title in the editor)

The previous implementation would break if you put the VRTK rig in any other than the **active** scene,
because it only searches in the **active** scene no matter where you are searching from.

The new implementation will find it if it is in a any **loaded** scene like FindObjectOfType<>.

Regarding your last question I would see two options for additional methods/flags now:

1. Find object in **active** scene (which reflects the previous implementation)
2. Find object in current/specific scene (would require a scene or a gameObject to be passed)

However I dont see where those methods would be usefull atm.

thestonefox:
ok @Peaj understood

let's see if @bddckr has some time to look over it and see if he has any comments too.

Peaj:
Any news on this @bddckr ?

bddckr:
Sorry guys I'm way too busy and totally didn't see the notification mail.
I'm fine with merging this as is, it's not like we can't ever revert or touch anything again.

Just keep in mind that this doesn't actually solve the issue everywhere - some VRTK scripts still won't be happy with cross-scene references, especially when unloading/switching scenes. But that's another issue I'd say.

kristianpd:
I've also run in to this issue in my first attempts to make my VR prefab stuff `DontDestroyOnLoad`. Since this was moved to a separate additive scene, the `SDK_InputSimulator` was dying for reasons above.

I don't have a strong opinion yet about whether or not searching ALL added scenes is good or bad, but this looks like it would solve my issue.

bddckr:
@kristianpd It won't, the scene `DontDestroyOnLoad` uses isn't accessible. (It is via hacks in the Editor only, in runtime it's not accessible at all.) See https://github.com/thestonefox/VRTK/issues/1316

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -344,41 +337,59 @@ public static float Mod(float a, float b)
...
-                            .Where(@object => !AssetDatabase.Contains(@object))
-#endif
-                            .ToArray();
+            List<T> results = new List<T>();
+            for (int sceneIndex = 0; sceneIndex < SceneManager.sceneCount; sceneIndex++)
+            {
+                Scene scene = SceneManager.GetSceneAt(sceneIndex);
+                if (scene.isLoaded)
+                {
+                    GameObject[] rootObjects = scene.GetRootGameObjects();
```
kristianpd: Isn't this going to only iterate over the root objects of the scene instead of all child objects as well? I feel like this is different from `Resources.FindObjectsOfTypeAll` that I blieve looks at absolutely everything.

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -344,41 +337,59 @@ public static float Mod(float a, float b)
...
+        /// This method returns components from active as well as inactive <see cref="GameObject"/>s in all loaded scenes. It doesn't return assets.
         /// For performance reasons it is recommended to not use this function every frame. Cache the result in a member variable at startup instead.
         /// </remarks>
         /// <typeparam name="T">The component type to search for. Must be a subclass of <see cref="Component"/>.</typeparam>
         /// <returns>All the found components. If no component is found an empty array is returned.</returns>
         public static T[] FindEvenInactiveComponents<T>() where T : Component
         {
-            Scene activeScene = SceneManager.GetActiveScene();
-            return Resources.FindObjectsOfTypeAll<T>()
-                            .Where(@object => @object.gameObject.scene == activeScene)
```
kristianpd: Why not just get rid of this clause instead of changing the implementation?

Assets/VRTK/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -344,41 +337,59 @@ public static float Mod(float a, float b)
...
+        /// This method returns components from active as well as inactive <see cref="GameObject"/>s in all loaded scenes. It doesn't return assets.
         /// For performance reasons it is recommended to not use this function every frame. Cache the result in a member variable at startup instead.
         /// </remarks>
         /// <typeparam name="T">The component type to search for. Must be a subclass of <see cref="Component"/>.</typeparam>
         /// <returns>The found component. If no component is found <see langword="null"/> is returned.</returns>
         public static T FindEvenInactiveComponent<T>() where T : Component
         {
-            Scene activeScene = SceneManager.GetActiveScene();
-            return Resources.FindObjectsOfTypeAll<T>()
-                            .Where(@object => @object.gameObject.scene == activeScene)
```
kristianpd: same question and concern (below) as above

kristianpd:
This PR can be closed, #1725 addresses this problem

PR #1676, opened 2017-12-29
fix(ui canvas): check for missing camera

interillusion:
When working with additive scenes there is null exception if:
-  first scene has a UICanvas, but no camera
- second scene is the one with a camera

As soon as the first scene is loaded, the raycast gives an error
because trying to do an update but there is no camera from 
where to start the raycast.

The error stops after the second scene is loaded.
- Verified that adding a camera to the first prevents the error

I checked the implementation of the unity GraphicRaycaster
and they have a check for the eventCamera been null,
in which case, they use a width/height of the display.


GitCop:

 * Commit: 728c53479919add28e354f9474f805ecba850ef8

---

thestonefox:
This PR had a bunch of commit history issues that needed reverting.

but to save you the hassle i just did a commit with this fix and credited you in the commit message

https://github.com/thestonefox/VRTK/pull/1678

PR #1706, opened 2018-01-18
feat(Interation-Locomotion): trigger swipe event on touchpad

skad:
Add Swipe event on VRTK_ControllerEvents
Made Touchpad control using horizontal and vertical swipe
Update the 017_CameraRig_TouchpadWalking with a second
TouchpadControlOption using swipe instead of Axis Change


thestonefox:

Issue #1723, opened 2018-02-08
VRTK_BaserPointerRenderer potentially references objects that can be destroyed


kristianpd:
### Environment

 * Source - Github
 * Branch - 3,3,0-alpha
 * Unity Version -2017.3
 * Platform - Any
 * SDK - SteamVR, Simulator

### Steps to reproduce

The base pointer renderers are created at runtime in your active scene. If that scene is unloaded (additive scene loading), the transform follow component on those objects keeps trying to reference them even though they are destroyed. 

This can be more simply reproduced in a teleport example scene by deleting the `Base_PointerRender` objects at runtime with the simulator.

Here's a video: https://youtu.be/sqqmK-K28aA

### Expected behavior

It should not keep attempting to access the transform follow component during fixed update.

I think it should reset the transform follow component when pointers are Reset (destroyed and created again). 

### Current behavior

When unloading a scene in which VRTK initialized the PointerRenders, the `VRTK_TransformFollow` reference is kept around even though it is no longer valid (object destroyed). The `Base_PointerRenderer` objects are created in the currently active scene, which in my case, is a different scene from the one in which my SDKs are initialized (additive scene load that I keep around between switching active scenes). 

https://github.com/thestonefox/VRTK/blob/release/3.3.0-alpha/Assets/VRTK/Source/Scripts/Pointers/PointerRenderers/VRTK_BasePointerRenderer.cs#L377

Is the line that raises the exception. This is because `pointerOriginTransformFollow` references a component that was destroyed when the object was destroyed.

PR #1724, opened 2018-02-08
fix(Pointers): handle destroyed origin transform follow

kristianpd:
## Problem

Proposed fix for https://github.com/thestonefox/VRTK/issues/1723

## Details
I can't think of a great way to know this object was destroyed from inside `VRTK_BasePointerRenderer` but the null check on the component reference does work.

This also recreates the references when the `ResetPointerObjects` metrhod is called. This is the hook I use on scene unload/load to re-create the objects. 

I know deleting these objects is unlikely in a single scene setup, but it's a quick way to see the error. 

## Alternative

Another possible solution here would be to create the objects in the same scene as the controllers and assume that would never be deleted. The benefit to this would be that it wouldn't require the call to `ResetPointers`, though that change may be warranted regardless? 

It would however require that I change the scene of the auto created objects: https://github.com/thestonefox/VRTK/blob/release/3.3.0-alpha/Assets/VRTK/Source/Scripts/Pointers/PointerRenderers/VRTK_BasePointerRenderer.cs#L606

and change scenes on the object if it does not reside in the same scene as the controller

### DIFF
--- a/Assets/VRTK/Source/Scripts/Pointers/PointerRenderers/VRTK_BasePointerRenderer.cs
+++ b/Assets/VRTK/Source/Scripts/Pointers/PointerRenderers/VRTK_BasePointerRenderer.cs
@@ -153,7 +153,9 @@ namespace VRTK
         /// </summary>
         public virtual void ResetPointerObjects()
         {
+            DestroyPointerOriginTransformFollow();
             DestroyPointerObjects();
+            CreatePointerOriginTransformFollow();
             CreatePointerObjects();
         }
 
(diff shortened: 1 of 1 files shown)


Assets/VRTK/Source/Scripts/Pointers/PointerRenderers/VRTK_BasePointerRenderer.cs
```diff
@@ -154,6 +154,7 @@ public virtual void InitalizePointer(VRTK_Pointer givenPointer, VRTK_PolicyList
         public virtual void ResetPointerObjects()
         {
             DestroyPointerObjects();
+            CreatePointerOriginTransformFollow();
```
kristianpd: Thinking about this more, I should have a `DestroyPointerRendererObjects` here too, to prevent multiple from being created

thestonefox:
do a `git commit --amend` and change the scope from `BasePointerRenderer` to `Pointers` so it's the same as the other pointer commits (makes it easier to group changes for the changelog)

kristianpd:
@thestonefox doine

PR #1725, opened 2018-02-08
fix(SharedMethods): change FindEvenInactive to search all loaded scenes

kristianpd:
This is an updated proposed fix for https://github.com/thestonefox/VRTK/issues/1580.

It builds on the work done by @peaj in this PR https://github.com/thestonefox/VRTK/pull/1645 with a few minor changes.  

# Problem

The problem is the same, I'm loading scenes additively and VRTK is only searching the active scene. 

# Solution

This PR fixes the problem (as originally proposed by @peaj) by iterating through all active scenes and searching for the component rather than only the active scene. 

# The change

The main difference in this version vs the one done by @peaj in https://github.com/thestonefox/VRTK/pull/1645 is just that I use a common private method with an option to avoid duplicating the code and leaving room for bugs in that sense.

This PR also remains backwards compatible with the `VRTK_SharedMethods` API, only searching the active scene by default while giving an option to search all scenes if needed.

### DIFF
--- a/Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
+++ b/Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
@@ -251,76 +251,59 @@ namespace VRTK
         /// Finds the first GameObject with a given name and an ancestor that has a specific component.
         /// </summary>
         /// <remarks>
-        /// This method returns active as well as inactive GameObjects in the scene. It doesn't return assets.
+        /// This method returns active as well as inactive GameObjects in all scenes. It doesn't return assets.
         /// For performance reasons it is recommended to not use this function every frame. Cache the result in a member variable at startup instead.
         /// </remarks>
         /// <typeparam name="T">The component type that needs to be on an ancestor of the wanted GameObject. Must be a subclass of `Component`.</typeparam>
         /// <param name="gameObjectName">The name of the wanted GameObject. If it contains a '/' character, this method traverses the hierarchy like a path name, beginning on the game object that has a component of type `T`.</param>
+        /// <param name="searchAllScenes">If this is true, all loaded scenes will be searched. If this is false, only the active scene will be searched.</param>
         /// <returns>The GameObject with name `gameObjectName` and an ancestor that has a `T`. If no such GameObject is found then `null` is returned.</returns>
-        public static GameObject FindEvenInactiveGameObject<T>(string gameObjectName = null) where T : Component
+        public static GameObject FindEvenInactiveGameObject<T>(string gameObjectName = null, bool searchAllScenes = false) where T : Component
         {
             if (string.IsNullOrEmpty(gameObjectName))
             {
-                T foundComponent = FindEvenInactiveComponent<T>();
+                T foundComponent = FindEvenInactiveComponentsInLoadedScenes<T>(searchAllScenes, true).FirstOrDefault();
                 return foundComponent == null ? null : foundComponent.gameObject;
             }
 
-            Scene activeScene = SceneManager.GetActiveScene();
-            IEnumerable<GameObject> gameObjects = Resources.FindObjectsOfTypeAll<T>()
-                                                           .Select(component => component.gameObject)
-                                                           .Where(gameObject => gameObject.scene == activeScene);
-
-#if UNITY_EDITOR
-            gameObjects = gameObjects.Where(gameObject => !AssetDatabase.Contains(gameObject));
-#endif
-
-            return gameObjects.Select(gameObject =>
-                              {
-                                  Transform transform = gameObject.transform.Find(gameObjectName);
-                                  return transform == null ? null : transform.gameObject;
-                              })
-                              .FirstOrDefault(gameObject => gameObject != null);
+            return FindEvenInactiveComponentsInLoadedScenes<T>(searchAllScenes)
+                       .Select(component => 
+                       {
+                            Transform transform = component.gameObject.transform.Find(gameObjectName);
+                            return transform == null ? null : transform.gameObject;
+                       })
+                       .FirstOrDefault(gameObject => gameObject != null);
         }
 
         /// <summary>
         /// Finds all components of a given type.
         /// </summary>
         /// <remarks>
-        /// This method returns components from active as well as inactive GameObjects in the scene. It doesn't return assets.
+        /// This method returns components from active as well as inactive GameObjects in all scenes. It doesn't return assets.
         /// For performance reasons it is recommended to not use this function every frame. Cache the result in a member variable at startup instead.
         /// </remarks>
         /// <typeparam name="T">The component type to search for. Must be a subclass of `Component`.</typeparam>
+        /// <param name="searchAllScenes">If this is true, all loaded scenes will be searched. If this is false, only the active scene will be searched.</param>
         /// <returns>All the found components. If no component is found an empty array is returned.</returns>
-        public static T[] FindEvenInactiveComponents<T>() where T : Component
+        public static T[] FindEvenInactiveComponents<T>(bool searchAllScenes = false) where T : Component
         {
-            Scene activeScene = SceneManager.GetActiveScene();
-            return Resources.FindObjectsOfTypeAll<T>()
-                            .Where(@object => @object.gameObject.scene == activeScene)
-#if UNITY_EDITOR
-                            .Where(@object => !AssetDatabase.Contains(@object))
-#endif
-                            .ToArray();
+            IEnumerable<T> results = FindEvenInactiveComponentsInLoadedScenes<T>(searchAllScenes);
+            return results.ToArray();
         }
 
         /// <summary>
         /// Finds the first component of a given type.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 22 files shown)


Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -261,66 +261,46 @@ public static float Mod(float a, float b)
...
-            return gameObjects.Select(gameObject =>
-                              {
-                                  Transform transform = gameObject.transform.Find(gameObjectName);
-                                  return transform == null ? null : transform.gameObject;
-                              })
-                              .FirstOrDefault(gameObject => gameObject != null);
+            return FindEvenInactiveComponentsInLoadedScenes<T>()
+                       .Select(component => 
+                       {
+                            Transform transform = component.gameObject.transform.Find(gameObjectName);
```
kristianpd: I've avoided the extra select on https://github.com/thestonefox/VRTK/pull/1725/files#diff-5fe0a238b22c48e3e2f459d8f78fbe2cL270 by simply referencing the `component.gameObject` here instead.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -634,5 +614,41 @@ public static BuildTargetGroup[] GetValidBuildTargetGroups()
...
+            List<T> results = new List<T>();
+            for (int sceneIndex = 0; sceneIndex < SceneManager.sceneCount; sceneIndex++)
+            {
+                Scene scene = SceneManager.GetSceneAt(sceneIndex);
+                if (scene.isLoaded)
+                {
+                    GameObject[] rootObjects = scene.GetRootGameObjects();
+                    foreach (GameObject rootObject in scene.GetRootGameObjects())
+                    {
+                        if (stopOnMatch)
```
kristianpd: The best thing I could come up with to avoid unnecessary iteration or extra array elements when looking for a single element. 

This is the main change from @peaj solution and lets us share this method between both `FindEvenInactiveComponents` and `FindEvenInactiveComponent`

thestonefox:
This seems to have the same issue as the other PR addressing the same problem.

The changes provide a different way of the method actually working, meaning if people are using them as a way of specifically only finding the object in the current scene then this change will break that.

As with the other PR, my proposal is to either pass a flag which defaults to keep existing behaviour or a separate method that finds in all scenes.

kristianpd:
Sorry i missed that comment.

I can add an optional bool parameter to search all scenes keeping API backwards compatible.

Alternatively i could make that flag a list of scenes to search instead but it may be overkill.

kristianpd:
ok @thestonefox @bddckr i've updated the PR to be backwards compatible in our `VRTK_SharedMethods` api, while switching VRTKs internal use of the methods to search all loaded scenes. 

This change is a bit more explicit in what it affects:

`VRTK_SDKManagerEditor`
`VRTK_SDKInputSimulator`
`SDK_UnityBoundaries`
`SDK_UnityController`
`SDK_UnityHeadset`
`VRTK_BaseOBjectControlAction`
`VRTK_SDKSetup`

I've looked through the code in each area, and I believe the change to search all loaded scenes is still what we want in each use case. 

I've tested this by going through each example scene to make sure there are no initialization errors. I've also tested it in my additive scene loading implementation and it works properly.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -634,5 +614,43 @@ public static BuildTargetGroup[] GetValidBuildTargetGroups()
...
+        /// <param name="stopOnMatch">If true, will stop searching objects as soon as a match is found.</param>
+        /// <returns></returns>
+        private static IEnumerable<T> FindEvenInactiveComponentsInLoadedScenes<T>(bool searchAllScenes, bool stopOnMatch = false) where T : Component
+        {
+            List<T> results = new List<T>();
+
+            for (int sceneIndex = 0; sceneIndex < SceneManager.sceneCount; sceneIndex++)
+            {
+                Scene scene = SceneManager.GetSceneAt(sceneIndex);
+                if (scene.isLoaded && (searchAllScenes || scene == SceneManager.GetActiveScene()))
```
kristianpd: @thestonefox here is the specific change you requested

kristianpd:
Found some missing changes when testing in my own game. All of the `#ifdef` chunks of code didn't originally show up when I found by references.

I can at least confirm the api was backwards compatible, as my game was not finding the steamvr setup when additively loaded.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -251,76 +251,56 @@ public static float Mod(float a, float b)
...
+        /// This method returns active as well as inactive GameObjects in all scenes. It doesn't return assets.
         /// For performance reasons it is recommended to not use this function every frame. Cache the result in a member variable at startup instead.
         /// </remarks>
         /// <typeparam name="T">The component type that needs to be on an ancestor of the wanted GameObject. Must be a subclass of `Component`.</typeparam>
         /// <param name="gameObjectName">The name of the wanted GameObject. If it contains a '/' character, this method traverses the hierarchy like a path name, beginning on the game object that has a component of type `T`.</param>
         /// <returns>The GameObject with name `gameObjectName` and an ancestor that has a `T`. If no such GameObject is found then `null` is returned.</returns>
-        public static GameObject FindEvenInactiveGameObject<T>(string gameObjectName = null) where T : Component
+        public static GameObject FindEvenInactiveGameObject<T>(string gameObjectName = null, bool searchAllScenes = false) where T : Component
```
thestonefox: There's no update to the comment to mention the new parameter.

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -251,76 +251,56 @@ public static float Mod(float a, float b)
...
         /// </summary>
         /// <remarks>
-        /// This method returns components from active as well as inactive GameObjects in the scene. It doesn't return assets.
+        /// This method returns components from active as well as inactive GameObjects in all scenes. It doesn't return assets.
         /// For performance reasons it is recommended to not use this function every frame. Cache the result in a member variable at startup instead.
         /// </remarks>
         /// <typeparam name="T">The component type to search for. Must be a subclass of `Component`.</typeparam>
         /// <returns>All the found components. If no component is found an empty array is returned.</returns>
-        public static T[] FindEvenInactiveComponents<T>() where T : Component
+        public static T[] FindEvenInactiveComponents<T>(bool searchAllScenes = false) where T : Component
```
thestonefox: No docs for new parameter

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -251,76 +251,56 @@ public static float Mod(float a, float b)
...
         /// </summary>
         /// <remarks>
-        /// This method returns components from active as well as inactive GameObjects in the scene. It doesn't return assets.
+        /// This method returns components from active as well as inactive GameObjects in all scenes. It doesn't return assets.
         /// For performance reasons it is recommended to not use this function every frame. Cache the result in a member variable at startup instead.
         /// </remarks>
         /// <typeparam name="T">The component type to search for. Must be a subclass of `Component`.</typeparam>
         /// <returns>The found component. If no component is found `null` is returned.</returns>
-        public static T FindEvenInactiveComponent<T>() where T : Component
+        public static T FindEvenInactiveComponent<T>(bool searchAllScenes = false) where T : Component
```
thestonefox: No Docs for new parameter

Assets/VRTK/Source/Scripts/Utilities/VRTK_SharedMethods.cs
```diff
@@ -634,5 +614,43 @@ public static BuildTargetGroup[] GetValidBuildTargetGroups()
...
 #endif
+
+        /// <summary>
+        /// The FindEvenInactiveComponentsInLoadedScenes method searches active and inactive game objects in all
+        /// loaded scenes for components matching the type supplied. 
+        /// </summary>
+        /// <param name="searchAllScenes">If true, will search all loaded scenes, otherwise just the active scene.</param>
+        /// <param name="stopOnMatch">If true, will stop searching objects as soon as a match is found.</param>
+        /// <returns></returns>
+        private static IEnumerable<T> FindEvenInactiveComponentsInLoadedScenes<T>(bool searchAllScenes, bool stopOnMatch = false) where T : Component
```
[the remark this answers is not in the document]
kristianpd: It was meant to be a shared method used by the other `FindEvenInactive(s)`. Since it has the slightly awkward parameter of stop early, i wasn't a big fan of it being public. I know private static methods are a little weird, but I prefer this to it being public i think

thestonefox:
@kristianpd all looks ok code wise, i'll try and test it tomorrow and then look to merge it.

PR #1730, opened 2018-02-15
fix(Highlighters): use whiteTexture in place of new Texture

### LINKED ISSUE #1758: Unity 2018.1.0b12: Can't create empty Texture (not allowed, needs to be subclass)
This line causes projects to fail compilation:

https://github.com/thestonefox/VRTK/blob/e43089a4dbefba55af2716d22de31458e0b615ca/Assets/VRTK/Scripts/Interactions/Highlighters/VRTK_MaterialColorSwapHighlighter.cs#L152

This can be fixed by replacing the line above with:

`renderer.material.SetTexture("_MainTex", new Texture2D(0,0));`

Cheers!

kristianpd:
Fixes #1758 

Base class `Texture` constructor is inaccessible due to protection level.

Update to use `Texture2D.whiteTexture` which has been confirmed not to create new textures at runtime.

### DIFF
--- a/Assets/VRTK/Source/Scripts/Interactions/Highlighters/VRTK_MaterialColorSwapHighlighter.cs
+++ b/Assets/VRTK/Source/Scripts/Interactions/Highlighters/VRTK_MaterialColorSwapHighlighter.cs
@@ -160,7 +160,7 @@ namespace VRTK.Highlighters
 
                     if (resetMainTexture && material.HasProperty("_MainTex"))
                     {
-                        renderer.material.SetTexture("_MainTex", new Texture());
+                        renderer.material.SetTexture("_MainTex", Texture2D.whiteTexture);
                     }
 
                     if (material.HasProperty("_Color"))


bddckr:
We're never supporting betas of Unity. We tried in the past but they were changing APIs without any documentation every second to third beta release, so we don't play cat and mouse anymore.

So this will probably take a while to get merged as we'll be waiting for Unity to release 2018.1 first.

kristianpd:
Fair enough, I do suspect this change will remain, but may as well wait.

dustinfreeman:
@kristianpd this was helpful for me, thanks. I also agree that big third-party libraries like VRTK shouldn't be supporting betas until they're final; I've lost lots of time doing my own low-level work playing cat-and-mouse.

kristianpd:
This can be merged whenever, it's not specific to any 2018 API itself. Other examples of doing the same thing we use `Texture2D.whiteTexture` it's just that in this particular case someone had used `new Texture()`

bddckr:
This brings up an interesting point: Does `Texture2D.whiteTexture` always return the same instance or do we have to destroy the texture ourselves? I'm afraid we're leaking (tiny bits of) memory.

I can't find anything in the documentation.

kristianpd:
Reply here: https://github.com/thestonefox/VRTK/issues/1758#issuecomment-379369482

thestonefox:
Ok so I think this is the solution.

2 things need to be done to the PR if you could be so kind @kristianpd 

Firstly, it needs merging into `master` branch now as `3.3.0-alpha` has been merged.

Secondly change the commit message to:

```
fix(Highlighters): use whiteTexture in place of new Texture

Base class Texture constructor is now protected, use the helper
`Texture2D.whiteTexture` instead as this also doesn't create new
instances of a texture where as calling `new Texture` will create
new instances.
```

kristianpd:
@thestonefox all ready to go

PR #1732, opened 2018-02-18
chore(SharedMethods): remove unnecessary variable assignment

kristianpd:
As noticed by @virror, my bad for not catching this one.

@bddckr @thestonefox


thestonefox:
Fixed in https://github.com/thestonefox/VRTK/pull/1739

PR #1734, opened 2018-02-20
feat(Interactive haptics): Add interactive haptics

kristianpd:
## UPDATED! April 11, 2018

A long overdue update to #1513 

- improves variable naming, documentation, and tooltips. 
- improve examples
- simplifies interactions by just using `VRTK_BaseControllable`

### What's this?

This PR provides a solution for interactive / continuous haptics when interacting with `InteractableObject`s. The main issue I have with the current `InteractHaptics` implementation, is that it's  either a basic `OnX` event, or in the alpha release, the ability to provide a static pulse/interval until the interaction ends. 

This PR introduces the idea I'm currently calling "Haptic Curves". Haptic curves allow the developer to specify an interaction curve that will be applied to an `InteractableObject`. A basic example of this would be an old wooden Door that opens with a "tighter" feel and slowly loosens up as you pull it out from the frame.

There are two components to an interactive haptic. `VRTK_InteractiveHaptics` which specifies the strength and interval of the feedback over a fixed range and `VRTK_InteractiveHapticsInput` which gives a normalized value of `0..1` representing that range.

## VRTK_InteractiveHaptics
image

`InteractionType` - `WhileTouching`, `WhileGrabbing`, `WhileUsing`. Specifies the start/stop conditions for haptic feedback.
`Strength curve` - Given a value from 0 to 1, evaluate a strength along the curve. Here, we use a constant strength.
`Pulse Interval curve` - Given a value from 0 to 1, evaluate an interval along the curve. Here, we want it to feel "tighter" in the middle of the interaction
`Fastest Pulse Interval/Slowest Pulse Interval` - Specify the interval range the curve applies to. `Y` value of 1 on the curve, corresponds to the `Min` interval (the fastest one), 0, the `Max` interval.
`Sensitivity Threshold` - Currently a solution I'm using to prevent constant feedback since it's impossible to hold something completely still, this corresponds to 1% in this example.
`ObjectToAffect` - `VRTK_BaseControllable` reference, assumed to be on current object if not specified
`InteractableObject` - `VRTK_InteractableObject` reference used to determine the controller in grab interactions. Assumed to be on the parent if not specified.

### Example scene
I've updated the Artifical and Physics based interactions in the `[005 - Interactions] InteractableObjects` scene to use interactive haptics

### DIFF
--- a/Assets/VRTK/Source/Scripts/Interactions/VRTK_InteractiveHaptics.cs
+++ b/Assets/VRTK/Source/Scripts/Interactions/VRTK_InteractiveHaptics.cs
@@ -1,0 +1,175 @@
+﻿// Interactive Haptics|Interactions|30101
+namespace VRTK
+{
+    using UnityEngine;
+    using VRTK.Controllables;
+
+    /// <summary>
+    /// The VRTK_InteractiveHaptics script is attached on the same GameObject as an a VRTK_BaseControllable subclass. This includes artifical and physics based interactions. VRTK_InteractiveHaptics provides customizable haptic feedback curves for more realistic interactions over the fixed range of the VRTK_BaseControllable object.
+    /// </summary>
+    /// <remarks>
+    ///  * Add this script to the same game object on which a subclass of `VRTK_BaseControllable` resides. This script uses the `VRTK_BaseControllable` subclass to determine the range over which it will operate.
+    ///  * Optionally specify a `VRTK_InteractableObject`. This will be taken from the parent object at runtime unless it is specified. This object will be used to determine which controller is interacting with the object while using the `While Grabbing` `InteractionType`.
+    ///  </remarks>
+    [AddComponentMenu("VRTK/Scripts/Interactions/VRTK_InteractiveHaptics")]
+    public class VRTK_InteractiveHaptics : MonoBehaviour
+    {
+        /// <summary>
+        /// The types of interaction that can be performed with Interactive Haptics.
+        /// </summary>
+        public enum InteractionType
+        {
+            WhileGrabbing,
+            WhileTouching
+        }
+
+        [Header("Haptic Response")]
+
+        [Tooltip("The type of interaction on which to apply this curve. If multiple interactions are needed, add multiple InteractiveHaptics components.")]
+        public InteractionType interactionType;
+
+        [Tooltip("Denotes the curve in which the normalized input will be evaluated. The output of the curve is the strength of the haptic feedback from 0 to 1.")]
+        public AnimationCurve strengthCurve;
+
+        [Tooltip("Denotes the curve in which the normalized input will be evaluated. The output of the curve is the interval between fastestPulseInterval and slowestPulseInterval represented by the curve.")]
+        public AnimationCurve pulseIntervalCurve;
+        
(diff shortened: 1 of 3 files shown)


kristianpd:
@andybak @bkensey @thebigman @virror @thebigdev433 you've all made your way on to my list and once you're on you can't get off. I could use some help testing. Simply run scene 45 in this PR and interact with the objects you see!

@bddckr @thestonefox you may not remember this one from ages ago, but it's up to date now!

kristianpd:
Should I be updating the build settings to include the new scene as well or is that something you like to do?

thestonefox:
@kristianpd all of the example scenes are changing, so any new example scene you have in this PR will become obsolete and probably create a merge conflict

kristianpd:
Alright @thestonefox i'll keep an eye out for the changes

thebigdev433:
I can vouch for everything in the scene working properly. The haptics feel great and this definitely deserves to be in VRTK.

kristianpd:
I'm reworking this PR to work properly with the physics classes and integrating it into the new interactive objects scene,  will take a week or two probably

Assets/VRTK/Source/Scripts/Interactions/VRTK_InteractiveHaptics.cs
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        protected virtual void StartHaptics(object sender, InteractableObjectEventArgs e)
+        {
+            controller = VRTK_ControllerReference.GetControllerReference(e.interactingObject);
+        }
+
+        private void Interact(object sender, ControllableEventArgs e)
+        {
+            VRTK_ControllerReference controllerReference = null;
```
kristianpd: @thestonefox @bddckr maybe you guys could help me out here?

I need to figure out which controller is interacting with the `VRTK_BaseControllable` object at `Interact` time in order to send the pulse. When using rotators/sliders with grab, I seem to only be able to get the controller reference via the `VRTK_InteractableObject` as `interactingTouchScript` is `null`. When using the Pusher classes, `interactingTouchScript` is set, and I can use that to get the controller reference. 

If this was always set to the interacting controller alias, even on grab, then I could get rid of this conditional check as well as all of this:

https://github.com/thestonefox/VRTK/pull/1734/files#diff-5a5fc1a08788396924878f51d1472f61R101

Am i missing an easier solution here?


If

SimonDarksideJ:
OK, I tried testing from WinMR, but couldn't get any haptic feedback in the scene anywhere.
Tried again in OpenVR, but it seems controller rendering has stopped working there.  Need to go back and check master, as I'm sure it used to work.
Thoughts @thestonefox ?

kristianpd:
Controller rendering sometimes gets messed up for me when I turn it on at runtime (like one renders, the other doesn't). Did you try restarting and going straight into openvr / steamvr mode? I'm using the standard trigger pulse API, have we confirmed this works in WinMR?

kristianpd:
Would greatly appreciate some feedback on this PR. Anyone have a few minutes to test it out?

ntindle:
What is the expected behavior? I’ll be trying later tonight

thebigdev433:
Everything works fine for me. Would make a wonderful addition to vrtk.

ntindle:
Seems good to me

thestonefox:
When 3.3.0 is released, I recommend we look to just merge this (but not yet!)

kristianpd:
zoidberg_blinking_futurama

thestonefox:

PR #1737, opened 2018-02-26
fix(SDKManager): Unity 2018 Requires "None" Device to Be Last

MattOstgard:
As it is right now "None" will be put at the top of the list in
Player Settings > XR Devices > VR Enabled Devices. In 2017 this
seemed be bypassed by default, but in 2018 the assumed intent is to
not start in VR mode.

This fix puts "None" last when the settings are auto managed.


bddckr:
>the assumed intent is to not start in VR mode

This is also the case for any other Unity version. This is exactly what VRTK is intending.

VRTK is using the SDK Setups you configure in the SDK Manager's list for it one by one, top to bottom and therefore wants `None` first.

Nothing changed with Unity 2018 here and this PR actually breaks the SDK switching VRTK does.

MattOstgard:
Oh sorry about that. Closing.

Issue #1693, opened 2018-01-09
Unwanted positionchange through VRTK_RotateObjectControlAction


RunenGamer:
### Environment

 * Source of VRTK: Github
 * Version of VRTK: release/3.3.0-alpha (e43089a4dbefba55af2716d22de31458e0b615ca)
 * Version of Unity3D (e.g. Unity 2017.3.0f3)
 * Hardware used (e.g. Vive)
 * SDK used (e.g. SteamVR)

### Steps to reproduce

1.     Pull latest from feature branch 3.3.0-alpha
2.     Open Example scene 017_CameraRig_TouchpadWalking
3.     Enable SteamVR SDK (download, enable, populate)
4.     Select [VRTK_Scripts]>PlayArea>LeftController>TouchpadControlOptions
5.     Disable GameObject "XAxis = Slide  -  YAxis = Slide"
6.     Enable "XAxis = SnapRotate  -  YAxis = Slide" (or another rotate one)
7.     Start scene, use touchpad to rotate and observe global position of [VRTK_SDKManager]>SDKSetups>SteamVR>[CameraRig]>Camera (eye)

### Expected behavior

No matter where the real player stand inside of the PlayArea, a rotation should only rotate the player without changing player's global position. 

### Current behavior

Depending on where the real player stand inside of the PlayArea, the rotation not only rotates the player (Camera (eye)) but also changes player's global position (because [VRTK_SDKManager]>SDKSetups>SteamVR>[CameraRig] gets rotated). 

I would like to fix that by reposition [VRTK_SDKManager]>SDKSetups>SteamVR>[CameraRig] accordingly. 
Therefore I suggest to push to the master branch. 
If you do not consider this behaviour as a bug, I suggest to implement a feature (public flag on these scripts to enable repositioning) and target the release/3.3.0-alpha branch.

LuckyJosun:
I've noticed this in the current release. I believe the headset is being rotated using the playspace's centre point as a pivot.

So, if the headset is in one corner of the playspace, and snap rotate is triggered, the headset's position can be dramatically changed instantly -- in my case, this causes players to move completely through virtual walls rotating near them.

I'm using an Oculus CV1 w/ Touch Controllers, the current SteamVR Unity Plugin (1.2.3 aka 54), the current release of VRTK (3.2.0), and Unity 2017.3.0f3 Personal. Windows 10.

bddckr:
[...]

PR #1740, opened 2018-03-02
fix(Locomotion): fix player position offset after rotation

RunenGamer:
Player used to have an offset after rotation
through the scripts
VRTK_RotateObjectControlAction and
VRTK_SnapRotateObjectControlAction
(see #1693).
Gameobject [CameraRig] is moved after
rotation to compensate the offset.


RunenGamer:
@thestonefox (regarding #1693 ): I pushed to master, because I consider it a bug (according to the CONTRIBUTING.md). Do you need any further info to check the commit (when you have time)?

thestonefox:
you should be checking if this is fixed in 3.3.0 and if not, then push to 3.3.0 because you'll probably get merge conflicts otherwise

PR #1745, opened 2018-03-10
feat(SDK): add native Unity Windows Mixed Reality support

thestonefox:
Support for the Unity Windows Mixed Reality WSA namespace to allow
for Windows Mixed Reality headsets and controllers to be used within
VRTK.

Controller visualisation is provided by a 3rd party repository and
instructions of how to use are provided in the WindowsMR README.md
file.

A massive thanks to @tomwim for all his hard work on this and thanks
to @DDReaper for continued support along the way.

PR #1749, opened 2018-03-13
Feat/windows mixed reality support - UWP Build update

SimonDarksideJ:
Updated reflection code to also work for UWP platform using #if defines around divergent code.

### DIFF
--- a/Assets/VRTK/Source/SDK/Base/SDK_DescriptionAttribute.cs
+++ b/Assets/VRTK/Source/SDK/Base/SDK_DescriptionAttribute.cs
@@ -6,6 +6,9 @@ namespace VRTK
 #endif
     using System;
     using System.Linq;
+#if UNITY_WSA && !UNITY_EDITOR
+    using System.Reflection;
+#endif
 
     /// <summary>
     /// Describes a class that represents an SDK. Only allowed on classes that inherit from SDK_Base.
@@ -138,10 +141,17 @@ namespace VRTK
 
         public static SDK_DescriptionAttribute[] GetDescriptions(Type type)
         {
+#if UNITY_WSA && !UNITY_EDITOR
+            return type.GetTypeInfo().GetCustomAttributes(typeof(SDK_DescriptionAttribute), false)
+                       .Cast<SDK_DescriptionAttribute>()
+                       .OrderBy(attribute => attribute.index)
+                       .ToArray();
+#else
             return type.GetCustomAttributes(typeof(SDK_DescriptionAttribute), false)
                        .Cast<SDK_DescriptionAttribute>()
                        .OrderBy(attribute => attribute.index)
                        .ToArray();
+#endif
         }
     }
 }
--- a/Assets/VRTK/Source/Scripts/Utilities/SDK/VRTK_SDKManager.cs
+++ b/Assets/VRTK/Source/Scripts/Utilities/SDK/VRTK_SDKManager.cs
@@ -641,6 +641,7 @@ namespace VRTK
                     setup => setup.usedVRDeviceNames.Contains(XRSettings.loadedDeviceName)
                 );
             }
+#if !UNITY_WSA
             else
             {
                 // If '-vrmode none' was used try to load the respective SDK Setup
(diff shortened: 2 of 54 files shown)


GitCop:

 * Commit: 2bcf7d87cf68489f7dc516ed0f6264a8f90054dd

---

Assets/VRTK/Source/SDK/Base/SDK_DescriptionAttribute.cs
```diff
@@ -138,10 +141,17 @@ public SDK_DescriptionAttribute(Type typeToCopyExistingDescriptionFrom, int inde
 
         public static SDK_DescriptionAttribute[] GetDescriptions(Type type)
         {
+#if UNITY_WSA && !UNITY_EDITOR
+            return type.GetTypeInfo().GetCustomAttributes(typeof(SDK_DescriptionAttribute), false)
```
bddckr: I'd prefer any `#if-#else-#endif` to only surround the one line that changed to prevent duplicated code getting out of sync between the two branches.

Here only the first line seems to be different.

bddckr: I'd prefer any `#if-#else-#endif` to only surround the one line that changed to prevent duplicated code getting out of sync between the two branches.

Here only the first line seems to be different.

Assets/VRTK/Source/Scripts/Utilities/SDK/VRTK_SDKManager.cs
```diff
@@ -1106,9 +1112,15 @@ private static void PopulateAvailableAndInstalledSDKInfos()
             Type fallbackType = SDKFallbackTypesByBaseType[baseType];
 
             availableSDKInfos.AddRange(VRTK_SDKInfo.Create<BaseType, FallbackType, FallbackType>());
+#if UNITY_WSA && !UNITY_EDITOR
+            availableSDKInfos.AddRange(baseType.GetTypeInfo().Assembly.GetExportedTypes()
+                                               .Where(type => type.GetTypeInfo().IsSubclassOf(baseType) && type != fallbackType && !type.GetTypeInfo().IsAbstract)
```
bddckr: Same as above.

DDReaper:
> Additionally I'd like to see a helper method in use to simplify the type vs. GetTypeInfo usage here, but I think the latter method is only available in .NET 4.5 which we aren't targeting. Therefore a method could not return the same type for both UNITY_WSA and regular Unity AFAIK.

It may be possible to write an extension method to switch out the call for just UWP, so the extension will only activate for UWP.  I'll test and see if I can simplify it.

Assets/VRTK/Source/SDK/Base/SDK_DescriptionAttribute.cs
DDReaper: It is only the first line, however it's part of an entire Linq query, which could be different in the future.  So better to encapsulate the whole linq query

DDReaper: if I can get the extension idea to work, it should be a lot cleaner.  But in my experience, UWP changes a lot of the rules sometimes and may need a different path.  Hopefully not in this case.

thestonefox:
@DDReaper I've updated the WMR pull request (https://github.com/thestonefox/VRTK/pull/1745) with these required scripting defines.

But rather than do it in each file, I've created a bunch of reflection helpers in the SharedMethods class, so only 1 file has to have the ifdefs in it.



(as mentioned on the other PR, builds work ok, but running the actual app still throws errors)

DDReaper:
Ok, @bddckr / @thestonefox 
I've posted an alternate proposal for the UWP Build issues.
https://github.com/ddreaper/VRTK/tree/feat/windows-mixed-reality-support-UWPUpdate2
This one uses extension methods to resolve the majority of the UWP Build issues, with the remaining items that couldn't be resolved with extensions done to the bare minimum.

Let me know which PR you want to merge.

SimonDarksideJ:
@thestonefox can't see your updates in #1745, did you forget to push the changes up?  Last change is reported as 6 days ago

bddckr:
@SimonDarksideJ The last commit is from two days ago. You can only see that in the PR's "discussion page": GitHub doesn't notice the commit dates changing in the "commit overview page" since the branch was changed and force-pushed so it still only contains that one commit.

PR #1751, opened 2018-03-22
refactor(PolicyList): provide alternative for deprecated EnumMaskField

thestonefox:
The `EnumMaskField` method on the `EditorGUILayout` class has been
deprecated in Unity 2017.3 and replaced with `EnumFlagsField`.

This short refactor just ensures the deprecated method is only used
in older versions of Unity where it is still supported.

Thanks to @qubytes for the fix.


qubytes:
Thanks for the credit @thestonefox ! No matter how small :D. Also, thanks for the effort you and others put into VRTK. It's an awesome library!

PR #1752, opened 2018-03-22
fix(Examples): display error if script define symbols have not been set

thestonefox:
There is an issue where the new example scenes don't set the required
VRTK script define symbols because this is done via the SDK Manager
script and that is no longer present in the example scenes as it is
dynamically loaded in via the constructor scene.

This fix simply throws an error if the example scene cannot find the
required scripting define symbols and attempts to provide useful
information for the user to rectify the issue.

### DIFF
--- a/Assets/VRTK/Examples/ExampleResources/SharedResources/Scripts/VRTKExample_AdditiveSceneLoader.cs
+++ b/Assets/VRTK/Examples/ExampleResources/SharedResources/Scripts/VRTKExample_AdditiveSceneLoader.cs
@@ -2,6 +2,8 @@
 {
 #if UNITY_EDITOR
     using UnityEditor;
+    using System.Collections.Generic;
+    using System.Linq;
 #endif
     using UnityEngine;
     using UnityEngine.SceneManagement;
@@ -26,6 +28,7 @@
 
         protected virtual void Awake()
         {
+            CheckScriptingDefineSymbols();
             if (!Application.isPlaying && Application.isEditor)
             {
                 ManageBuildSettings();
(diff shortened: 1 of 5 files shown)


Assets/VRTK/Examples/ExampleResources/SharedResources/Scripts/VRTKExample_AdditiveSceneLoader.cs
```diff
@@ -45,6 +48,21 @@ protected virtual void Awake()
...
+        protected virtual void CheckScriptingDefineSymbols()
+        {
+#if UNITY_EDITOR
+            IEnumerable<string> nonSDKSymbols = PlayerSettings.GetScriptingDefineSymbolsForGroup(EditorUserBuildSettings.selectedBuildTargetGroup)
+                    .Split(';')
+                    .Where(symbol => symbol.StartsWith(SDK_ScriptingDefineSymbolPredicateAttribute.RemovableSymbolPrefix, System.StringComparison.Ordinal));
+
+            HashSet<string> checkSymbols = new HashSet<string>(nonSDKSymbols);
+
+            if (checkSymbols.Count == 0)
```
bddckr: No need to create a `HashSet` and check for its `Count` here:

```c#
bool isMissingSDKSymbols = PlayerSettings
    .GetScriptingDefineSymbolsForGroup(EditorUserBuildSettings.selectedBuildTargetGroup)
    .Split(';')
    .All(symbol => !symbol.StartsWith(SDK_ScriptingDefineSymbolPredicateAttribute.RemovableSymbolPrefix, System.StringComparison.Ordinal));

if (isMissingSDKSymbols)
{
    VRTK_Logger.Error("No VRTK scripting define symbols have been found. " + VRTK_Logger.GetCommonMessage(VRTK_Logger.CommonMessageKeys.SCRIPTING_DEFINE_SYMBOLS_NOT_FOUND), true);
}
```

PR #1787, opened 2018-05-09
fix(UnityXR): fix HTC Vive button mapping

### LINKED ISSUE #1786: (UnityXR SDK) HTC Vive menu button doesn't work
### Environment

 * Source of VRTK: Github
 * Version of VRTK: 3.3.0-alpha 2364628
 * Version of Unity3D: 2017.4.2f2
 * Hardware used: HTC Vive
 * SDK used: Unity XR

### Steps to reproduce

1. Start the construction scene and an example scene that uses the menu button
  e.g. 004 Locomotion (Dash toggle), 005 Interactions (Hands toggle)
2. Press the menu button and observe that nothing happens.

### Expected behavior

The menu button should work:
Scene 004 - Should change teleport mode between blink and dash
Scene 005 - Should toggle the hands
Scene 006 - Should show / hide the menu

### Current behavior

The menu button does not work in any scene

jamre:
Change the Vive Controller Menu Button (Button Two) to button ids:
 2 (left) and 0 (right).
Change button KeyCodes to come from any joystick.
Comment out calls to SetControllerButtonValues since the joystickIndex
shouldn't be different between right and left controllers.

Fixes #1786

### DIFF
--- a/Assets/VRTK/Source/SDK/Unity/SDK_UnityController.cs
+++ b/Assets/VRTK/Source/SDK/Unity/SDK_UnityController.cs
@@ -71,22 +71,23 @@ namespace VRTK
         */
 
         protected int[] rightControllerTouchCodes = new int[] { 15, 17, 10, 11 };
-        protected int[] rightControllerPressCodes = new int[] { 9, 0, 1, 7 };
+        protected int[] rightControllerPressCodes = new int[] { 9, 1, 0, 7 };
+        protected int[] rightOculusRemotePressCodes = new int[] { 9, 0, 1, 7 };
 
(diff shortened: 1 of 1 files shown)


jamre:
I commented out the SetControllerButtonValues calls instead of removing them since I didn't know if they are needed for some other controller type that I don't have.

jamre:
Using the [any joystick] buttons works, but it is also working with the [specific joystick] buttons now too, so I uncommented it.

I also added a special case mapping for the Oculus Remote which, according to the Unity docs, has button one & two codes reversed from the Open VR controllers and Oculus touch.

Issue #1696, opened 2018-01-12
VRTK_PhysicsRotator and VRTK_ArtificialRotator rotations wrong when disabled and enabled


jbienzms:
### Environment

 * Source of VRTK: Github
 * Version of VRTK: release/3.3.0-alpha
 * Version of Unity3D: 2017.2.0f3
 * Hardware used: Tested on Vive and Samsung Odyssey (WMR)
 * SDK used: Tested on SteamVR and OpenVR

#### For **VRTK_ArtificialRotator**:

1. Get latest from **release/3.3.0-alpha** branch
2. Open **025_Controls_Overview** scene
3. Click **Play** in the Unity Editor
4. Use the controller to rotate controls under **ExampleWorldObjects/Rotators/Dials**
5. Make note of the rotations and values of each control.
6. Use the Unity Inspector to disable **ExampleWorldObjects/Rotators/Dials**
7. Use the Unity Inspector to enable **ExampleWorldObjects/Rotators/Dials**
8. Note that the rotations and values of each control have changed
9. Disable and Enable **ExampleWorldObjects/Rotators/Dials** again
10. Note that rotations and values of some controls have changed again

#### For **VRTK_PhysicsRotator**:

Follow the same steps above but use **ExampleWorldObjects/Rotators/Wheels** or **ExampleWorldObjects/Rotators/Levers** to see the same issue with PhysicsRotator.

### Expected behavior

Control / Wheel / Lever rotations and values should be the same after a control has been re-enabled as they were before the control was disabled.

### Current behavior

Control / Wheel / Lever rotations and values change (and in some cases continue to change) each time the control is re-enabled.

### Additional Information

This [post on Stack Overflow](https://answers.unity.com/questions/368249/hingejoint-stops-working-properly-after-disableena.html) suggests that HingeJoint does not correctly remember its angles between disable and re-enable. This could explain the issue with **VRTK_PhysicsRotator**, but it doesn't appear **VRTK_ArtificialRotator** leverages a HingeJoint.

jbienzms:
I would like to help fix this issue but I'm not sure where to start. I thought sure HingeJoint was the culprit until I saw the same behavior with **VRTK_ArtificialRotator** which doesn't seem to use one. If anyone can suggest how I might start trying to fix this bug I'm more than happy to try and fix it and submit a pull request.

PR #1808, opened 2018-05-18
fix(controllables): no longer resets value on re-enable

JWNJWN:
a SetValue function was added to the base class
all classes implement this
implemented for Sliders & Rotators
not for Pusher because resets anyway

fixes: #1696

### DIFF
--- a/Assets/VRTK/Source/Scripts/Interactions/Interactables/Controllables/Artificial/VRTK_ArtificialPusher.cs
+++ b/Assets/VRTK/Source/Scripts/Interactions/Interactables/Controllables/Artificial/VRTK_ArtificialPusher.cs
@@ -68,6 +68,14 @@ namespace VRTK.Controllables.ArtificialBased
         }
 
         /// <summary>
+        /// The SetValue method is not implemented as the pusher resets automatically.
+        /// </summary>
+        /// <param name="value">Not used.</param>
+        public override void SetValue(float value)
+        {
+        }
+
+        /// <summary>
         /// The IsResting method returns whether the pusher is currently at it's resting position.
         /// </summary>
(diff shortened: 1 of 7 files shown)


Assets/VRTK/Source/Scripts/Interactions/Interactables/Controllables/Artificial/VRTK_ArtificialPusher.cs
```diff
@@ -67,6 +67,11 @@ public override float GetNormalizedValue()
             return VRTK_SharedMethods.NormalizeValue(GetValue(), originalLocalPosition[(int)operateAxis], PressedPosition()[(int)operateAxis]);
         }
 
+        //Has no implementation as pushers reset automatically anyway.
```
thestonefox: Put proper docs in here otherwise the generator won't pick it up.

```
/// <summary>
/// The SetValue method is not implemented as the pusher resets automatically.
/// </summary>
/// <param name="value">Not used.</param>
```

Assets/VRTK/Source/Scripts/Interactions/Interactables/Controllables/Physics/VRTK_PhysicsPusher.cs
```diff
@@ -68,6 +68,11 @@ public override float GetNormalizedValue()
             return VRTK_SharedMethods.NormalizeValue(GetValue(), originalLocalPosition[(int)operateAxis], PressedPosition()[(int)operateAxis]);
         }
 
+        //Has no implementation as pushers reset automatically anyway.
```
thestonefox: same as above, proper docs

Issue #1698, opened 2018-01-13
feat: Advanced VRTK_PolicyList


RunenGamer:
### Environment

    Source of VRTK: Github
    Version of VRTK: release/3.3.0-alpha (e43089a)
    Version of Unity3D (e.g. Unity 2017.3.0f3)
    Hardware used (e.g. Vive)
    SDK used (e.g. SteamVR)

### Steps to reproduce

1. Open "041_Controller_ObjectSnappingToDropZones"
2. See that that "Valid Object List Policy" only supports Tags, Layers and Scripts
3. The scene uses tags to identify correct objects
4. In my game, i got > 100 snap drop zones/objects. Therefore I would need >100 different Tags/Layers/Scripts. 

### Expected behavior

I propose a forth Option for VRTK_PolicyList: "String Identifier"

If you chose VRTK_PolicyList.CheckType = "String Identifier"
Then VRTK_PolicyList.TypeCheck would check if the GameObject obj got an Script VRTK_PolicyListStringIdentifier. If so, it checks whether the string specified in VRTK_PolicyListStringIdentifier matches one identifier of VRTK_PolicyList.

Basically it's the same system like tags, but it don't mess up a projects tag-structure.

If you like it, I can start an PR immeadiatly, because I already programmed this change.

moscartong:
+1. I did something similar in one of my projects. Using a string identity is more convenient when there are lots of different types of objects.

bddckr:
I don't get this feature request I think:

> PolicyList supports tags
> The scene uses tags

???

bddckr:
Additionally:

> PolicyList supports Scripts
> I want VRTK_PolicyList.TypeCheck that checks if the GameObject obj got an Script

This is exactly what the Script functionality does, isn't it?!

RunenGamer:
Not really. The Script functionality checks whether a specified script is attached. 

Example:
You got 100 different Keys with 100 Attachpoints. 
If you do not want to use tags (see above), you can try to use the "script functionality". Therefore you need 100 different scripts (for example "Policy_Key_1.cs" ... "Policy_Key_100.cs").

My solution got only both VRTK_Scripts, where a string (for example "Key_1"..."Key_100") is specified in VRTK_PolicyList_StringIdentifier.


To answer the question:
> Additionally:
>
>  PolicyList supports Scripts
>  I want VRTK_PolicyList.TypeCheck that checks if the GameObject obj got an Script
>
[...]

PR #1813, opened 2018-05-19
feat(Interaction): extension to the VRTK_PolicyList

RunenGamer:
Adds the stringIdentifier option to VRTK_PolicyList
(see #1698).
Allows user to specify a string on any game object.
This string will be checked against any
VRTK_PolicyList's identifiers to determine, whether
two objects should interact.

### DIFF
--- a/Assets/VRTK/Source/Scripts/Utilities/VRTK_PolicyList.cs
+++ b/Assets/VRTK/Source/Scripts/Utilities/VRTK_PolicyList.cs
@@ -5,10 +5,10 @@ namespace VRTK
     using System.Collections.Generic;
 
     /// <summary>
-    /// The Policy List allows to create a list of either tag names, script names or layer names that can be checked against to see if another operation is permitted.
+    /// The Policy List allows to create a list of either tag names, script names, layer names or string identifier that can be checked against to see if another operation is permitted.
     /// </summary>
     /// <remarks>
-    /// A number of other scripts can use a Policy List to determine if an operation is permitted based on whether a game object has a tag applied, a script component on it or whether it's on a given layer.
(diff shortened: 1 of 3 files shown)


bddckr:
What's the difference between this and adding (empty, dumb) components that just tag an object which is then used with the existing `CheckTypes.Script`?

RunenGamer:
Example: (It is not limited to snap functionality.)
If you got 100 Doors with one keyhole each. Also, there are 100 individual keys, one for each door (for snap function to open doors - If you need more of a 1:1 mapping of items to snap drop zones.). Currently you need to make 100 Scripts or Tags: "Key1.cs"... and therefore screw up the script or the tag structure.
With the improvement you can add the the VRTK_PolicyListStringIdentifier to all 100 keys and simple specify the string there "Key1"... 

@m-oscartong also add this feature in his project.

bddckr:
In that case it is probably easier to have a generic component on the key and a generic subclass of the SnapDropZone or an in-between script that listens to the snap and does something (like unsnap when it's the wrong key).
I see the use for it done in this way, though.

RunenGamer:
@bddckr I'm not sure, whether I get your point right...
But your 
> generic component on the key

may now be the VRTK_PolicyListStringIdentifier and the
> generic subclass of the SnapDropZone or an in-between script that listens to the snap and does something 

is the extended VRTK_PolicyList (which is not a subclass of SnapDropZone of course).
The value of this feature is not limited to SnapDropZones. 

Thanks for your review so far.

RunenGamer:
@thestonefox : After 3.3.0 release: Please add the new to come "Requires Community Testing/Approval" label to this PR.

thestonefox:

Issue #1697, opened 2018-01-13
(pointer) When click off "LeftHandAnchor" and then click it on , the pointer useless.


BensonZheng2016:
### Environment

 * Source of VRTK (Github)
 * Version of VRTK (3.3.0-alpha) (6cb308689b6d786ad62f28bb7e34ca8dec1f6018)
 * Version of Unity3D (Unity 2017.1.0f3)
 * Hardware used (Oculus)
 * SDK used (Oculus Utilities 1.16.0)

### Steps to reproduce
1.Open "034_Controls_InteractingWithUnityUI" scene.
2.Start it,
1) Disable the "LeftHandAnchor" game object.
![image](https://user-images.githubusercontent.com/18454958/34903933-9c7ac232-f876-11e7-825d-92e180f5a6fe.png)
![image](https://user-images.githubusercontent.com/18454958/34903941-b3f80d70-f876-11e7-860c-146fdbf0e7ec.png)
![image](https://user-images.githubusercontent.com/18454958/34903958-d3f1b0a4-f876-11e7-817a-de62bf88f049.png)
2) Enable the "LeftHandAnchor" game object 

### Current behavior
The left pointer is invalid. It couldn't click button and so on.

JWNJWN:
I can confirm that this behaviour occurs.
Though I am not certain whether it is expected behaviour that you can disable and enable the SDK specific objects whenever you want.
What do you think? @thestonefox

bddckr:
I'd vote for VRTK to still work even if you disable+enable controller objects.

thestonefox:
Closing this as it's related to an older version of VRTK (v3.3) the master branch is now on VRTK v4. Feel free to request to reopen if the issue is still present.

PR #1832, opened 2018-06-15
fix(OculusSDK): Invalid TrackedController Index

JWNJWN:
Added a check for invalid Index along with the null check.

fixes: #1697


thestonefox:

PR #1875, opened 2018-08-18
fix(Tooltip): change from world position into local, added offset

OriginLive:
Hello, I've changed the tooltip's line renderer into local space, instead of world space. This speeds it up and makes it draw a lot faster. The example is similar to what valve's sdk uses. I've also added an offset, so that you can draw a line a bit above your target (eg. above trackpad) instead of dead center.


thestonefox:

PR #1877, opened 2018-08-20
fix(Controller): fix thumbstick axis movement on Oculus Touch

Biliogadafr:
Allow to detect thumbstick movement on Oculus Touch when sensitive pad at the top is not
touched and SteamVR backend is used.
The problem was described and solved for Oculus SDK in this commit https://github.com/thestonefox/VRTK/commit/b213073daec55c4a98f947d5fab11416ec492908 . 
But it was fixed only for OculusSDK + Oculus and not for SteamVR + Oculus.


thestonefox:

PR #1886, opened 2018-09-01
fix(SDK): handle touch and classic controller plugged in - fixes #1866

### LINKED ISSUE #1866: Oculus Integrations not working correctly with VRTK
VRTK Source:Github
VRTK Version: 3.3.0 (I think or whatever most recent master is called)
Unity3D Version: 2018.1.0f2
Using Oculus
Oculus Utilities 1.27.0

Steps to reproduce
Switch the Oculus SDK in an pointer example scene to Oculus Utilities and play.

Expected behavior

Each hands pointer is dependent on its own hand's touch pad inputs.

Current behavior

Left hand inputs control both hands inputs. So, when you touch the joystick both hand's pointers activate, but on right hand inputs no output.

Other Notes

Local avatar works perfectly fine finger positions respond accurately to actual finger positions. Tracking is fine. Touch and near touch still work correctly, but in order to grab an object with right hand left hand grab needs to be used. All inputs work correctly through UnityXR.

TMaloteaux:
Added the missing switch case to handle the value returned by OVRInput.GetConnectedControllers when touch controllers
and a classic controller are active in the same time.

### DIFF
--- a/Assets/VRTK/Source/SDK/Oculus/SDK_OculusController.cs
+++ b/Assets/VRTK/Source/SDK/Oculus/SDK_OculusController.cs
@@ -822,6 +822,7 @@ namespace VRTK
 
             switch (activeControllerType)
             {
+                case OVRInput.Controller.Gamepad | OVRInput.Controller.RTouch | OVRInput.Controller.LTouch:
                 case OVRInput.Controller.Touch:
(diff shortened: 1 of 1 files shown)


thestonefox:
Try and muster support to get this tested and approved and I'll merge it :)

grizzlypath26:
Just tested in the example scenes and touch controllers work properly now while an xbox controller is plugged in

PR #1891, opened 2018-09-11
fix(SDKSetup): don't try to fix open but unloaded scenes

jamre:
This is related to #1889.

The same problem happens in SDKSetupEditor, but I had Clear on Play enabled and the error was removed from the console before I saw it.


thestonefox:
No longer merge-able with the master branch. Closing

PR #1914, opened 2018-10-23
fix(Interaction): Flag controllerAttachPoint externally set

KoltesDigital:
When a `VRTK_Pointer` has both `Activate on Enable` and `Grab To Pointer Tip` checked, when the game runs, the grabbing is made with the controller itself, not the pointer tip. The tip grabs only after deactivating and reactivating the pointer.

The reason is: `VRTK_BasePointerRenderer` sets the `controllerAttachPoint` to the tip, and just after `VRTK_InteractGrab` resets it to the default attach point, on the controller.

This patch adds a bool to flag whether `controllerAttachPoint` has been set by `VRTK_BasePointerRenderer`.


thestonefox:
No longer merge-able with the master branch. Closing

PR #1993, opened 2019-04-28
Feat/pointer events

thestonefox:


thestonefox:
This is the precursor to the destination point work. I can continue with the dest point stuff as commits on this PR if need be, but if it can be approved and merged before that work is started then that would be better I think because the changes to the pointer will potentially affect users who are already using the ObjectPointer events

Prefabs/Pointers/SharedResources/Scripts/PointerFacade.cs
```diff
@@ -62,6 +63,39 @@ public enum SelectionType
...
         #endregion
 
+        #region Pointer Events
+        /// <summary>
+        /// Emitted when the <see cref="ObjectPointer"/> becomes active.
+        /// </summary>
+        [Header("Pointer Events"), DocumentedByXml]
+        public ObjectPointer.UnityEvent Activated = new ObjectPointer.UnityEvent();
+        /// <summary>
+        /// Emitted when the <see cref="ObjectPointer"/> is deactivated
```
bddckr:
```suggestion
        /// Emitted when the <see cref="ObjectPointer"/> is deactivated.
```

PR #1997, opened 2019-05-03
Feat/destpoints

thestonefox:


Prefabs/Locomotion/DestinationLocations/SharedResources/Scripts/DestinationLocation.cs
```diff
@@ -0,0 +1,238 @@
...
+            IsActivated = false;
+            Deactivated?.Invoke();
+        }
+
+        /// <summary>
+        /// Generates the payload to emit on the hovering events of <see cref="Entered"/> and <see cref="Exited"/>.
+        /// </summary>
+        /// <param name="data">The data that is mutating the hover state.</param>
+        /// <returns>The data to emit.</returns>
+        protected virtual SurfaceData GenerateHoverPayload(SurfaceData data)
```
bddckr: Instead of `Generate` use `Create` as that seems more widely used in .NET API. (Also update the XML docs on this method to say `Creates` then.)

Prefabs/Locomotion/DestinationLocations/SharedResources/Scripts/DestinationLocation.cs
```diff
@@ -0,0 +1,238 @@
...
+            data.CollisionData = hoverHit;
+            return data;
+        }
+
+        /// <summary>
+        /// Generates the payload to emit on the <see cref="Activated"/> event.
+        /// </summary>
+        /// <param name="data">The default data to potentially mutate.</param>
+        /// <returns>The data to emit.</returns>
+        protected virtual TransformData GenerateSelectedPayload(SurfaceData data)
```
bddckr: Same here: `Create(s)` instead of `Generate(s)`.

Prefabs/Locomotion/DestinationLocations/SharedResources/Scripts/DestinationLocationFacade.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        #region Location Settings
+        /// <summary>
+        /// Determines if the location is in the locked and unusable state.
+        /// </summary>
+        [Serialized]
+        [field: Header("Location Settings"), DocumentedByXml]
+        public bool IsLocked { get; set; }
+        /// <summary>
+        /// Apply the rotation of the custom destination location to the selected action output.
```
bddckr: `Whether to apply the rotation of the custom destination location to the selected action output.`

Prefabs/Locomotion/DestinationLocations/SharedResources/Scripts/DestinationLocationFacade.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public bool ApplyDestinationRotation { get; set; } = true;
```
bddckr: `AppliesDestinationRotation`

Prefabs/Locomotion/DestinationLocations/SharedResources/Scripts/DestinationLocationLockedStateTag.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace VRTK.Prefabs.Locomotion.DestinationLocations
+{
+    using UnityEngine;
+
+    public class DestinationLocationLockedStateTag : MonoBehaviour
+    {
+        public virtual void Start() { }
```
bddckr: I can't find it anymore, but I think we have/had a comment on another empty method like this somewhere to explain that we only do implement this as an empty method to allow showing the enabled checkbox in the inspector. Maybe we do this here, too?

Prefabs/Pointers/SharedResources/Scripts/PointerFacadeComponentGameObjectExtractor.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <summary>
+        /// Sets the <see cref="Source"/> based on a given <see cref="GameObject"/>.
+        /// </summary>
+        /// <param name="source">The data that contains the <see cref="PointerFacade"/> component.</param>
+        [RequiresBehaviourState]
+        public virtual void SetSource(GameObject source)
+        {
+            if (source == null)
+            {
+                return;
```
bddckr: It looks like before returning we should do `Source = null;` here to ensure the "set to null" is actually done in this case.

Prefabs/Pointers/SharedResources/Scripts/PointerFacadeComponentGameObjectExtractor.cs
```diff
@@ -0,0 +1,119 @@
...
+            }
+
+            Source = source.GetComponent<PointerFacade>();
```
bddckr: Use our extension method `TryGetComponent` here to work around the editor garbage generation in case that component doesn't exist.

If we apply this suggestion I think we can also get rid of the `if` in this method body as the extension method handles calling it on a `null` object!

Prefabs/Locomotion/DestinationLocations/SharedResources/Scripts/DestinationLocationFacade.cs
thestonefox: We use `Apply` everywhere else (e.g. ApplyOffset, ApplyPositionOffsetOnAxis)

I'd say we raise a ticket to change it globally in the future (as a global change will break existing prefab setups) instead of creating an inconsistency now

PR #2001, opened 2019-05-08
Feat/pointer grabbing

thestonefox:


thestonefox:
Due to GitHub not understanding git order and ordering commits by date, the actual commit order is:

```
7ee4279 feat(Interactions): digest Interactor grab from SurfaceData
98a5878 chore(Dependencies): update Zinnia submodule to latest head
b7f3b10 feat(examples): add ability to use pointer to pick up objects
```

(bottom being the newest commit)

Prefabs/Interactions/Interactors/SharedResources/Scripts/InteractorFacade.cs
```diff
@@ -117,6 +118,20 @@ public virtual void Grab(InteractableFacade interactable)
             Grab(interactable, null, null);
         }
 
+        /// <summary>
+        /// Attempts to grab a found Interactable in the given <see cref="SurfaceData"/> to the current Interactor.
```
bddckr: `by the current Interactor`

thestonefox: what does that replace? what would the full sentence be? the comment already mentions it is grabbed to the current Interactor.

bddckr: "grab to something" is wrong AFAIK. My suggestion is to use "grabbed by".

Then additionally this is a method on the InteractorFacade, thus there is no way to have multiple interactors, i.e. "current Interactor" is kinda making me think "wait a second, how can I have multiple with this single facade?!" as a reader.

```c#
/// Attempts to grab a found Interactable in the given <see cref="SurfaceData"/> by this Interactor.
```

thestonefox: All of the other methods should be update. but still `by this Interactor` makes little sense too.

Maybe they should read like

`/// Attempts to attach the Interactable found in the given SurfaceData to this Interactor`

PR #2024, opened 2019-05-31
fix(examples): ensure destination point models render correctly

thestonefox:
There was an issue when the models were imported into a scene in Unity
2019.1 and then opened in an earlier version of Unity (2018.3) where
the model prefab broke.

This has re-built the models in Unity 2018 to prevent any future
breaks.


thestonefox:
@bddckr I've got a number of example scene improvements to make. rather than wait for them to be reviewed and approved, should I just go ahead and self merge so the approval process isn't holding anything up?

PR #2040, opened 2019-06-22
Fix/interactables - fixes #2038

### LINKED ISSUE #2038: Interactable Rigidbody state isn't managed correctly
### Environment

* GitHub Master
* master
* 2019.1.5f1
* Oculus
* Oculus Integration

### Steps to reproduce

Case 1:

* Set an interactable to `Follow Tracking: Follow Transform`
* Set the Interactable rigidbody to kinematic
* grab the interactable
* ungrab the interactable

Case 2:

* Set an interactable to `Follow Tracking: Follow Rigidbody`
* Set the Interactable rigidbody to kinematic
* grab the interactable
* ungrab the interactable

Case 3:

* Set an interactable to `Follow Tracking: Follow Rigidbody`
* Snap the interactable to a snap drop zone
* grab the interactable

### Expected behavior

Case 1: The interactable is still kinematic on release as it was before grabbing

Case 2: The interactable follows the interactor as it's being grabbed

Case 3: The interactable is unsnapped from the snap zone.

### Current behavior

Case 1: The interactable goes back to being non-kinematic on release even though it's initial state was kinematic because the logic is always forcing it off.

Case 2: The interactable doesn't move because it needs to be non-kinematic for the rigidbody tracking to work but no kinematic state is changed.

Case 3: The interactable won't unsnap, because on snap the kinematic state is changed to be kinematic but the unsnap never resets the kinematic state to what it originally was.

thestonefox:


Prefabs/Interactions/Interactables/SharedResources/Scripts/Grab/Action/GrabInteractableFollowAction.cs
```diff
@@ -173,6 +181,7 @@ protected virtual void OnEnable()
         /// </summary>
         protected virtual void ConfigureFollowTracking()
         {
+            IsKinematicWhenInactive = GrabSetup ? GrabSetup.Facade.ConsumerRigidbody.isKinematic : false;
```
bddckr:
```suggestion
            IsKinematicWhenInactive = GrabSetup == null ? false : GrabSetup.Facade.ConsumerRigidbody.isKinematic;
```
`GrabSetup` inherits from `UnityEngine.Object`!

Too bad Unity forces us to use this null check, otherwise this could have just been `GrabSetup?.Facade.ConsumerRigidbody.isKinematic ?? false` 😞

PR #2066, opened 2019-10-13
UPM package & CI+CD - resolves #2065

### LINKED ISSUE #2065: Make compatible with the Unity Package Manager and remove git submodule
Currently VRTK imports Zinnia into the project via a git submodule, which in turn includes the hard dependency of Malimbe.

There is some work to be done on Zinnia to make it a valid Unity package and remove the hard dependency of Malimbe.

When this happens it will break VRTK as there is a requirement to pull in the dependencies via a git submodule.

Moving forward, including VRTK will require including the package in the project manifest.

Zinnia ticket reference https://github.com/ExtendRealityLtd/Zinnia.Unity/issues/430

bddckr:


bddckr:
Depends on https://github.com/ExtendRealityLtd/Zinnia.Unity/pull/436.

PR #2069, opened 2019-10-15
fix(examples): use correct layer for pointer grab validity rule

extendreality:
The pointer grab validity rule was still set to Water meaning pointer
grabbing wasn't working. It has now been set to the correct layer.

OpenVR has also been added to the manifest.json just so it's already
included for anyone wanting to use an OpenVR headset.


thestonefox:
> No love for the Oculus package? 😏 I'm sure this is fine 😄

The oculus package was already included

PR #2129, opened 2021-09-23
feat(examples): migrate project to unity 2020.3 input system and openXR

thestonefox:
The project has now been migrated to Unity 2020.3 LTS and utilises the
updated Unity Input System and the OpenXR framework for hardware
support.

This is the recommended way of developing for VR in Unity moving
forward so it is best to show the example this way.

All of the Tilia packages and Zinnia still work fine in Unity
2018.3.10f and above.


thestonefox:
This is the bug in Unity
https://issuetracker.unity3d.com/issues/unity-does-not-execute-code-weavers-when-its-opened-for-the-first-time

